PR #3, opened 2024-02-26
feat: gas service

canhtrinh:
Draft PR with remaining questions:
- [x] can we review overall structure of this impl? largely followed the structure of the gateway contracts but want to ensure it follows the same proxy/impl approach we want
- [x] for amounts, the solidity gas services uses u256 but transfer methods for tokens take i128. 
- [x] what is the right pattern for auth? for example, want to make sure collect_fees and refund can only be invoked by gas_collector (which, in our Solidity impl, are set in the contract constructor) UPDATE: think i see the pattern, but do still want to discuss it before this is merged
- [x] how to check for address zero?
- [x] in my tests, not getting token transfers to work. 
- [x] in my tests, not getting assertion of emitted events to work.


contracts/axelar-gas-service/src/test.rs
```diff
@@ -0,0 +1,67 @@
...
+fn refund() {
+    let (env, contract_id, client) = setup_env();
+
+    let tx_hash: BytesN<32> = bytesn!(
+        &env,
+        0xfded3f55dec47250a52a8c0bb7038e72fa6ffaae33562f77cd2b629ef7fd424d
+    );
+    let log_index: U256 = U256::from_u32(&env, 1);
+    let receiver: Address = Address::generate(&env);
+    let token_addr: Address = Address::generate(&env);
```
haiyizxx, line 59: this is admin address, get passed as token address in refund function

contracts/axelar-gas-service/src/test.rs
```diff
@@ -0,0 +1,160 @@
...
+    assert_eq!(supply, token_client.balance(&contract_id));
+
+    client.collect_fees(&gas_collector, &token_address, &refund_amount);
+
+    assert_eq!(refund_amount, token_client.balance(&gas_collector));
+    assert_eq!(supply - refund_amount, token_client.balance(&contract_id));
+
+    assert_emitted_event(
+        &env,
+        3, //events 0-2 are related to token setup and transfer
```
milapsheth, line 120: use `-1` to assert the last emitted event
```suggestion
        -1,
```

contracts/axelar-gas-service/src/test.rs
```diff
@@ -0,0 +1,160 @@
...
+        &token_address,
+        &refund_amount,
+    );
+
+    assert_eq!(refund_amount, token_client.balance(&receiver));
+    assert_eq!(supply - refund_amount, token_client.balance(&contract_id));
+
+    assert_emitted_event(
+        &env,
+        2, //events 0 and 1 are related to token setup
```
milapsheth, line 155: same

contracts/axelar-gas-service/src/types.rs
```diff
@@ -0,0 +1,11 @@
+use soroban_sdk::{contracttype, Address, BytesN, U256};
+
+#[contracttype]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct Refunded {
```
milapsheth, line 5: unused

contracts/axelar-gas-service/src/test.rs
```diff
@@ -0,0 +1,160 @@
...
+
+#[test]
+fn collect_fees() {
+    let (env, contract_id, gas_collector, token_address, token_client, supply, client) =
+        setup_env_with_token();
+
+    let refund_amount = 1;
+
+    assert_eq!(0, token_client.balance(&gas_collector));
+    assert_eq!(supply, token_client.balance(&contract_id));
```
milapsheth, line 111: this assertion is a bit unnecessary since it's true by design. It's better to mint tokens on the contract within the test instead of during setup (since the gas service won't start off with tokens by default)

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -0,0 +1,109 @@
...
+
+#[contractimpl]
+impl AxelarGasServiceInterface for AxelarGasService {
+    fn pay_native_gas_for_contract_call(
+        env: Env,
+        sender: Address,
+        destination_chain: String,
+        destination_address: String,
+        payload: Bytes,
+        refund_address: Address,
```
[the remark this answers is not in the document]
canhtrinh: ah, yes, thanks for catching the token/amt lock. added a simple transfer for now. 

and yeah good point on attaching token with call. nothing obvious in the docs or code samples on this; they do highlight differentiations to Solidity payable methods but not the equivalent here. i'll ask

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -0,0 +1,120 @@
...
+    ) {
+        sender.require_auth();
+
+        let TokenDetails { token_addr, amount } = token_details.clone();
+
+        if amount == 0 {
+            panic_with_error!(env, Error::InvalidAmount);
+        }
+
+        token::Client::new(&env, &token_addr).transfer(
```
milapsheth, line 53: [transfer_from](https://docs.rs/soroban-sdk/latest/soroban_sdk/token/struct.TokenClient.html#method.transfer_from)?

contract can't transfer from sender. The tests should avoiding mocking the auth to be able to catch these issues, but need to check how issue a tx from a user address to set it up correctly.

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -0,0 +1,120 @@
...
+            sender,
+            destination_chain,
+            destination_address,
+            payload,
+            refund_address,
+            token_details,
+        );
+    }
+
+    fn collect_fees(env: Env, receiver: Address, token_addr: Address, amount: i128) {
```
milapsheth, line 70:
```suggestion
    fn collect_fees(env: Env, receiver: Address, token: TokenDetails) {
```

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -0,0 +1,120 @@
...
+#[contractimpl]
+impl AxelarGasServiceInterface for AxelarGasService {
+    fn pay_gas_for_contract_call(
+        env: Env,
+        sender: Address,
+        destination_chain: String,
+        destination_address: String,
+        payload: Bytes,
+        refund_address: Address,
+        token_details: TokenDetails,
```
milapsheth, line 43:
```suggestion
        token: Token,
```
TokenDetails sounds like it contains name, symbol etc as well

packages/axelar-soroban-std/src/types.rs
```diff
@@ -1,3 +1,10 @@
-use soroban_sdk::BytesN;
+use soroban_sdk::{contracttype, Address, BytesN};
 
 pub type Hash = BytesN<32>;
+
+#[contracttype]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct TokenDetails {
+    pub token_addr: Address,
+    pub amount: i128,
```
milapsheth, line 9:
```suggestion
pub struct Token {
    pub address: Address, // TODO: check if this can be changed to a TokenClient type instead which is richer than Address, or a generic type implementing TokenInterface
    pub amount: i128,
```
best to avoid truncated names. `token.address`, `token.amount` usage should be more readable

contracts/axelar-soroban-interfaces/src/axelar_gas_service.rs
```diff
@@ -0,0 +1,30 @@
...
+        sender: Address,
+        destination_chain: String,
+        destination_address: String,
+        payload: Bytes,
+        refund_address: Address,
+        token_details: TokenDetails,
+    );
+
+    /// Allows the gasCollector to collect accumulated fees from the contract.
+    fn collect_fees(env: Env, receiver: Address, token_addr: Address, amounts: i128);
```
milapsheth, line 19:
```suggestion
    fn collect_fees(env: Env, receiver: Address, token: Token);
```
amounts -> amount

contracts/axelar-soroban-interfaces/src/axelar_gas_service.rs
```diff
@@ -0,0 +1,30 @@
...
+
+    /// Refunds gas payment to the receiver in relation to a specific cross-chain transaction. Only callable by the gasCollector.
+    fn refund(
+        env: Env,
+        tx_hash: Hash,
+        log_index: U256,
+        receiver: Address,
+        token: Address,
```
milapsheth, line 27: same

contracts/axelar-gas-service/src/event.rs
```diff
@@ -0,0 +1,40 @@
...
+        (destination_address, payload, refund_address, token_details),
+    );
+}
+
+pub(crate) fn refunded(
+    env: &Env,
+    tx_hash: Hash,
+    log_index: U256,
+    receiver: &Address,
+    token: &Address,
```
milapsheth, line 30: reuse Token type

packages/axelar-soroban-std/src/types.rs
```diff
@@ -1,3 +1,10 @@
-use soroban_sdk::BytesN;
+use soroban_sdk::{contracttype, Address, BytesN};
 
 pub type Hash = BytesN<32>;
+
+#[contracttype]
+#[derive(Clone, Debug, Eq, PartialEq)]
+pub struct TokenDetails {
```
milapsheth, line 7: :+1:

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -0,0 +1,120 @@
...
+            panic_with_error!(env, Error::InsufficientBalance);
+        }
+
+        event::fee_collected(&env, &gas_collector, &token_addr, amount);
+    }
+
+    fn refund(
+        env: Env,
+        tx_hash: Hash,
+        log_index: U256,
```
milapsheth, line 99: combine into a single messageId string to future proof it

contracts/axelar-gas-service/src/contract.rs
canhtrinh: in principle i agree with you, but seems like the mechanism is different here, (though i'd like to confirm with them).

* their docs (https://developers.stellar.org/docs/learn/migrate/evm/solidity-and-rust-advanced-concepts#soroban-5) emphasize how the equivalent Solidity convention of requiring approval is replaced/replaceable entirely by the built-in functions in Soroban, e.g. `require_auth`. 
* on the token client implementations (https://github.com/stellar/soroban-examples/blob/v20.0.0/token/src/contract.rs#L100-L127) seem more generalized for both `transfer` and `transfer_from`, in each case accepting both a sender address parameter and receiver address parameter, and there's an auth check on the sender and spender, respectively.
* many of their examples (here, for example: https://github.com/stellar/soroban-examples/blob/v20.0.0/atomic_swap/src/lib.rs#L60), assume the implementation above and have contracts that invoke the `transfer` method when sending funds from a sender to the contract

that said, i'm not entirely sure what the use case is for a `transfer_from` method, as the implementations there seem to invoke transfer from a to b with an allowance from some third party c (where the contract is none of a, b, or c).

milapsheth: `account.require_auth()` in it's simplest form means that `account` must be the `msg.sender`. If you compare to an ERC20 example, it's the user/sender who calls ERC20.approve first (in a separate tx). Then the contract (spender), calls transfer_from within the contract. The token contract `transfer_from` checks if the spender is the caller of the method, whereas `transfer` checks that `sender` is the caller. If the contract calls `transfer`, the contract address is the caller, so `sender.require_auth()` will fail. The implementation should translate directly from EVM, take a look at `payGasForContractCall` usage.

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -29,31 +27,27 @@ impl AxelarGasService {
...
         refund_address: Address,
-        token_details: TokenDetails,
+        token: Token,
     ) {
         sender.require_auth();
 
-        let TokenDetails { token_addr, amount } = token_details.clone();
-
-        if amount == 0 {
+        if token.amount == 0 {
```
milapsheth, line 42:
```suggestion
        if token.amount <= 0 {
```
since it's a signed int

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -76,31 +70,24 @@ impl AxelarGasServiceInterface for AxelarGasService {
 
         gas_collector.require_auth();
 
-        if amount == 0 {
+        if token.amount == 0 {
```
milapsheth, line 73:
```suggestion
        if token.amount <= 0 {
```

contracts/axelar-gas-service/src/event.rs
```diff
@@ -18,23 +18,16 @@ pub(crate) fn gas_paid_for_contract_call(
...
-    env.events().publish(topics, (token, amount));
+pub(crate) fn refunded(env: &Env, message_id: String, receiver: Address, token: Token) {
+    let topics = (symbol_short!("refunded"), message_id, receiver, token);
+    env.events().publish(topics, ());
 }
 
-pub(crate) fn fee_collected(env: &Env, receiver: &Address, token_address: &Address, amount: i128) {
-    let topics = (symbol_short!("collected"), receiver, token_address, amount);
+pub(crate) fn fee_collected(env: &Env, receiver: Address, token: Token) {
+    let topics = (symbol_short!("collected"), receiver, token.clone());
```
milapsheth, line 31:
```suggestion
    let topics = (symbol_short!("collected"), receiver, token);
```
why clone it?

contracts/axelar-gas-service/src/test.rs
```diff
@@ -32,8 +34,8 @@ fn pay_gas_for_contract_call() {
     let token_addr: Address = env.register_stellar_asset_contract(Address::generate(&env));
```
milapsheth, line 34:
```suggestion
    let token_address: Address = env.register_stellar_asset_contract(Address::generate(&env));
```

contracts/axelar-gas-service/src/test.rs
```diff
@@ -77,11 +87,14 @@ fn collect_fees() {
...
```
milapsheth, line 87:
```suggestion
    let token_address: Address = env.register_stellar_asset_contract(Address::generate(&env));
```

milapsheth (approved):

contracts/axelar-gas-service/src/event.rs
canhtrinh: yes, unnecessary

PR #8, opened 2024-03-07
feat: test coverage part one

deanamiel:
Soroban contract test coverage part 1 as per [AXE-3331](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-3331)

[AXE-3331]: https://axelarnetwork.atlassian.net/browse/AXE-3331?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -44,6 +106,116 @@ fn test_validate_proof() {
...
+
+    initialize(&env, &client, user, randint(0, 10), randint(1, 10));
+
+    let different_signers = generate_signer_set(&env, randint(1, 10), false);
+
+    let msg = Bytes::from_array(&env, &[0x01, 0x02, 0x03]);
+    let msg_hash = env.crypto().keccak256(&msg);
+    let proof = generate_proof(&env, msg_hash.clone(), different_signers);
+
+    // should panic, epoch should return zero for invalid signer set
```
milapsheth, line 123:
```suggestion
    // should panic, epoch should return zero for unknown signer set
```

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -44,6 +106,116 @@ fn test_validate_proof() {
...
+#[test]
+#[should_panic(expected = "invalid signatures")]
+fn test_fail_validate_proof_invalid_signer_set() {
+    let (env, _, client) = setup_env();
+    let user = Address::generate(&env);
+
+    let signers = initialize(&env, &client, user, randint(0, 10), randint(1, 10));
+
+    let msg = Bytes::from_array(&env, &[0x01, 0x02, 0x03]);
+    let msg_hash = env.crypto().keccak256(&msg);
```
milapsheth, line 173: repetition, create a util to generate random payload and hash. you can use `BytesN::<10>::random(&env);`

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -79,6 +251,155 @@ fn test_transfer_operatorship() {
...
+        randint(1, 10),
+    );
+
+    let empty_signer_set = generate_empty_signer_set(&env);
+
+    let empty_operator_set = empty_signer_set.to_xdr(&env);
+
+    // should throw an error, empty signer set
+    let res = client.try_transfer_operatorship(&empty_operator_set);
+    assert!(res.is_err());
```
milapsheth, line 275: same

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -79,6 +251,155 @@ fn test_transfer_operatorship() {
...
+    );
+
+    let mut new_signers = generate_signer_set(&env, randint(1, 10), false);
+
+    let env = &env;
+    let zero = U256::from_u32(env, 0);
+    let one = U256::from_u32(env, 1);
+    let mut total_weight = zero.clone();
+
+    for weight in new_signers.signer_set.signers.iter().map(|s| s.1) {
```
milapsheth, line 351: try to avoid using for loops whenever possible. use the reduce/fold functional pattern to accumulate
https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.fold

the extra `map` also isn't needed. you can destructure the argument in the fold itself

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -79,6 +251,155 @@ fn test_transfer_operatorship() {
...
+        randint(1, 10),
+    );
+
+    let mut new_signers = generate_signer_set(&env, randint(1, 10), false);
+
+    let len = new_signers.signer_set.signers.len();
+
+    // create a new vec and reverse signer order
+    let mut reversed_signers = Vec::new(&env);
+    for i in (0..len).rev() {
```
milapsheth, line 388: use functional approach. i recommend keeping the std library doc at handy. lot of useful functions available there, such as reversing a vec

contracts/axelar-auth-verifier/src/contract.rs
```diff
@@ -70,7 +70,7 @@ impl AxelarAuthVerifierInterface for AxelarAuthVerifier {
             .storage()
             .persistent()
             .get(&DataKey::EpochBySignerHash(signer_set_hash))
-            .unwrap();
+            .unwrap_or(0);
```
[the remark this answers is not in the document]
deanamiel: I made the change, now I don't see how this code can ever be reached:
```
  if signer_set_epoch == 0 {
            panic!("invalid epoch");
  }
```
If the signer set does not exist then `.get()` will return `none` which will panic with the expect statement above. Also once the contract is initialized the epoch is zero until `rotate_signer_set` is called in the initialize function which will increment the epoch by 1.

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -28,6 +31,65 @@ fn test_initialize() {
...
+    let owner = Address::generate(&env);
+
+    // create an empty WeightedSigners vector
+    let empty_signer_set = generate_empty_signer_set(&env);
+
+    // serialize the empty signer set to Bytes
+    let empty_operator_set = empty_signer_set.to_xdr(&env);
+
+    // call should panic because signer set is empty
+    let res = client.try_initialize(&owner, &randint(0, 10), &empty_operator_set);
```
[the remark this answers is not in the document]
deanamiel: I've only been able to use `should_panic` with functions that panic with string messages, not error types. The previous function checks the panic that errors with the string "Already initialized" whereas this test is expecting a panic with the error type `Error::InvalidOperators`. Let me know if there is a better way to listen for a specific error type

contracts/axelar-auth-verifier/src/testutils.rs
```diff
@@ -40,6 +44,13 @@ pub fn generate_signer_set(env: &Env, num_signers: u32) -> TestSignerSet {
         })
         .collect();
 
+    if include_zero_weight {
```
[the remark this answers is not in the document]
deanamiel: Makes sense, I reverted the changes to `generate_signer_set` and just made the last signers weight zero in the test

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -28,6 +31,65 @@ fn test_initialize() {
...
+    let user_two = Address::generate(&env);
+
+    initialize(&env, &client, user_one, randint(0, 10), randint(1, 10));
+
+    // second initialization should panic
+    initialize(&env, &client, user_two, randint(0, 10), randint(1, 10));
+}
+
+#[test]
+fn test_fails_with_empty_signer_set() {
```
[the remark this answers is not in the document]
deanamiel: just kept it for `test_initialize` and `test_transfer_operatorship` to avoid function name collusion

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -28,6 +31,65 @@ fn test_initialize() {
...
+    // transfer ownership to the new owner
+    client.transfer_ownership(&new_owner);
+
+    assert_invocation(
+        &env,
+        &initial_owner,
+        &client.address,
+        "transfer_ownership",
+        (new_owner.clone(),),
+    );
```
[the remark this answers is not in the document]
deanamiel: Doesn't line 85 check that the `transfer_ownership` event was emitted?

contracts/axelar-auth-verifier/src/test.rs
deanamiel: https://github.com/axelarnetwork/axelar-cgp-soroban/pull/8#discussion_r1518276754

contracts/axelar-auth-verifier/src/test.rs
deanamiel: changed

contracts/axelar-auth-verifier/src/test.rs
deanamiel: Makes sense but unless I'm wrong the Soroban sdk vec implementation doesn't have a lot of the functions found in the std library such as reverse, collect, etc:
https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.Vec.html

contracts/axelar-auth-verifier/src/test.rs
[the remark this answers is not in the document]
milapsheth: Ok

contracts/axelar-auth-verifier/src/test.rs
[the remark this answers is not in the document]
milapsheth: The method above only checks if the call was invoked from a certain sender. It doesn't check if the event was emitted on transfer

contracts/axelar-auth-verifier/src/test.rs
milapsheth: You can convert to/from std Vec. Add the alloc feature to the dependency and you can use some more methods

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -80,7 +245,164 @@ fn test_transfer_operatorship() {
...
+    let mut new_signers = generate_signer_set(&env, randint(1, 10));
+
+    let zero = U256::from_u32(&env, 0);
+    let one = U256::from_u32(&env, 1);
+
+    let total_weight = new_signers
+        .signer_set
+        .signers
+        .iter()
+        .fold(zero, |acc, (_, weight)| acc.add(&weight));
```
milapsheth, line 354: You can use reduce instead of fold to make it a bit cleaner

contracts/axelar-auth-verifier/src/testutils.rs
```diff
@@ -25,6 +25,16 @@ pub fn randint(a: u32, b: u32) -> u32 {
     rand::thread_rng().gen_range(a..b)
 }
 
+pub fn generate_random_payload_and_hash(env: &Env) -> BytesN<32> {
+    let payload = <soroban_sdk::BytesN<10> as soroban_sdk::testutils::BytesN<10>>::random(env);
```
milapsheth, line 29: look at existing usage or the docs to simplify code
```suggestion
    let payload = BytesN<10>::random(env);
```

contracts/axelar-auth-verifier/src/testutils.rs
```diff
@@ -9,7 +9,7 @@ use rand::rngs::OsRng;
 use rand::Rng;
 use secp256k1::{Message, PublicKey, Secp256k1, SecretKey};
 use sha3::{Digest, Keccak256};
-use soroban_sdk::{vec, U256};
+use soroban_sdk::{vec, Vec, U256};
 
 use soroban_sdk::{symbol_short, xdr::ToXdr, Address, Bytes, BytesN, Env};
```
milapsheth, line 15: import testutils BytesN as _ like we do elsewhere

contracts/axelar-auth-verifier/src/testutils.rs
```diff
@@ -70,6 +80,11 @@ pub fn generate_signer_set(env: &Env, num_signers: u32) -> TestSignerSet {
     }
 }
 
+#[allow(dead_code)]
```
milapsheth, line 83: dead_code?

milapsheth, line 83:
```suggestion
#[cfg(test)]
```
to scope it test visibility. alternatively, move it to test.rs where functions are defined for unit tests only (if you aren't using this method later outside this crate)

contracts/axelar-auth-verifier/src/testutils.rs
```diff
@@ -25,6 +25,16 @@ pub fn randint(a: u32, b: u32) -> u32 {
     rand::thread_rng().gen_range(a..b)
 }
 
+pub fn generate_random_payload_and_hash(env: &Env) -> BytesN<32> {
+    let payload = <soroban_sdk::BytesN<10> as soroban_sdk::testutils::BytesN<10>>::random(env);
+
+    let payload_bytes: Bytes = Bytes::from_slice(env, &payload.to_array());
```
milapsheth, line 31:
```suggestion
    let payload_bytes: Bytes = BytesN<10>::random(env).into();
```
probably should work

contracts/axelar-auth-verifier/src/test.rs
```diff
@@ -80,7 +255,165 @@ fn test_transfer_operatorship() {
...
+
+    let one = U256::from_u32(&env, 1);
+
+    let total_weight = new_signers
+        .signer_set
+        .signers
+        .iter()
+        .map(|(_, weight)| weight)
+        .reduce(|acc, weight| acc.add(&weight))
+        .expect("Empty signers");
```
milapsheth, line 365:
```suggestion
        .unwrap();
```
since a panic is never expected, unwrap is fine

contracts/axelar-auth-verifier/src/testutils.rs
deanamiel: Changed to [this](https://github.com/axelarnetwork/axelar-cgp-soroban/pull/8#discussion_r1524310085)

contracts/axelar-auth-verifier/src/testutils.rs
deanamiel: Changed to `#[cfg(test)]`

deanamiel: Moved function to `test.rs`

contracts/axelar-auth-verifier/src/test.rs
deanamiel: The problem is that during testing it did seem to panic with an empty signer set which I was explaining [above](https://github.com/axelarnetwork/axelar-cgp-soroban/pull/8#discussion_r1518268717). Let me know if I'm missing something here

contracts/axelar-auth-verifier/src/contract.rs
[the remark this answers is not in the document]
milapsheth: useful in case initialize isn't called. at the least we want to just do a pure `unwrap`, and we shouldn't default to some invalid value.

contracts/axelar-auth-verifier/src/test.rs
milapsheth: change the test generation so empty signer set can never be created

milapsheth (approved):

PR #18, opened 2024-06-24
fix(AxelarAuthVerifier): simplified initializer

re1ro:
* [x] `AxelarAuthVerifier`: refactoring `initialize` to take params in plain format


re1ro:
Closing this PR as I was able to encode nested struct for the original implementation:
https://github.com/axelarnetwork/axelar-contract-deployments/pull/280

PR #24, opened 2024-10-02
feat(axelar-operators): improve error handling

hydrobeam:
https://axelarnetwork.atlassian.net/browse/AXE-5707

# Summary

- renamed `Error` in the crate to `OperatorsError` for clarity
- Added two error variants to the  `OperatorsError`:
  - `AlreadyInitialized`/`NotInitialized`
- Moved `OperatorsError` to the interfaces crate
   - Deleted `error.rs` because of this
- annotated `AxelarOperatorsInterface` return types with `Result`
- removed all explicit panics/unwraps from `contract.rs`, instead returning appropriate error variants
   - `require_auth` and `invoke_contract` can still panic, but those are external 🤷 
- update tests to expect specific error types
  - added test to catch `NotInitialized` error and added checks for expected outcomes `test_initialize`


contracts/axelar-operators/src/contract.rs
```diff
@@ -82,17 +104,17 @@ impl AxelarOperatorsInterface for AxelarOperators {
...
         args: Vec<Val>,
-    ) -> Val {
+    ) -> Result<Val, OperatorError> {
         operator.require_auth();
 
         let key = DataKey::Operators(operator.clone());
 
         if !env.storage().persistent().has(&key) {
-            panic_with_error!(env, Error::NotAnOperator);
+            Err(OperatorError::NotAnOperator)?
```
ahramy, line 113: Just wondering, what is the best practice to use `?` ? From the Soroban examples, it uses `return Err( );` statement instead of `?` for error handling. i.e: https://github.com/stellar/soroban-examples/blob/e595983e7e7c30c2b745f0cdde1096319bb7c90c/account/src/lib.rs#L152

contracts/axelar-soroban-interfaces/src/axelar_operators.rs
```diff
@@ -22,5 +21,15 @@ pub trait AxelarOperatorsInterface {
         contract: Address,
         function_name: Symbol,
         args: Vec<Val>,
-    ) -> Val;
+    ) -> Result<Val, OperatorError>;
+}
+
+#[contracterror]
```
ahramy, line 27: How about moving this to the top, before the pub trait interface, for better code readability?

contracts/axelar-operators/src/contract.rs
milapsheth: yeah, a bit weird to use `?` in this manner. Create an `ensure` macro in the axelar-soroban-std library, something like:
```
#[macro_export]
macro_rules! ensure {
    ($cond:expr, $e:expr $(,)?) => {
        if !$cond {
            return Err($e);
        }
    };
}

# usage
ensure!(env.storage().persistent().has(&key), OperatorError::NotAnOperator);
```

contracts/axelar-soroban-interfaces/src/axelar_operators.rs
```diff
@@ -22,5 +21,15 @@ pub trait AxelarOperatorsInterface {
...
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum OperatorError {
```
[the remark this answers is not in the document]
hydrobeam: `Error` feels like a weird name for an error type, like it belongs in the standard library or something. `ContractError` would make more sense imo, but it's kind of a NIT regardless

contracts/axelar-operators/src/test.rs
```diff
@@ -257,3 +281,14 @@ fn fail_execute_when_target_panics() {
...
 }
+
+#[test]
+fn fail_on_uninitialized() {
+    let (env, _, client, _) = setup_env();
+
+    let operator = Address::generate(&env);
+    assert_eq!(
+        client.try_add_operator(&operator),
+        Err(Ok(OperatorError::NotInitialized))
```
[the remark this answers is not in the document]
hydrobeam: Yes. It's not explained very well in the docs, but soroban seems to transform functions which return `Result<A, Err>` to just return `A` (and panic on any error by default), and the way to actually handle errors is with the `try_*` version of the function. Returning `Result` just allows you to handle what kind of error is returned.

image

contracts/axelar-soroban-interfaces/src/axelar_operators.rs
hydrobeam: Sounds good

contracts/axelar-operators/src/contract.rs
hydrobeam: It probably compiles down to the same thing, but since `?` technically de-sugars down to a match statement I guess it's more appropriate to use `return` directly. No real reason for it, just a habit I've picked up.

contracts/axelar-operators/src/test.rs
```diff
@@ -36,21 +40,40 @@ fn setup_env<'a>() -> (Env, Address, AxelarOperatorsClient<'a>, Address) {
...
 #[test]
 fn test_initialize() {
-    let (env, _, client, _) = setup_env();
+    let (env, contract_id, client, _) = setup_env();
     let user = Address::generate(&env);
 
     client.initialize(&user);
+
+    assert!(env.as_contract(&contract_id, || {
+        env.storage()
```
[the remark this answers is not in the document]
hydrobeam: do you mean this: https://docs.rs/assert_ok/latest/assert_ok/ ? feels a little overkill to add a dependency for something like this. 

I added this test to make sure that `initialize()` is impacting the contract's state as intended. If we're just going to check that the function didn't error then it'd probably make more sense to remove it since all the other functions run run `initialize()` too?

contracts/axelar-soroban-interfaces/src/axelar_operators.rs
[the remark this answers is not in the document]
milapsheth: Yeah, `ContractError` is fine as well. it's scoped by the crate so at the end of the day we rely on it to separate namespaces

contracts/axelar-operators/src/test.rs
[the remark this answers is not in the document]
milapsheth: Ok, I vaguely remember this. Can you create an `assert_err!` macro to make this easier to use

packages/axelar-soroban-std/src/lib.rs
```diff
@@ -10,3 +10,16 @@ pub use testutils::*;
 pub mod traits;
 
 pub mod types;
+
+/// Return with an error if a condition is not met.
+///
+///
+/// Simplifies the pattern of checking for a condition and returning with an error.
+#[macro_export]
```
milapsheth, line 18: add this under a new `error.rs`

contracts/axelar-operators/src/test.rs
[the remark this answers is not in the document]
milapsheth: decided to use assert_ok (can copy or re-export it from soroban-std) when the test is meant to check if the given method didn't error. We can keep using .unwrap for helpers or methods that already have their own dedicated tests

contracts/axelar-operators/src/contract.rs
```diff
@@ -1,43 +1,53 @@
...
+            .instance()
+            .get(&DataKey::Owner)
+            .ok_or(OperatorError::NotInitialized)?;
+
         owner.require_auth();
 
         env.storage().instance().set(&DataKey::Owner, &new_owner);
 
         event::transfer_ownership(&env, owner, new_owner);
+        Ok(())
```
milapsheth, line 25:
```suggestion

                Ok(())
```

contracts/axelar-operators/src/test.rs
```diff
@@ -151,8 +152,10 @@ fn fail_add_operator_duplicate() {
     client.add_operator(&operator);
 
     // set existing operator as an operator, should panic
-    let res = client.try_add_operator(&operator);
-    assert_eq!(res, Err(Ok(OperatorError::OperatorAlreadyAdded)));
+    assert_err!(
+        client.try_add_operator(&operator),
+        Ok(OperatorError::OperatorAlreadyAdded)
```
milapsheth, line 157: having `Ok` looks weird, can you remove the need for that

packages/axelar-soroban-std/src/error.rs
```diff
@@ -0,0 +1,74 @@
...
+            core::option::Option::None => {
+                panic!("Expected value when calling {}, got None", stringify!($x));
+            }
+        }
+    };
+}
+
+/// Assert that a [`Result`] is [`Err`] and matches a desired error
+#[macro_export]
+macro_rules! assert_err {
```
milapsheth, line 57:
```suggestion
macro_rules! assert_contract_err {
```
Since we want to make it specific to checking wrapped Soroban errors

packages/axelar-soroban-std/src/error.rs
```diff
@@ -52,23 +54,36 @@ macro_rules! assert_some {
...
+/// The function needs to fail, and successfully pass down the intended error type.
+/// So, for the assert to succeed, the parameters would be in the form:
+///
+/// given: `Err(Ok(ContractError))`
+/// expected: `ContractError`
+///
+/// Putting it together in a function call:
+///
+/// `assert_contract_err(client.try_fun(...), ContractError);`
+pub fn assert_contract_err<A, B, C, D>(given: Result<Result<A, B>, Result<C, D>>, expected: C)
```
milapsheth, line 69: why not make it a macro? i'd prefer that for consistency

milapsheth (approved):

packages/axelar-soroban-std/src/error.rs
hydrobeam: I was thinking I wanted to enforce the type being passed into it and that a function would somewhat be self-documenting in that sense and be more helpful when used incorrectly. But I suppose it doesn't particularly matter either way, and the docstring should help with understanding usage. Rewrote it as a macro.

ahramy (approved):

PR #25, opened 2024-10-07
feat(axelar-gateway): increase gateway coverage and add coverage report

ahramy:
[AXE-5747](https://axelarnetwork.atlassian.net/browse/AXE-5747)

[AXE-5747]: https://axelarnetwork.atlassian.net/browse/AXE-5747?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,9 +32,7 @@ cargo llvm-cov
 ## Coverage Report
 
 ```bash
-cargo install cargo-tarpaulin
-cargo tarpaulin --out Html
-open tarpaulin-report.html
+cargo llvm-cov --html or cargo llvm-cov --open
 ```
 
 ## Deploy
--- a/README.md
+++ b/README.md
@@ -31,15 +31,23 @@ cargo llvm-cov
 
 ## Coverage Report
 
+### Generate report
+
 ```bash
-cargo llvm-cov --html # or cargo llvm-cov --open
+cargo llvm-cov --html
+```
+
+### Generate and Open report
+
+```bash
+cargo llvm-cov --open
 ```
 
 ## Deploy
 
-- Check configuration for CLI and Identity before deployment: https://developers.stellar.org/docs/build/smart-contracts/getting-started/setup
+> Check configuration for CLI and Identity before deployment: https://developers.stellar.org/docs/build/smart-contracts/getting-started/setup
 
-#### Build, Optimize and Deploy contract:
+### Build, Optimize and Deploy contract:
 
 ```bash
 soroban contract build
--- a/README.md
+++ b/README.md
@@ -32,7 +32,7 @@ cargo llvm-cov
 ## Coverage Report
 
 ```bash
-cargo llvm-cov --html or cargo llvm-cov --open
+cargo llvm-cov --html # or cargo llvm-cov --open
 ```
 
 ## Deploy
--- a/README.md
+++ b/README.md
@@ -29,6 +29,14 @@ cargo install cargo-llvm-cov
 cargo llvm-cov
 ```
 
+## Coverage Report
+
+```bash
+cargo install cargo-tarpaulin
+cargo tarpaulin --out Html
+open tarpaulin-report.html
+```
+
 ## Deploy
 
 - Check configuration for CLI and Identity before deployment: https://developers.stellar.org/docs/build/smart-contracts/getting-started/setup
--- a/contracts/axelar-gateway/src/test.rs
+++ b/contracts/axelar-gateway/src/test.rs
@@ -276,3 +276,30 @@ fn rotate_signers() {
         (source_chain, source_address),
     );
 }
+
+#[test]
+fn transfer_operatorship() {
+    let (env, contract_id, client) = setup_env();
+    let operator = Address::generate(&env);
+    let new_operator = Address::generate(&env);
+
+    initialize(&env, &client, operator.clone(), 1, randint(1, 10));
+
+    assert_eq!(client.operator(), operator);
+
+    client.transfer_operatorship(&new_operator);
+
+    assert_emitted_event(
+        &env,
+        -1,
+        &contract_id,
+        (
+            String::from_str(&env, "transferred"),
+            operator.clone(),
+            new_operator.clone(),
+        ),
+        (),
+    );
+
+    assert_eq!(client.operator(), new_operator);
+}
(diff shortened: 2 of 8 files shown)


README.md
```diff
@@ -29,6 +29,14 @@ cargo install cargo-llvm-cov
 cargo llvm-cov
 ```
 
+## Coverage Report
+
+```bash
+cargo install cargo-tarpaulin
```
hydrobeam, line 35: How does `cargo-tarpaulin` compare to `cargo llvm-cov`? do we need both?

Also just curious how you came to settle on `tarpaulin` / what else was considered. For example, a quick search led me to https://github.com/mozilla/grcov which seems to be what mozilla uses for firefox 🤔

ahramy: It's a great point. `cargo tarpaulin` was the first tool I found when searching for Rust coverage tools. After doing some further research, there's no need to add additional tools. We can just use `cargo llvm-cov --html` or `cargo llvm-cov --open`. `grcov` is also another option, but it requires more customized configurations.

README.md
```diff
@@ -29,6 +29,12 @@ cargo install cargo-llvm-cov
 cargo llvm-cov
 ```
 
+## Coverage Report
+
+```bash
+cargo llvm-cov --html or cargo llvm-cov --open
```
hydrobeam, line 35:
```suggestion
cargo llvm-cov --html # or cargo llvm-cov --open
```
making this a comment would be useful for copy-pasting

README.md
```diff
@@ -29,6 +29,12 @@ cargo install cargo-llvm-cov
...
+```
+
```
hydrobeam, line 37:
```suggestion
The finished report can be found at `target/llvm-cov/html/`
```

the report tells you where you can find it when it's done, but I think it'd also be nice to have it here

ahramy: Since the `cargo llvm-cov --html` command will display `Finished report saved to /Users/user_path/axelar-cgp-soroban/target/llvm-cov/html` after it's completion, I think it's not necessary

README.md
ahramy: Good call, updated.

contracts/axelar-gateway/src/test.rs
```diff
@@ -276,3 +276,65 @@ fn rotate_signers() {
...
+            invoke: &MockAuthInvoke {
+                contract: &contract_id,
+                fn_name: "transfer_operatorship",
+                args: (&new_operator,).into_val(&env),
+                sub_invokes: &[],
+            },
+        }])
+        .transfer_operatorship(&new_operator);
+
+    assert_emitted_event(
```
milapsheth, line 302: can you create a `assert_last_emitted_event` helper? don't need to pass a weird looking `-1` everytime

contracts/axelar-gateway/src/test.rs
```diff
@@ -276,3 +276,65 @@ fn rotate_signers() {
...
+    let operator = Address::generate(&env);
+    let new_operator = Address::generate(&env);
+
+    initialize(&env, &client, operator.clone(), 1, randint(1, 10));
+
+    assert_eq!(client.operator(), operator);
+
+    client
+        .mock_auths(&[MockAuth {
+            address: &operator,
```
milapsheth, line 292: :+1:

contracts/axelar-gateway/src/test.rs
```diff
@@ -276,3 +276,65 @@ fn rotate_signers() {
...
+            invoke: &MockAuthInvoke {
+                contract: &contract_id,
+                fn_name: "transfer_operatorship",
+                args: (&new_operator,).into_val(&env),
```
milapsheth, line 296: for some methods, we only need to test that the caller is a certain address, so it's a bit unnecessary to have to specify the full invocation everytime. Maybe there's a less brittle way to do this (for e.g. this will break on refactoring, if the function name changes. You could ask in Stellar discord perhaps

README.md
```diff
@@ -29,6 +29,12 @@ cargo install cargo-llvm-cov
 cargo llvm-cov
 ```
 
+## Coverage Report
+
+```bash
+cargo llvm-cov --html # or cargo llvm-cov --open
```
milapsheth, line 35:
```suggestion
# Generate report
cargo llvm-cov --html

# Generate and Open report
cargo llvm-cov --open
```

milapsheth (approved):

milapsheth:
add scope to PR title `feat(axelar-gateway): ...`

contracts/axelar-gateway/src/test.rs
ahramy: Agreed. created a `assert_last_emitted_event ` helper and updated all related unit tests.

README.md
ahramy: updated.

contracts/axelar-gateway/src/test.rs
ahramy: I agree that for some methods, we don't need to specify the full invocation every time. But for the `transfer_operatorship` unit test, it specifically checks whether the caller is `operator` or not, so I added two cases: (1) only the operator can transfer Operatorship (2) non-operator cannot transfer Operatorship. (Unauthorized). I believe this should break if the function name changes during refactoring. Let me know if you still think we need to remove the mocking of auth for operator in `transfer_operatorship` unit test.

fyi I followed this step to test the sender is authorized: https://discord.com/channels/897514728459468821/1265827786111586344/1292434336712495114

milapsheth: Yeah, testing that the caller is the operator is good. I just don't like the dependency on the function name which can break due to refactoring. The actual contract logic doesn't care about the function name. If there's no way to leave the function name, and args unspecified, perhaps we can create a macro that wraps the call to transfer_operatorship and the caller account we provide, and adds the mock auth on top for us. You can investigate as a separate task

milapsheth (approved):

contracts/axelar-gateway/src/test.rs
ahramy: Got it. I think its worth researching on it. I will investigate as a separate task.

PR #26, opened 2024-10-08
feat(gas-service): improve error handling

hydrobeam:
https://axelarnetwork.atlassian.net/browse/AXE-5773

## Summary

- added two new error variants: `NotInitialized` and `AlreadyInitialized`
- moved error type to interface file
- replaced `if !cond { panic_with_error!()}` pattern with `ensure!()`
- modified `contract.rs` to use errors without explicit panics 
- updated the interface to return `Result<(), GasServiceError>` types
- added tests to check for error cases
   - some test duplication here but I think it's fine


contracts/axelar-gas-service/src/test.rs
```diff
@@ -22,14 +26,87 @@ fn setup_env<'a>() -> (Env, Address, Address, AxelarGasServiceClient<'a>) {
...
+
+    assert_contract_err!(
+        client.try_initialize(&gas_collector),
+        GasServiceError::AlreadyInitialized
+    );
+}
+
+#[test]
+fn fail_pay_gas_invalid_amount() {
+    let (env, contract_id, gas_collector, client) = setup_env();
```
ahramy, line 69: running `cargo test` before submitting PR, it will catch unused variables

ahramy: I also run those before submitting PR 
```
cargo fmt --all -- --check;
cargo clippy -- -D warnings -A deprecated;
```

contracts/axelar-gas-service/src/test.rs
```diff
@@ -22,14 +26,87 @@ fn setup_env<'a>() -> (Env, Address, Address, AxelarGasServiceClient<'a>) {
...
+            &token,
+        ),
+        GasServiceError::InvalidAmount
+    );
+}
+
 #[test]
 fn pay_gas_for_contract_call() {
-    let (env, contract_id, _, client) = setup_env();
+    let (env, contract_id, gas_collector, client) = setup_env();
```
ahramy, line 109: same here, unused variable, keep it as `_`

contracts/axelar-gas-service/src/test.rs
```diff
@@ -81,9 +159,57 @@ fn pay_gas_for_contract_call() {
...
 
+#[test]
+fn fail_collect_fees_invalid_amount() {
+    let (env, contract_id, gas_collector, client) = setup_env();
+    client.initialize(&gas_collector);
+
+    let token_address: Address = env.register_stellar_asset_contract(Address::generate(&env));
+
+    let supply: i128 = 1000;
+    // refund_amount <= 0
```
ahramy, line 170: Maybe update comment (adding more explanation) or remove it?

contracts/axelar-gas-service/src/test.rs
```diff
@@ -81,9 +159,57 @@ fn pay_gas_for_contract_call() {
...
+}
+
+#[test]
+fn fail_collect_fees_insufficient_balance() {
+    let (env, contract_id, gas_collector, client) = setup_env();
+    client.initialize(&gas_collector);
+
+    let token_address: Address = env.register_stellar_asset_contract(Address::generate(&env));
+
+    // supply < refund_amount
```
ahramy, line 192: Maybe update comment (adding more explanation) or remove it?

contracts/axelar-gas-service/src/test.rs
```diff
@@ -22,14 +26,87 @@ fn setup_env<'a>() -> (Env, Address, Address, AxelarGasServiceClient<'a>) {
 
     let client = AxelarGasServiceClient::new(&env, &contract_id);
     let gas_collector: Address = Address::generate(&env);
-    client.initialize(&gas_collector);
```
ahramy, line 25: Not sure why you move `client.initialize` outside of setup_env(). Does it have to be moved out of the function?

milapsheth, line 25: instead of removing this, duplicate part of the setup solely for test_initialize. All other tests will need to have an initialized contract anyways

contracts/axelar-gas-service/src/test.rs
```diff
@@ -22,14 +26,87 @@ fn setup_env<'a>() -> (Env, Address, Address, AxelarGasServiceClient<'a>) {
...
+#[test]
+fn test_initialize() {
+    let (env, contract_id, gas_collector, client) = setup_env();
+    client.initialize(&gas_collector);
+
+    assert!(env.as_contract(&contract_id, || {
+        env.storage()
+            .instance()
+            .get::<DataKey, bool>(&DataKey::Initialized)
+            .unwrap()
```
milapsheth, line 42: use assert_ok instead of unwrap this isn't just helper code but what we're testing for

contracts/axelar-gas-service/src/test.rs
```diff
@@ -22,14 +26,87 @@ fn setup_env<'a>() -> (Env, Address, Address, AxelarGasServiceClient<'a>) {
...
+    }));
+
+    assert_eq!(
+        env.as_contract(&contract_id, || {
+            env.storage()
+                .instance()
+                .get::<DataKey, Address>(&DataKey::GasCollector)
+                .unwrap()
```
milapsheth, line 50: same

contracts/axelar-gas-service/src/test.rs
```diff
@@ -22,14 +26,87 @@ fn setup_env<'a>() -> (Env, Address, Address, AxelarGasServiceClient<'a>) {
...
+        client.try_initialize(&gas_collector),
+        GasServiceError::AlreadyInitialized
+    );
+}
+
+#[test]
+fn fail_pay_gas_invalid_amount() {
+    let (env, contract_id, gas_collector, client) = setup_env();
+
+    let token_address: Address = env.register_stellar_asset_contract(Address::generate(&env));
```
milapsheth, line 71:
```suggestion
    let asset = StellarAssetClient::new (env.register_stellar_asset_contract(Address::generate(&env)));
```
then you can reuse it later, and get the address via `asset.address`

contracts/axelar-gas-service/src/test.rs
milapsheth: we could also standardize `setup_env` and `setup_and_initialize` as separate helpers

contracts/axelar-gas-service/src/test.rs
hydrobeam: added an explanation of the test as a docstring and removed the comment

contracts/axelar-gas-service/src/test.rs
hydrobeam: added an explanation of the test as a docstring and removed the comment

contracts/axelar-gas-service/src/test.rs
hydrobeam: done

contracts/axelar-gas-service/src/test.rs
hydrobeam: good point, moved it back. I think I was planning to test `NotInitialized`  (so needed to setup without initializing), but didn't get around to it. Even then it's worth just duplicating the preamble for the one test instead

contracts/axelar-gas-service/src/test.rs
hydrobeam: done in [dc05398](https://github.com/axelarnetwork/axelar-cgp-soroban/pull/26/commits/dc05398e7b4a479295fa4a497815d0c2ba163fba).  the function `register_stellar_asset_contract_v2` returns a `StellarAssetClient` and `register_stellar_asset_contract` is deprecated in favour of `v2`, so I used that instead for these changes.

contracts/axelar-gas-service/src/test.rs
hydrobeam: my thinking was that it's not enough to just test for `Ok(_)` but also that the `DataKey::Initialized` is set to `true`, hence the `unwrap` to pull the value out of the `Option`. could maybe do something like this to avoid the unwrap:
```rs
    assert_eq!(
        env.as_contract(&contract_id, || {
            env.storage()
                .instance()
                .get::<DataKey, bool>(&DataKey::Initialized)
        }),
        Some(true)
    );
```

but it's slightly janky. Would you want to just test for existence instead of ensuring the values match?

milapsheth: Add an `assert_some!` macro that similar to `assert_ok` tests that it's a `Some(value)` and returns the unwrapped value, so we can still test if the output is `true`

contracts/axelar-gas-service/src/test.rs
```diff
@@ -81,15 +159,60 @@ fn pay_gas_for_contract_call() {
     );
 }
 
+#[test]
+/// Ensure failure occurs when refund_amount <= 0
+fn fail_collect_fees_invalid_amount() {
```
milapsheth, line 164:
```suggestion
fn fail_collect_fees_zero_refund_amount() {
```
instead of adding comments, make the test name more descriptive

milapsheth:
merge `main` on this branch to resolve merge conflicts

milapsheth (approved):

contracts/axelar-gas-service/src/test.rs
```diff
@@ -82,22 +82,18 @@ fn fail_not_initialized() {
...
-                .unwrap()
-        }),
-        gas_collector
-    );
+    let gc_addr = assert_some!(env.as_contract(&contract_id, || {
+        env.storage()
+            .instance()
+            .get::<DataKey, Address>(&DataKey::GasCollector)
+    }));
+    assert_eq!(gc_addr, gas_collector);
```
milapsheth, line 102: `gc_addr` -> `stored_gas_collector` or something similar. Avoid abbreviations

ahramy (approved):

PR #27, opened 2024-10-08
feat(gateway): add rotation delay governance

ahramy:
[AXE-5723](https://axelarnetwork.atlassian.net/browse/AXE-5723)

- TODO: add docstring about how `bypass_rotation_delay` supposed to be used.

[AXE-5723]: https://axelarnetwork.atlassian.net/browse/AXE-5723?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/axelar-gateway/src/auth.rs
+++ b/contracts/axelar-gateway/src/auth.rs
@@ -1,48 +1,10 @@
 use axelar_soroban_interfaces::types::{ProofSignature, ProofSigner, WeightedSigner};
-use soroban_sdk::{crypto::Hash, panic_with_error, xdr::ToXdr, Bytes, BytesN, Env, Vec};
+use soroban_sdk::{crypto::Hash, panic_with_error, xdr::ToXdr, Bytes, BytesN, Env};
 
 use crate::error::AuthError;
 use crate::event;
 use crate::storage_types::DataKey;
 use axelar_soroban_interfaces::types::{Proof, WeightedSigners};
-
-pub fn initialize_auth(
-    env: Env,
-    domain_separator: BytesN<32>,
-    minimum_rotation_delay: u64,
-    previous_signers_retention: u64,
-    initial_signers: Vec<WeightedSigners>,
-) {
-    if env.storage().instance().has(&DataKey::AuthInitialized) {
-        panic!("Already initialized");
-    }
-
-    env.storage().instance().set(&DataKey::Initialized, &true);
-
-    env.storage().instance().set(&DataKey::Epoch, &0_u64);
-
-    // TODO: Do we need to manually expose these in a query, or can it be read directly off of storage in Stellar?
-    env.storage().instance().set(
-        &DataKey::PreviousSignerRetention,
-        &previous_signers_retention,
-    );
-
-    env.storage()
-        .instance()
-        .set(&DataKey::DomainSeparator, &domain_separator);
-
-    env.storage()
-        .instance()
-        .set(&DataKey::MinimumRotationDelay, &minimum_rotation_delay);
-
-    if initial_signers.is_empty() {
-        panic_with_error!(env, AuthError::InvalidSigners);
-    }
-
-    for signers in initial_signers.into_iter() {
-        rotate_signers(&env, &signers, false);
-    }
-}
 
 pub fn validate_proof(env: &Env, data_hash: BytesN<32>, proof: Proof) -> bool {
     let signers_set = proof.weighted_signers();
@@ -105,7 +67,7 @@ pub fn rotate_signers(env: &Env, new_signers: &WeightedSigners, enforce_rotation
         &new_epoch,
     );
 
-    event::rotate_signers(env, new_signers_hash, new_signers.clone());
+    event::rotate_signers(env, new_signers.clone());
 }
 
 fn message_hash_to_sign(env: &Env, signers_hash: BytesN<32>, data_hash: BytesN<32>) -> Hash<32> {
--- a/contracts/axelar-gateway/src/auth.rs
+++ b/contracts/axelar-gateway/src/auth.rs
@@ -1,8 +1,8 @@
 use axelar_soroban_interfaces::types::{ProofSignature, ProofSigner, WeightedSigner};
-use soroban_sdk::xdr::ToXdr;
-use soroban_sdk::{crypto::Hash, panic_with_error, Bytes, BytesN, Env, Vec};
+use soroban_sdk::{crypto::Hash, panic_with_error, xdr::ToXdr, Bytes, BytesN, Env, Vec};
 
 use crate::error::AuthError;
+use crate::event;
 use crate::storage_types::DataKey;
 use axelar_soroban_interfaces::types::{Proof, WeightedSigners};
 
@@ -10,7 +10,7 @@ pub fn initialize_auth(
     env: Env,
     domain_separator: BytesN<32>,
     minimum_rotation_delay: u64,
-    previous_signer_retention: u64,
+    previous_signers_retention: u64,
     initial_signers: Vec<WeightedSigners>,
 ) {
     if env.storage().instance().has(&DataKey::AuthInitialized) {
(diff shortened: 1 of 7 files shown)


contracts/axelar-gateway/src/auth.rs
```diff
@@ -1,16 +1,16 @@
...
 
 pub fn initialize_auth(
     env: Env,
     domain_separator: BytesN<32>,
     minimum_rotation_delay: u64,
-    previous_signer_retention: u64,
+    previous_signers_retention: u64,
     initial_signers: Vec<WeightedSigners>,
 ) {
     if env.storage().instance().has(&DataKey::AuthInitialized) {
```
milapsheth, line 16: AuthInitialized is different than Initialized key below

milapsheth, line 16: also auth initialize isn't needed anymore since gateway itself should be initialized and it's the same contract

contracts/axelar-gateway/src/contract.rs
```diff
@@ -173,13 +173,18 @@ impl AxelarGatewayInterface for AxelarGateway {
         env: Env,
         signers: axelar_soroban_interfaces::types::WeightedSigners,
         proof: axelar_soroban_interfaces::types::Proof,
+        enforce_rotation_delay: bool,
```
milapsheth, line 176: Nice, I guess this is a bit better than having to pass in the caller everytime. It does allow the operator to also enforce rotation delay as well so slightly different behaviour than EVM, but makes sense. Add docstring on how this is supposed to be used. It's not clear currently that not just anyone can provide a `false` value

contracts/axelar-gateway/src/contract.rs
```diff
@@ -173,13 +173,18 @@ impl AxelarGatewayInterface for AxelarGateway {
...
     ) {
         let data_hash: BytesN<32> = env
             .crypto()
             .keccak256(&(CommandType::RotateSigners, signers.clone()).to_xdr(&env))
             .into();
 
-        // TODO: Add rotation delay governance
+        if enforce_rotation_delay {
```
milapsheth, line 183:
```suggestion
        if !enforce_rotation_delay {
```
operator is required when delay is not enforced. Double check the tests/behaviour with EVM to help catch this kind of issues

contracts/axelar-gateway/src/contract.rs
```diff
@@ -173,13 +173,18 @@ impl AxelarGatewayInterface for AxelarGateway {
...
+            let operator: Address = Self::operator(&env);
+            operator.require_auth();
```
milapsheth, line 185:
```suggestion
            Self::operator(&env).require_auth();
```

contracts/axelar-gateway/src/event.rs
```diff
@@ -30,9 +30,9 @@ pub(crate) fn execute_contract_call(env: &Env, message_id: String) {
     env.events().publish(topics, ());
 }
 
-pub(crate) fn rotate_signers(env: &Env, signers: WeightedSigners) {
+pub(crate) fn rotate_signers(env: &Env, signers_hash: BytesN<32>, signers: WeightedSigners) {
```
milapsheth, line 33: change the event in a separate PR for clarity since this is an external dependency in Amplifier and relayer

contracts/axelar-gateway/src/testutils.rs
```diff
@@ -58,6 +58,10 @@ pub fn get_rotation_hash(env: &Env, new_signers: WeightedSigners) -> BytesN<32>
         .into()
 }
 
+pub fn get_signers_hash(env: &Env, new_signers: WeightedSigners) -> BytesN<32> {
```
milapsheth, line 61: define a method on WeightedSigners type instead

contracts/axelar-gateway/src/testutils.rs
```diff
@@ -20,27 +20,27 @@ const DESTINATION_ADDRESS: &str = "0x4EFE356BEDeCC817cb89B4E9b796dB8bC188DC59";
 #[derive(Clone, Debug)]
 pub struct TestSignerSet {
     pub signers: std::vec::Vec<SigningKey>,
-    pub signer_set: WeightedSigners,
+    pub signers_set: WeightedSigners,
```
milapsheth, line 23: Should be `signer_set` since it's a single set of signers, and not a `Vec<WeightedSigners>`

Alternatively, call this `signers` and change existing `signers` to `signer_keys`

ahramy: updated `signers_set: WeightedSigners` to `signers`  and `signers: std::vec::Vec<SigningKey>` to `signer_keys`

contracts/axelar-gateway/src/testutils.rs
ahramy: this is related to the `rotate_signers` event, which will be in a separate PR.

contracts/axelar-gateway/src/contract.rs
ahramy: I think we need to add docstrings everywhere in the soroban repo. I added as `TODO` in the code, and will add it in a separate PR later.

contracts/axelar-gateway/src/testutils.rs
```diff
@@ -31,19 +31,19 @@ pub fn initialize(
     previous_signers_retention: u32,
     num_signers: u32,
 ) -> TestSignerSet {
-    let signers = generate_signers_set(env, num_signers, BytesN::random(env));
-    let signers_set = vec![&env, signers.signers_set.clone()];
+    let signer_set = generate_signers_set(env, num_signers, BytesN::random(env));
+    let signers = vec![&env, signer_set.signers.clone()];
```
milapsheth, line 35: confusing to use signers for a list and the individual weighted signers. Instead create a vector in the call below so you don't need to name it?

contracts/axelar-gateway/src/test.rs
```diff
@@ -277,28 +275,32 @@ fn rotate_signers_with_enforce_rotation_delay() {
...
-    let signers_hash = get_signers_hash(&env, new_signers.signers_set.clone());
-    let data_hash = get_rotation_hash(&env, new_signers.signers_set.clone());
-    let proof = generate_proof(&env, data_hash.clone(), signers);
+    let data_hash = get_rotation_hash(&env, new_signers.signers.clone());
+    let proof = generate_proof(&env, data_hash.clone(), signers.clone());
+    let enforce_rotation_delay = true;
 
     client
         .mock_auths(&[MockAuth {
             address: &operator,
```
milapsheth, line 284: the caller doesn't matter if rotation delay is `true`. This test case should be for when the delay is being bypassed

contracts/axelar-gateway/src/contract.rs
```diff
@@ -169,20 +186,20 @@ impl AxelarGatewayInterface for AxelarGateway {
...
     }
 
+    // TODO: add docstring about how enforce_rotation_delay supposed to be used.
     fn rotate_signers(
         env: Env,
-        signers: axelar_soroban_interfaces::types::WeightedSigners,
-        proof: axelar_soroban_interfaces::types::Proof,
+        signers: WeightedSigners,
+        proof: Proof,
         enforce_rotation_delay: bool,
```
milapsheth, line 194:
```suggestion
        skip_rotation_delay: bool,
```
or `bypass_rotation_delay` or something similar that's opposite of enforcing the delay (which should be the default behaviour). While the implementation changed compared to the reference, a lack of name change here seems to be causing confusion

contracts/axelar-gateway/src/contract.rs
```diff
@@ -169,20 +186,20 @@ impl AxelarGatewayInterface for AxelarGateway {
         }
     }
 
+    // TODO: add docstring about how enforce_rotation_delay supposed to be used.
```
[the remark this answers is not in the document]
ahramy: I see. address it in this PR.

milapsheth (approved):

contracts/axelar-gateway/src/contract.rs
```diff
@@ -169,17 +169,22 @@ impl AxelarGatewayInterface for AxelarGateway {
         }
     }
 
+    // TODO: add docstring about how bypass_rotation_delay supposed to be used.
```
milapsheth, line 172: doc is missing

PR #29, opened 2024-10-15
feat: update execute event

ahramy:
[AXE-6048](https://axelarnetwork.atlassian.net/browse/AXE-6048)

[AXE-6048]: https://axelarnetwork.atlassian.net/browse/AXE-6048?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/axelar-gateway/src/event.rs
+++ b/contracts/axelar-gateway/src/event.rs
@@ -15,25 +15,13 @@ pub(crate) fn call_contract(
 }
 
 pub(crate) fn approve_message(env: &Env, message: Message) {
-    let topics = (
-        symbol_short!("approved"),
-        message.message_id,
-        message.contract_address,
-        message.payload_hash,
-    );
-    env.events()
-        .publish(topics, (message.source_chain, message.source_address));
+    let topics = (symbol_short!("approved"),);
+    env.events().publish(topics, message);
 }
 
 pub(crate) fn execute_contract_call(env: &Env, message: Message) {
-    let topics = (
-        symbol_short!("executed"),
-        message.message_id,
-        message.contract_address,
-        message.payload_hash,
-    );
-    env.events()
-        .publish(topics, (message.source_chain, message.source_address));
+    let topics = (symbol_short!("executed"),);
+    env.events().publish(topics, message);
 }
 
 pub(crate) fn rotate_signers(env: &Env, signers: WeightedSigners) {
--- a/contracts/axelar-gateway/src/event.rs
+++ b/contracts/axelar-gateway/src/event.rs
@@ -25,9 +25,15 @@ pub(crate) fn approve_message(env: &Env, message: Message) {
         .publish(topics, (message.source_chain, message.source_address));
 }
 
-pub(crate) fn execute_contract_call(env: &Env, message_id: String) {
-    let topics = (symbol_short!("executed"), message_id);
-    env.events().publish(topics, ());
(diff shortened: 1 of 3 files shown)


contracts/axelar-gateway/src/event.rs
```diff
@@ -25,9 +25,15 @@ pub(crate) fn approve_message(env: &Env, message: Message) {
         .publish(topics, (message.source_chain, message.source_address));
 }
 
-pub(crate) fn execute_contract_call(env: &Env, message_id: String) {
-    let topics = (symbol_short!("executed"), message_id);
-    env.events().publish(topics, ());
+pub(crate) fn execute_contract_call(env: &Env, message: Message) {
+    let topics = (
+        symbol_short!("executed"),
+        message.message_id,
```
milapsheth, line 31: Shall we index the `source_chain`, `message_id`, and `payload_hash` for both approved/executed events instead in the topic? @sdaveas

contracts/axelar-gateway/src/test.rs
```diff
@@ -168,8 +168,13 @@ fn approve_message() {
...
-        (symbol_short!("executed"), message_id.clone()),
-        (),
+        (
+            symbol_short!("executed"),
+            message_id.clone(),
+            contract_address.clone(),
+            payload_hash.clone(),
+        ),
+        (source_chain.clone(), source_address.clone()),
     );
```
milapsheth, line 178: test @axelarnetwork/sniper

contracts/axelar-gateway/src/event.rs
milapsheth, line 31: @ahramy As per Stellar [docs](https://developers.stellar.org/docs/build/guides/events/publish), event topics can't contain arbitrarily length data such as a Vec, Bytes etc. (similar to solidity). They don't mention `String` there but I'm wondering if that could potentially cause an issue (in the future, since it seems to work atm).

ahramy: @milapsheth I asked a question and got an answer, confirmed that there's no limits for topic types.

> dmytro_stellar: sorry, the doc seems incorrect/outdated. there are no limits for topic types or even count. the only limitation is that a transaction can't emit more than 8 KB of events combined. that said if you have too many/too complex topics, in theory you might have issues with indexers. that said, I don't know if there are any such restrictions in reality

milapsheth: Since they said
> there are no limits for topic types or even count.
let's use the whole message for the topic?

@sdaveas Will you need the message fields to separated in the topics tuple, or a single element tuple with the full message is fine?

sdaveas: it doesn't make a big difference for us because we track every event emitted by the contract, and we don't use topic attributes in the event filter. We want to consume the whole event to construct the whole message, so if the message is partially in topic and values of the event, that's fine. Ideally it'd be the same for approve and execute so that we can reuse our parsing code, but if there's a good reason to diverse, that's fine also.

milapsheth (approved):

contracts/axelar-gateway/src/event.rs
ahramy: updated to have the full message data which would be more simpler for both approved/executed events.

PR #32, opened 2024-10-17
feat(gateway): add upgrade and version

ahramy:
[AXE-6063](https://axelarnetwork.atlassian.net/browse/AXE-6063)

[AXE-6063]: https://axelarnetwork.atlassian.net/browse/AXE-6063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/axelar-gateway/src/contract.rs
+++ b/contracts/axelar-gateway/src/contract.rs
@@ -1,7 +1,7 @@
 use axelar_soroban_interfaces::types::{Message, Proof};
 use soroban_sdk::xdr::ToXdr;
 use soroban_sdk::{
-    contract, contractimpl, panic_with_error, Address, Bytes, BytesN, Env, String, Vec,
+    contract, contractimpl, panic_with_error, vec, Address, Bytes, BytesN, Env, String, Vec,
 };
 
 use crate::storage_types::{DataKey, MessageApprovalKey, MessageApprovalValue};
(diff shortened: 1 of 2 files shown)


contracts/axelar-gateway/src/contract.rs
```diff
@@ -223,6 +223,16 @@ impl AxelarGatewayInterface for AxelarGateway {
     fn epoch(env: &Env) -> u64 {
         auth::epoch(env)
     }
+
+    fn version(env: Env) -> Vec<u32> {
+        vec![&env, 0, 0, 1]
+    }
+
+    fn upgrade(env: Env, new_wasm_hash: BytesN<32>) {
+        Self::operator(&env).require_auth();
```
milapsheth, line 232: we need to add a separate owner role, similar to EVM gateway. can add in a separate PR though

contracts/axelar-gateway/src/contract.rs
```diff
@@ -223,6 +223,16 @@ impl AxelarGatewayInterface for AxelarGateway {
     fn epoch(env: &Env) -> u64 {
         auth::epoch(env)
     }
+
+    fn version(env: Env) -> Vec<u32> {
+        vec![&env, 0, 0, 1]
```
milapsheth, line 228: we should get these from cargo env
https://github.com/axelarnetwork/axelar-amplifier/blob/main/contracts/axelarnet-gateway/src/contract.rs#L18
you can get the major minor individually as well i believe

ahramy: I was able to get:
```
axelar_gateway::contract 
const CONTRACT_VERSION: &str = "0.1.0"
```
updated.

milapsheth (approved):

PR #36, opened 2024-10-21
feat(gateway): prevent repeated signers

ahramy:
[AXE-6096](https://axelarnetwork.atlassian.net/browse/AXE-6096)

### DIFF
--- a/contracts/axelar-gateway/src/auth.rs
+++ b/contracts/axelar-gateway/src/auth.rs
@@ -43,11 +43,7 @@ pub fn validate_proof(env: &Env, data_hash: BytesN<32>, proof: Proof) -> bool {
 
     let signers_hash: BytesN<32> = env.crypto().keccak256(&signers_set.to_xdr(env)).into();
 
-    let signers_epoch: u64 = env
-        .storage()
-        .persistent()
-        .get(&DataKey::EpochBySignerHash(signers_hash.clone()))
-        .unwrap_or(0);
+    let signers_epoch: u64 = get_signers_epoch(&env, &signers_hash);
 
     if signers_epoch == 0 {
         panic_with_error!(env, AuthError::InvalidSigners);
@@ -93,7 +89,11 @@ pub fn rotate_signers(env: &Env, new_signers: &WeightedSigners, enforce_rotation
         .persistent()
         .set(&DataKey::SignerHashByEpoch(new_epoch), &new_signers_hash);
 
-    // If new_signers has been rotated to before, we will overwrite the epoch to point to the latest
+    // signers must be distinct, since nonce should guarantee uniqueness even if signers are repeated
+    if get_signers_epoch(&env, &new_signers_hash) != 0 {
+        panic_with_error!(env, AuthError::DuplicateSigners);
+    }
+
     env.storage().persistent().set(
         &DataKey::EpochBySignerHash(new_signers_hash.clone()),
         &new_epoch,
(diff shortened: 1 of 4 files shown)


contracts/axelar-gateway/src/auth.rs
```diff
@@ -202,3 +202,10 @@ fn validate_signers(env: &Env, weighted_signers: &WeightedSigners) {
         panic_with_error!(env, AuthError::InvalidThreshold);
     }
 }
+
+fn get_signers_epoch(env: &Env, signers_hash: &BytesN<32>) -> u64 {
+    env.storage()
+        .persistent()
+        .get(&DataKey::EpochBySignerHash(signers_hash.clone()))
+        .unwrap_or(0)
```
milapsheth, line 210: use of default `0` is a bad practice. `get` returns a rich `Option` type which we should use directly. Due to the way `initialize` works, `0` would never be a valid return for epoch by signers hash anyways, since first rotation always assign epoch 1

contracts/axelar-gateway/src/auth.rs
```diff
@@ -43,11 +43,7 @@ pub fn validate_proof(env: &Env, data_hash: BytesN<32>, proof: Proof) -> bool {
...
     let signers_hash: BytesN<32> = env.crypto().keccak256(&signers_set.to_xdr(env)).into();
 
-    let signers_epoch: u64 = env
-        .storage()
-        .persistent()
-        .get(&DataKey::EpochBySignerHash(signers_hash.clone()))
-        .unwrap_or(0);
+    let signers_epoch: u64 = get_signers_epoch(env, &signers_hash);
 
     if signers_epoch == 0 {
```
milapsheth, line 48: remove the use of `0` and check for `None` directly, you can unwrap via a let

`let Some(epoch) = env...get(DataKey::Epoch) else { panic_with_error }`

contracts/axelar-gateway/src/auth.rs
```diff
@@ -93,7 +89,11 @@ pub fn rotate_signers(env: &Env, new_signers: &WeightedSigners, enforce_rotation
         .persistent()
         .set(&DataKey::SignerHashByEpoch(new_epoch), &new_signers_hash);
 
-    // If new_signers has been rotated to before, we will overwrite the epoch to point to the latest
+    // signers must be distinct, since nonce should guarantee uniqueness even if signers are repeated
```
milapsheth, line 92: also need to update gateway::rotate_signers where duplication is checked by storing a `RotationExecuted` key. The goal of this change is that we don't need to do that separately, similar to the EVM gateway

contracts/axelar-gateway/src/auth.rs
ahramy: got it, updated code.

contracts/axelar-gateway/src/auth.rs
ahramy: I see. updated.

contracts/axelar-gateway/src/auth.rs
ahramy: updated.

contracts/axelar-gateway/src/auth.rs
```diff
@@ -202,3 +199,9 @@ fn validate_signers(env: &Env, weighted_signers: &WeightedSigners) {
         panic_with_error!(env, AuthError::InvalidThreshold);
     }
 }
+
+fn get_signers_epoch(env: &Env, signers_hash: &BytesN<32>) -> Option<u64> {
```
milapsheth, line 203:
```suggestion
fn signers_epoch(env: &Env, signers_hash: &BytesN<32>) -> Option<u64> {
```
`get` is redundant and inconsistent with our naming convention

ahramy: got it. updated.

milapsheth (approved):

PR #37, opened 2024-10-21
test(axelar-gateway): add golden test for weighted signers

ahramy:
[AXE-6094](https://axelarnetwork.atlassian.net/browse/AXE-6094)

[AXE-6094]: https://axelarnetwork.atlassian.net/browse/AXE-6094?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelar-soroban-interfaces/src/types.rs
```diff
@@ -57,10 +63,15 @@ pub struct Message {
 }
 
 impl WeightedSigners {
-    // Get hash of WeightedSigners
     pub fn hash(&self, env: &Env) -> BytesN<32> {
         env.crypto().keccak256(&self.clone().to_xdr(env)).into()
     }
+
+    pub fn signers_rotation_hash(&self, env: &Env) -> BytesN<32> {
```
milapsheth, line 70: good to add this method but we're trying to move away from adding things to the interface crate. i'll approve to unblock the golden test, but we should move away from this crate asap

contracts/axelar-soroban-interfaces/src/types.rs
```diff
@@ -1,5 +1,4 @@
 use soroban_sdk::{contracttype, xdr::ToXdr, Address, BytesN, Env, String, Vec};
-
```
milapsheth, line 2: why remove the newline

contracts/axelar-soroban-interfaces/src/types.rs
```diff
@@ -79,3 +90,79 @@ impl Proof {
...
+                &hex!("8784bf7be5a9baaeea47e12d9e8ad0dec29afcbc3617d97f771e3c24fa945dce"),
+            ),
+        };
+
+        let hash = weighted_signers.hash(&env);
+        let signers_rotation_hash = weighted_signers.signers_rotation_hash(&env);
+
+        let expected_hash = BytesN::<32>::from_array(
+            &env,
+            &hex!("18b1ff823e202dab87cada621717e5be4955734bb973151eb489e6f1576ce3d4"),
```
milapsheth, line 157: use golden files instead of hardcoding vectors in the test itself (using `goldie`)

contracts/axelar-soroban-interfaces/Cargo.toml
```diff
@@ -8,4 +8,5 @@ crate-type = ["rlib"]
 
 [dependencies]
 axelar-soroban-std = { workspace = true }
+hex-literal = "0.3"
```
milapsheth, line 11: this is a dev dependency, since it's only used for tests

contracts/axelar-soroban-interfaces/src/types.rs
ahramy: I think it was removed while add/remove dependencies by mistake. updated.

contracts/axelar-soroban-interfaces/Cargo.toml
ahramy: Good call, moved under dev dependency

contracts/axelar-soroban-interfaces/src/types.rs
ahramy: I see. I added `signers_rotation_hash` function similar to [the imp WeightedSigners](https://github.com/axelarnetwork/axelar-amplifier/blob/7decfc9b0f2277a362c621e2bb886b5005514094/external-gateways/stellar/src/lib.rs#L174) in axelar-amplifier. 
Then I'm wondering if we should move the `hash` function as well. If we only move `signers_rotation_hash`, the test code would look like the example below, which seems inconsistent.

```
        let hash = weighted_signers.hash(&env);
        let signers_rotation_hash = signers_rotation_hash(&env, weighted_signers.signers.clone());
```

contracts/axelar-soroban-interfaces/src/types.rs
ahramy: I should've left a comment here before your review. I already tried it, but the Goldie library requires using `std`, which gives an error. 

Error:
```
failed to resolve: could not find std in the list of imported crates
could not find std in the list of imported cratesrustcClick for full compiler diagnostic
lib.rs(105, 15): Actual error occurred here
```

Goldie Library:
https://github.com/rossmacarthur/goldie/blob/81daa5650aba992782b5d752c8839890bdd246d9/src/lib.rs#L104

```
/// Assert the JSON golden file matches.
#[macro_export]
macro_rules! assert_json {
    ($actual:expr) => {{
        let g = $crate::_new_goldie!();
        if let Err(err) = g.assert_json($actual) {
            ::std::panic!("{}", err);
        }
    }};
}
```

I asked in the Discord channel if there are any other libraries or substitutes for a golden test.

contracts/axelar-soroban-interfaces/src/types.rs
milapsheth: we'll be removing the interface crate altogether and move all the types to the appropriate crates. the usage is here fine until then

contracts/axelar-soroban-interfaces/src/types.rs
milapsheth: Yes, because this is a `no_std` crate. You can read about it in Rust docs. But here we're importing it in tests, so it should be added to dev dependencies and not dependencies (see axelar-amplifier repo for reference)

contracts/axelar-soroban-interfaces/src/types.rs
ahramy: Got it. thanks for clarifying it

contracts/axelar-gateway/src/auth.rs
```diff
@@ -1,5 +1,5 @@
 use axelar_soroban_interfaces::types::{ProofSignature, ProofSigner, WeightedSigner};
```
milapsheth, line 1: The PR title should use `test(axelar-gateway): ..` as the scope. See the conventional commit page
https://www.conventionalcommits.org/en/v1.0.0/#summary

contracts/axelar-soroban-interfaces/src/types.rs
```diff
@@ -163,7 +164,9 @@ mod tests {
             &hex!("4ad8f3015146ac68334fd405f90e6ca75fbf2c276b333a8747c9ba83d9c3f1f6"),
         );
 
-        assert_eq!(hash, expected_hash);
-        assert_eq!(signers_rotation_hash, expected_signers_rotation_hash);
+        //assert_eq!(hash, expected_hash);
+        //assert_eq!(signers_rotation_hash, expected_signers_rotation_hash);
+
+        goldie::assert_json!(&vec![hash, signers_rotation_hash]);
```
milapsheth, line 170: use `std::vec!`

contracts/axelar-soroban-interfaces/src/types.rs
```diff
@@ -95,9 +95,10 @@ impl Proof {
 #[cfg(test)]
 mod tests {
     use hex_literal::hex;
-    use soroban_sdk::{BytesN, Env, Vec};
+    use soroban_sdk::{vec, BytesN, Env, Vec};
 
     use crate::types::{WeightedSigner, WeightedSigners};
+    extern crate std;
```
milapsheth, line 101: remove this

contracts/axelar-gateway/src/auth.rs
ahramy: got it. updated.

milapsheth (approved):

PR #39, opened 2024-10-22
feat(ITS): allow owner to add/remove trusted addresses

AttissNgo:
[AXE-6066](https://axelarnetwork.atlassian.net/browse/AXE-6066#)

[AXE-6066]: https://axelarnetwork.atlassian.net/browse/AXE-6066?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelar-interchain-token-service/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "axelar-interchain-token-service"
```
milapsheth, line 2:
```suggestion
name = "interchain-token-service"
```
we don't really prefix ITS with axelar for other chains

contracts/axelar-interchain-token-service/Cargo.toml
```diff
@@ -0,0 +1,17 @@
...
+
+[dependencies]
+axelar-soroban-interfaces = { workspace = true }
+axelar-soroban-std = { workspace = true }
+soroban-sdk = { workspace = true }
+
+[dev_dependencies]
+soroban-sdk = { workspace = true, features = ["testutils"] }
+axelar-soroban-std = { workspace = true, features = ["testutils"] }
+
```
milapsheth, line 17: extra line

contracts/axelar-interchain-token-service/src/error.rs
```diff
@@ -0,0 +1,12 @@
...
+#[contracterror]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum InterchainTokenServiceError {
+    NotInitialized = 1,
+    AlreadyInitialized = 2,
+    NotOwner = 3,
+    TrustedAddressAlreadyAdded = 4,
+    NotTrustedAddress = 5,
+}
```
milapsheth, line 12: missing trailing newline char. update your editor config to add this by default

contracts/axelar-interchain-token-service/src/error.rs
```diff
@@ -0,0 +1,12 @@
+use soroban_sdk::contracterror;
+
+#[contracterror]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum InterchainTokenServiceError {
```
milapsheth, line 6:
```suggestion
pub enum ContractError {
```
since we aren't using the interface crate anymore, we can give this a common name

contracts/axelar-interchain-token-service/src/error.rs
```diff
@@ -0,0 +1,12 @@
+use soroban_sdk::contracterror;
+
+#[contracterror]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
```
milapsheth, line 4: i forget whether soroban errors require all of these. `Ord` is an interesting one to require for error types

contracts/axelar-interchain-token-service/src/event.rs
```diff
@@ -0,0 +1,16 @@
+use soroban_sdk::{symbol_short, String, Address, Env};
+
+pub(crate) fn set_trusted_address(env: &Env, chain: String, trusted_address: String) {
+    let topics = (symbol_short!("set"), chain, trusted_address);
```
milapsheth, line 4:
```suggestion
    let topics = (symbol_short!("trusted_address_set"), chain, trusted_address);
```
can you take a look at efficiency difference between short symbols vs normal symbols? The deployment scripts have a `--estimateCost` flag that can give you resource consumption details to compare. It's probably minor, in which case we should just switch to more readable names

contracts/axelar-interchain-token-service/src/lib.rs
```diff
@@ -0,0 +1,18 @@
+#![no_std]
+
+mod error;
+mod event;
+mod storage_types;
+
+pub mod contract;
+
+// #[cfg(all(target_family = "wasm", feature = "testutils"))]
```
milapsheth, line 9: remove this

contracts/axelar-interchain-token-service/src/storage_types.rs
```diff
@@ -0,0 +1,17 @@
...
+    pub trusted_address: String,
+}
+
+#[contracttype]
+#[derive(Clone, Debug)]
+pub enum DataKey {
+    Initialized,
+    AuthModule,
+    Owner,
+    TrustedAddress(TrustedItsAddress),
```
milapsheth, line 16: the key is the chain name, the value is the address. See `SignerHashByEpoch` in the gateway for reference

milapsheth, line 16:
```suggestion
    TrustedAddress {
        chain: String
    },
```

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+use crate::error::InterchainTokenServiceError;
+use crate::event;
+use crate::storage_types::{DataKey, TrustedItsAddress};
+
+#[contract]
+pub struct InterchainTokenService;
+
+#[contractimpl]
+impl InterchainTokenService {
+    pub fn initialize(
```
milapsheth, line 13:
```suggestion
    pub fn initialize_its(
```
naming collisions aren't allowed when one contract imports another in soroban

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+        env: Env,
+        owner: Address
+    ) -> Result<(), InterchainTokenServiceError> {
+        ensure!(
+            env.storage()
+                .instance()
+                .get(&DataKey::Initialized)
+                .unwrap_or(true),
```
milapsheth, line 21:
```suggestion
                .is_none(),
```
we should ensure it's false/set to none instead, or fail otherwise

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        env.storage().instance().set(&DataKey::Initialized, &true);
+
+        env.storage().instance().set(&DataKey::Owner, &owner);
+
+        Ok(())
+    }
+
+    pub fn owner(env: &Env) -> Address {
+        env.storage().instance().get(&DataKey::Owner).unwrap()
```
milapsheth, line 33: I'm fine with an unwrap here instead of a result type since it can only happen due to incomplete initialization.

@hydrobeam Does stellar team have a recommendation? Also, Stellar doesn't distinguish txs from queries, so do we need to expose this as an endpoint in our contracts? Can we query this info from storage of the contract directly in an easy to use manner or do we need to expose an entrypoint? Maybe you can find more info here?

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+    pub fn transfer_ownership(env: Env, new_owner: Address) {
+        let owner = Self::owner(&env);
+        owner.require_auth();
+
+        env.storage().instance().set(&DataKey::Owner, &new_owner);
+
+        event::transfer_ownership(&env, owner, new_owner);
+    }
+
+    pub fn is_trusted_address(env: &Env, chain: String, address: String) -> bool {
```
milapsheth, line 45:
```suggestion
    pub fn trusted_address(env: &Env, chain: String) -> String {
```
Retrieve the trusted address instead

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+
+        ensure!(
+            !env.storage().persistent().has(&key),
+            InterchainTokenServiceError::TrustedAddressAlreadyAdded
+        );
+
+        env.storage().persistent().set(&key, &true);
+
+        event::set_trusted_address(&env, chain, address);
+        Ok(())
```
milapsheth, line 72:
```suggestion

        Ok(())
```

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -0,0 +1,100 @@
...
+            InterchainTokenServiceError::NotTrustedAddress
+        );
+
+        env.storage().persistent().remove(&key);
+
+        event::remove_trusted_address(&env, chain, address);
+        Ok(())
+    }
+}
+
```
milapsheth, line 100:
```suggestion
```

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -0,0 +1,149 @@
...
+    initialize_its(&env, &client, owner.clone());
+
+    assert_eq!(
+        client.owner(), owner
+    );
+}
+
+#[test]
+#[should_panic(expected = "HostError: Error(Auth, InvalidAction)")] // Unauthorized
+fn add_trusted_address_fails_if_not_owner() {
```
milapsheth, line 41:
```suggestion
fn set_trusted_address_fails_if_not_owner() {
```

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -0,0 +1,149 @@
...
+            fn_name: "set_trusted_address",
+            args: (chain.clone(), trusted_address.clone()).into_val(&env),
+            sub_invokes: &[],
+        },
+    }]).set_trusted_address(&chain, &trusted_address);
+
+    assert_eq!(
+        client.is_trusted_address(&chain, &trusted_address),
+        true
+    );
```
milapsheth, line 86:
```suggestion
    assert!(
        client.is_trusted_address(&chain, &trusted_address),
    );
```

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -0,0 +1,149 @@
...
+    
+    client.mock_auths(&[MockAuth {
+        address: &owner,
+        invoke: &MockAuthInvoke {
+            contract: &contract_id,
+            fn_name: "set_trusted_address",
+            args: (chain.clone(), trusted_address.clone()).into_val(&env),
+            sub_invokes: &[],
+        },
+    }]).set_trusted_address(&chain, &trusted_address);
```
milapsheth, line 81: add test coverage for failing if address is already set

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -0,0 +1,149 @@
...
+#[test]
+fn remove_trusted_address() {
+    let (env, contract_id, client) = setup_env();
+    let owner = Address::generate(&env);
+
+    initialize_its(&env, &client, owner.clone());
+
+    let chain = String::from_str(&env, "chain");
+    let trusted_address = String::from_str(&env, "trusted_address");
+
```
milapsheth, line 98: add test coverage for failing to removing an address if not set

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -0,0 +1,149 @@
...
+            args: (&new_owner,).into_val(&env),
+            sub_invokes: &[],
+        },
+    }]).transfer_ownership(&new_owner);
+
+    assert_eq!(
+        client.owner(),
+        new_owner
+    );
+}
```
milapsheth, line 148: assert emitted events in tests

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -0,0 +1,149 @@
...
+            contract: &contract_id,
+            fn_name: "transfer_ownership",
+            args: (&new_owner,).into_val(&env),
+            sub_invokes: &[],
+        },
+    }]).transfer_ownership(&new_owner);
+
+    assert_eq!(
+        client.owner(),
+        new_owner
```
milapsheth, line 146: you can check the codecov links to ensure coverage for the PR is always 100%

contracts/axelar-interchain-token-service/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
```
milapsheth, line 1: PR titles should follow this style: https://www.conventionalcommits.org/en/v1.0.0/#summary

see previous merged PRs

contracts/axelar-interchain-token-service/src/error.rs
AttissNgo: Hmm let me check on that...

AttissNgo: According to Soroban's example contract explanation: 
"Contract errors are Rust u32 enums where every variant of the enum is assigned an integer. The #[contracterror] attribute is used to set the error up so it can be used in the return value of contract functions.

The enum has some constraints:

    It must have the #[repr(u32)] attribute.
    It must have the #[derive(Copy)] attribute.
    Every variant must have an explicit integer value assigned.
"

Not sure why so many traits are included in their example. Let me see if there's better docs somewhere else.

contracts/axelar-interchain-token-service/src/contract.rs
AttissNgo: I'm confused - in an earlier comment you instructed me to add a newline at the end of a file. Please help me understand what the best practice is

contracts/axelar-interchain-token-service/src/contract.rs
AttissNgo: My assumption here is is that explicit interaction through entrypoints is desirable. I haven't found any info contrary to this in the docs. I can reach out to Stellar team

contracts/axelar-interchain-token-service/src/test.rs
AttissNgo: To clarify - the expected behavior here is that the function will return an error if the owner attempts to set a trusted address which is already set for that chain, correct? The owner should still be able to CHANGE the trusted address for a given chain, right?

contracts/axelar-interchain-token-service/src/test.rs
AttissNgo: Since we have changed the `is_trusted_address` getter to `trusted_address`, returning address rather than bool, this test has been changed to assert equivalence (address in storage == address passed as arg).

contracts/axelar-interchain-token-service/src/test.rs
AttissNgo: Codecov complaining about attributes required by Soroban

contracts/axelar-interchain-token-service/Cargo.toml
```diff
@@ -13,5 +13,4 @@ soroban-sdk = { workspace = true }
 
 [dev_dependencies]
 soroban-sdk = { workspace = true, features = ["testutils"] }
-axelar-soroban-std = { workspace = true, features = ["testutils"] }
-
+axelar-soroban-std = { workspace = true, features = ["testutils"] }
```
milapsheth, line 16: still seeing trailing newline warning, take a look at editor setting to fix this

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -42,59 +39,52 @@ impl InterchainTokenService {
...
         Ok(())
-    } 
+    }
 
     pub fn remove_trusted_address(
-        env: Env, 
-        chain: String, 
+        env: Env,
+        chain: String,
         address: String,
```
milapsheth, line 72: When removing a trusted address, only the chain (i.e key) needs to be passed in. See the EVM ITS methods for reference.
https://github.com/axelarnetwork/interchain-token-service/blob/main/contracts/interfaces/IAddressTracker.sol#L12

contracts/axelar-interchain-token-service/src/storage_types.rs
```diff
@@ -1,17 +1,9 @@
...
 #[contracttype]
 #[derive(Clone, Debug)]
 pub enum DataKey {
     Initialized,
-    AuthModule,
     Owner,
-    TrustedAddress(TrustedItsAddress),
-}
\ No newline at end of file
+    TrustedAddress(String),
```
milapsheth, line 8:
```suggestion
    TrustedAddress {
        chain: String
    },
```
can we use named args instead? It might be a bit less efficient if it's getting XDR encoded though, something to investigate as part of determining the storage pattern

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -16,12 +18,8 @@ fn setup_env<'a>() -> (Env, Address, InterchainTokenServiceClient<'a>) {
     (env, contract_id, client)
 }
 
-fn initialize_its(
-    _env: &Env, 
-    client: &InterchainTokenServiceClient,
-    owner: Address,
-) {
-    client.initialize(&owner);
+fn initialize_its(_env: &Env, client: &InterchainTokenServiceClient, owner: Address) {
```
milapsheth, line 21:
```suggestion
fn initialize(_env: &Env, client: &InterchainTokenServiceClient, owner: Address) {
```
test method can be called `initialize` since there's no name collision concern

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -31,59 +29,145 @@ fn initialize() {
...
+    let owner = Address::generate(&env);
+
+    initialize_its(&env, &client, owner.clone());
+
+    let chain = String::from_str(&env, "chain");
+    let trusted_address = String::from_str(&env, "trusted_address");
+    let another_trusted_address = String::from_str(&env, "another_trusted_address");
+
+    client
+        .mock_auths(&[MockAuth {
```
milapsheth, line 121: This and the following test isn't checking for auth, so better to just mock all auths `env.mock_all_auths();`

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -31,59 +29,145 @@ fn initialize() {
...
+                    fn_name: "set_trusted_address",
+                    args: (chain.clone(), trusted_address.clone()).into_val(&env),
+                    sub_invokes: &[],
+                },
+            }])
+            .try_set_trusted_address(&chain, &trusted_address),
+        ContractError::TrustedAddressAlreadySet
     );
+
+    // to change trusted address, owner must remove and set again
```
milapsheth, line 147: once auth mocking is removed, behaviour should be evident so comment isn't needed

contracts/axelar-interchain-token-service/src/test.rs
milapsheth: yeah, those are expected, our tests might not make use of something like Debug trait of that type etc. which doesn't require coverage. We care more about the coverage of the logic

contracts/axelar-interchain-token-service/src/storage_types.rs
milapsheth: yep `TrustedAddress(String)`

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -66,24 +66,21 @@ impl InterchainTokenService {
...
-            || Self::trusted_address(&env, chain.clone()).unwrap() != address
-        {
-            return Err(ContractError::NotTrustedAddress);
-        }
+        ensure!(
+            env.storage().persistent().has(&key),
+            ContractError::NoTrustedAddressSet
+        );
+
+        let trusted_address = Self::trusted_address(&env, chain.clone()).unwrap();
```
milapsheth, line 79:
```suggestion
        ensure!(
            env.storage().persistent().has(&key),
            ContractError::NoTrustedAddressSet
        );

        let Some(trusted_address) = Self::trusted_address(&env, chain.clone()) else {
            return ContractError::NoTrustedAddressSet;
        }
```
avoid an extra state query

contracts/axelar-interchain-token-service/src/contract.rs
```diff
@@ -66,24 +66,21 @@ impl InterchainTokenService {
...
 
         env.storage().persistent().remove(&key);
```
milapsheth, line 81:
```suggestion
        env.storage().persistent().remove(&DataKey::TrustedAddress(chain.clone()));
```
can use it directly

contracts/axelar-interchain-token-service/src/error.rs
```diff
@@ -8,5 +8,5 @@ pub enum ContractError {
     AlreadyInitialized = 2,
```
milapsheth, line 8: remove explicit numbering from the error

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -18,16 +18,16 @@ fn setup_env<'a>() -> (Env, Address, InterchainTokenServiceClient<'a>) {
...
 }
 
-fn initialize_its(_env: &Env, client: &InterchainTokenServiceClient, owner: Address) {
+fn initialize(_env: &Env, client: &InterchainTokenServiceClient, owner: Address) {
     client.initialize_its(&owner);
 }
 
 #[test]
-fn initialize() {
+fn initialize_sets_owner() {
```
milapsheth, line 26:
```suggestion
fn initialize_succeeds() {
```
perhaps?

contracts/axelar-interchain-token-service/src/test.rs
```diff
@@ -108,101 +100,44 @@ fn set_trusted_address_fails_if_not_owner() {
 
 #[test]
 fn set_trusted_address_fails_if_already_set() {
-    let (env, contract_id, client) = setup_env();
+    let (env, _contract_id, client) = setup_env();
```
milapsheth, line 103:
```suggestion
    let (env, _, client) = setup_env();
```
perhaps

contracts/axelar-interchain-token-service/src/test.rs
milapsheth, line 1: Move tests to the integration `tests` folder

contracts/axelar-interchain-token-service/src/lib.rs
```diff
@@ -0,0 +1,12 @@
+#![no_std]
+
+mod error;
```
milapsheth, line 3:
```suggestion
pub mod error;
```
allows use in integration tests and for consumers

contracts/axelar-interchain-token-service/Cargo.toml
milapsheth, line 1: Update crate name to remove `axelar` prefix

contracts/axelar-interchain-token-service/Cargo.toml
```diff
@@ -0,0 +1,16 @@
+[package]
+name = "interchain-token-service"
```
milapsheth, line 2: Update root `Cargo.toml` dependencies to include this crate

contracts/axelar-interchain-token-service/src/error.rs
AttissNgo: It will not compile without explicit numbering: `enums are supported as contract errors only when all variants have an explicit integer literal`

contracts/axelar-interchain-token-service/src/test.rs
AttissNgo: Do you not find the `_ignore` pattern potentially useful, say for example if later on the test changes and we need to use the `contract_id`?

contracts/axelar-interchain-token-service/src/test.rs
AttissNgo: Please advise on how to rename `test.rs` when moving to `integration-tests > tests` directory. Should it be `its.rs`? `interchain-token-service.rs`? Something else?

AttissNgo: There is an issue here with the InterchainTokenService contract failing to satisfy the trait bound `ContractFunctionSet` when trying to call the `env.register_contract()` method. I can't find this trait in the Soroban SDK docs. When creating separate `tests` folder for the gateway, this is not an issue, only with ITS contract. Does this imply that we need to implement an interface or is it something else?

milapsheth (approved):

PR #40, opened 2024-10-22
feat(gateway): add ownership for upgrade

ahramy:
[AXE-6076](https://axelarnetwork.atlassian.net/browse/AXE-6076)

[AXE-6076]: https://axelarnetwork.atlassian.net/browse/AXE-6076?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/axelar-gateway/src/test_gateway.rs
+++ b/contracts/axelar-gateway/src/test_gateway.rs
@@ -32,6 +32,7 @@ fn setup_env<'a>() -> (Env, Address, AxelarGatewayClient<'a>) {
 #[test]
 fn fails_if_already_initialized() {
     let (env, _contract_id, client) = setup_env();
+    let owner = Address::generate(&env);
     let operator = Address::generate(&env);
 
     // doing the process from testutils::initialize() manually so we can
@@ -44,6 +45,7 @@ fn fails_if_already_initialized() {
     let minimum_rotation_delay = 0;
 
     client.initialize(
+        &owner,
         &operator,
         &signer_set.domain_separator,
         &minimum_rotation_delay,
@@ -53,6 +55,7 @@ fn fails_if_already_initialized() {
 
     assert_contract_err!(
         client.try_initialize(
+            &owner,
             &operator,
             &signer_set.domain_separator,
             &minimum_rotation_delay,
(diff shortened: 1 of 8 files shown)


contracts/axelar-gateway/src/test_gateway.rs
```diff
@@ -476,3 +489,65 @@ fn transfer_operatorship_unauthorized() {
...
+            new_owner.clone(),
+        ),
+        (),
+    );
+
+    assert_eq!(client.owner(), new_owner);
+}
+
+#[test]
+#[should_panic(expected = "HostError: Error(Auth, InvalidAction)")] // Unauthorized
```
milapsheth, line 531: use `try_*` version of the function and assert err using the new pattern from Laith

AttissNgo (approved):

contracts/axelar-gateway/src/test_gateway.rs
ahramy: I attempted to add `GatewayError::Unauthorized` and integrate `ensure!` logic for `try_` error handling. However, since `require_auth()` panics, it can't be handled by the `try_*` functions. I confirmed this with Laith. @milapsheth Let me know your thoughts.

milapsheth: I see, define the error string as a constant so we can reuse it for all auth failure tests. I'm a bit surprised that it panics vs returns a certain type of error. Ask stellar team about it so we know how to do this in the future?

milapsheth (approved):

contracts/axelar-gateway/src/test_gateway.rs
ahramy: I tried to define `const ERROR_STRING: &str = "HostError: Error(Auth, InvalidAction)";` and tried to use it for all auth failure test, it won't let me use a variable to the `expected=` section. I created [a ticket](https://axelarnetwork.atlassian.net/browse/AXE-6136) to track this issue, will ask to the stellar team.

PR #42, opened 2024-10-23
refactor(gateway): move gateway types out of `axelar-soroban-interfaces`

hydrobeam:
closes [AXE-6155](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-6155)

## Summary

- moved contents of `interfaces/types.rs` to `gateway/types.rs`
  - maybe merge with `storage_types.rs`? held off since `storage_types` is a private module, meanwhile `types` needs to be `pub` since it's used in the integration tests.
-  dropped `AxelarGatewayInterface` in f88a5d9b881c1ca48b0944ecc32775cd91f14c8f.  required setting themethods in the `AxelarGateway` to `pub`
    - brought over the comments attached to the methods
- moved over the tests in the interfaces crate as unit tests defined in `types.rs`.
   - because the tests depend on `goldie` which relies on std, I had to add a crate level `cfg[test] + extern crate std;`. otherwise I couldn't seem to get `goldie` to compile. adding `extern crate std` in the test module itself did not work
- brought over the `AxelarExecutableInterface` to `exectuable.rs`
   - it depends on the AxelarGatewayClient so it made sense to handle in the same pr.
- fix many imports

PR #43, opened 2024-10-24
refactor(gas-service,operators): move out of `axelar-soroban-interfaces`

hydrobeam:
closes [AXE-6193](https://axelarnetwork.atlassian.net/browse/AXE-6193)

Similar to #42 except these contracts are standalone, so moving them out of the interfaces crate is pretty straightforward.

## Summary

- renamed respective error types to `ContractError`
- Dropped "Interface" traits in favour of impls on structs. 
- Copied over comments that were attached to interface traits to struct impls

[AXE-6193]: https://axelarnetwork.atlassian.net/browse/AXE-6193?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelar-gas-service/src/error.rs
```diff
@@ -0,0 +1,12 @@
+use soroban_sdk::contracterror;
+
+#[contracterror]
+#[derive(Copy, Clone, Debug, Eq, PartialEq, PartialOrd, Ord)]
+#[repr(u32)]
+pub enum ContractError {
+    InvalidAddress = 1,
```
ahramy, line 7: Just wondering, how are we going to handle the order of error names? Should they be in alphabetical order or based on occurence? In other contracts, NotInitialized / AlreadyInitialized are assigned higher numbers. (e.g. [GatewayError::NotInitialized](https://github.com/axelarnetwork/axelar-cgp-soroban/blob/3e7d28a8806fec2c689989b2e50de1860587190c/contracts/axelar-soroban-interfaces/src/axelar_gateway.rs#L12) comes first). I think it would be great to have some kind of convention or simple rules for this.

milapsheth: you can see if Rust recommends something. We should avoid specifying error enum values though, and let it use the defaults

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -88,17 +95,20 @@ impl AxelarGasServiceInterface for AxelarGasService {
         Ok(())
     }
 
-    fn refund(
+    /// Refunds gas payment to the receiver in relation to a specific cross-chain transaction.
+    ///
+    /// Only callable by the gas_collector.
```
milapsheth, line 100:
```suggestion
    /// Only callable by the `gas_collector`.
```

contracts/axelar-gas-service/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![no_std]
 
 pub mod contract;
+mod error;
```
milapsheth, line 4:
```suggestion
pub mod error;
```

contracts/axelar-operators/src/lib.rs
```diff
@@ -1,6 +1,7 @@
 #![no_std]
 
 pub mod contract;
+mod error;
```
milapsheth, line 4:
```suggestion
pub mod error;
```

milapsheth (approved):

milapsheth (approved):

contracts/axelar-gas-service/src/error.rs
hydrobeam: i don't think there's a particular recommendation. it'd be nice if they were all aligned but not really that important since (as far as i know) you rarely interact with the values behind integer variants explicitly.

PR #45, opened 2024-10-28
chore(gateway): coverage for upgrade and version

ahramy:
[AXE-6267](https://axelarnetwork.atlassian.net/browse/AXE-6267)

[AXE-6267]: https://axelarnetwork.atlassian.net/browse/AXE-6267?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelar-gateway/src/test_gateway.rs
```diff
@@ -555,3 +555,56 @@ fn transfer_ownership_unauthorized() {
...
+
+#[test]
+#[should_panic(expected = "HostError: Error(Storage, MissingValue)")]
+fn upgrade() {
+    let (env, _, client) = setup_env();
+    let owner = Address::generate(&env);
+    let operator = Address::generate(&env);
+    let new_wasm_hash = BytesN::<32>::from_array(&env, &[0; 32]);
+
+    initialize(&env, &client, owner.clone(), operator, 1, randint(1, 10));
```
milapsheth, line 581:
```suggestion
    initialize(&env, &client, owner, operator, 1, randint(1, 10));
```

contracts/axelar-gateway/src/test_gateway.rs
```diff
@@ -555,3 +555,56 @@ fn transfer_ownership_unauthorized() {
...
+
+    // Should panic with invalid wasm hash
+    client.upgrade(&new_wasm_hash);
```
milapsheth, line 584: Can we use try_upgrade and assert err? Or does it implicitly panic?

contracts/axelar-gateway/src/test_gateway.rs
```diff
@@ -555,3 +555,56 @@ fn transfer_ownership_unauthorized() {
...
+}
+
+#[test]
+#[should_panic(expected = "HostError: Error(Auth, InvalidAction)")] // Unauthorized
```
milapsheth, line 588:
```suggestion
#[should_panic(expected = "HostError: Error(Auth, InvalidAction)")]
```

contracts/axelar-gateway/src/test_gateway.rs
```diff
@@ -555,3 +555,56 @@ fn transfer_ownership_unauthorized() {
...
+
+    assert_eq!(
+        client.version(),
+        String::from_str(&env, env!("CARGO_PKG_VERSION"))
+    );
+}
+
+#[test]
+#[should_panic(expected = "HostError: Error(Storage, MissingValue)")]
+fn upgrade() {
```
milapsheth, line 575:
```suggestion
fn upgrade_invalid_wasm_hash() {
```

contracts/axelar-gateway/src/test_gateway.rs
```diff
@@ -555,3 +555,56 @@ fn transfer_ownership_unauthorized() {
...
+        .mock_auths(&[MockAuth {
+            address: &user,
+            invoke: &MockAuthInvoke {
+                contract: &contract_id,
+                fn_name: "upgrade",
+                args: (new_wasm_hash.clone(),).into_val(&env),
+                sub_invokes: &[],
+            },
+        }])
+        .upgrade(&new_wasm_hash);
```
milapsheth, line 609: Is there a positive test case for upgrade?

contracts/axelar-gateway/src/test_gateway.rs
ahramy: It does implicitly panic similar to require_auth.

contracts/axelar-gateway/src/test_gateway.rs
ahramy: I was trying to add a positive test, but it requires a new WASM hash which is obtained from `env.deployer().upload_contract_wasm(new_contract::WASM);` with `new_contract` imported from a WASM file. I'm trying to avoid using a file path here, and I also dug into the SDK, but no luck. I left a question on Discord and will share the answer once I get it.

ahramy: I got an answer from them: 
> Getting the contract Wasm Hash without specifying the file path isn't possible to do this at the moment. A native contract defined in the test/crate when registered doesn't have a wasm hash and can't be referenced by a wasm hash. In general the test enviroment doesn't support installing the contract, it's always registered as a contract instance.

Thus, I need to add an integration test with an actual Wasm file for a positive test. WIP.

integration-tests/tests/upgrade.rs
```diff
@@ -0,0 +1,50 @@
+#![cfg(test)]
+
+use axelar_gateway::testutils::initialize;
+
+use axelar_gateway::AxelarGatewayClient;
+use soroban_sdk::{testutils::Address as _, Address, Env, String};
+
+mod old_contract {
+    soroban_sdk::contractimport!(
+        file = "../integration-tests/tests/testdata/axelar_gateway_old.wasm"
```
milapsheth, line 10: instead of checking it in, retrieve the build dynamically from the `target`? you can use cargo macros to get the release type etc. to get the path perhaps. Maybe ask chatgpt/claude for help

ahramy: I tried asking ChatGPT and Claude for help in multiple ways, but none of my attempts worked. I believe the issue is that Stellar SDK procedural macros, like `contractimport!`, require a static string for the file path and don’t support dynamically generated paths via other macros. I'm inclined to keep it as it is for now, but I agree that we should move this under the gateway test.

milapsheth: Ok, add comments on how you generated these for reproducibility

milapsheth (approved):

integration-tests/tests/upgrade.rs
ahramy: updated.

PR #47, opened 2024-10-29
feat: raise clippy lint level to `clippy::nursery` and apply lints

hydrobeam:
closes [AXE-6293](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-6293)

## Summary

Apply lints from the `clippy::nursery` lint level and update the `lint.yaml` workflow to include it. The lints that required a change were:

- [`redundant_clones`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone)
- [`redundant_pub(crate)`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_pub_crate)
- [`derive_partial_eq_without_eq`](https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq)

## Motivation

Nursery lints are "new lints that are still under development" (https://doc.rust-lang.org/clippy/). IMO they're all reasonable and can catch inaccuracies (list of nursery lints: https://rust-lang.github.io/rust-clippy/master/index.html#/nursery )

While looking into how to reduce some of the clones in some contracts, I encountered some places where the [`reundant_clone`](https://rust-lang.github.io/rust-clippy/master/index.html#redundant_clone) lint was catching some redundant clones. It won't catch everything and (by its own admission) has false-negatives, but it should improve the repo's code quality overall. Same goes for the other lints it adds.

If we encounter a situation where they're too restrictive we can relax the lint level, or just apply a local directive (-> https://doc.rust-lang.org/clippy/configuration.html#allowingdenying-lints)

## Misc

Looked into the `clippy::pedantic` lint level, but there were a lot of false positives and opinionated lints, so I opted against it. Although there were a few valid points in some of them

[AXE-6293]: https://axelarnetwork.atlassian.net/browse/AXE-6293?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


ahramy (approved):
LGTM!

.github/workflows/lint.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
         uses: actions-rs/cargo@v1
         with:
           command: clippy
-          args: -- -D warnings -A deprecated
+          args: -- -D warnings -A deprecated -W clippy::nursery
```
milapsheth, line 54: add this in clippy config in Cargo.toml or a config file so it applies locally as well (we don't have to pass it in everytime)

milapsheth (approved):

.github/workflows/lint.yaml
hydrobeam: done

milapsheth (approved):

PR #49, opened 2024-10-30
refactor: move contract tests to integration tests

hydrobeam:
closes AXE-6253

## Summary

Most of the contract tests we have test the public facing API, so it's better for them to be integration tests

- moved each contract's `test.rs` to a a `tests/test.rs` located within the contract's crate
  - gateway's `test_gateway.rs` relies only on the public API, so it was moved there, renamed to `tests/test.rs`
  - the `test_auth.rs` is entirely focused on the private `auth` API, so it was moved to the `auth` module under a test module
- added `rlib` to each contract's `crate-type`.
  - this is needed for integration tests to compile, relevant forum post: https://users.rust-lang.org/t/does-crate-type-has-effect-on-integration-test/76634 
- a few tests were moved to their respective `contract.rs` under a test module
  - they needed access to the `DataKey` type to query internal storage


hydrobeam:
On the latest `soroban_sdk` version the tests run fine, so I guess this kind of depends on #50

ahramy (approved):

contracts/axelar-operators/src/contract.rs
```diff
@@ -131,3 +131,30 @@ impl AxelarOperators {
...
+        let client = AxelarOperatorsClient::new(&env, &contract_id);
+        let user = Address::generate(&env);
+
+        client.initialize(&user);
+
+        assert_some!(env.as_contract(&contract_id, || {
+            env.storage()
+                .instance()
+                .get::<DataKey, bool>(&DataKey::Initialized)
+        }));
```
milapsheth, line 156:
```suggestion
        assert_env!(contract_id, assert_some!(
            env.storage()
                .instance()
                .get::<DataKey, bool>(&DataKey::Initialized)
        )));
```
wonder if we should some macro like this for convenience. `env` can be assumed to exist in the scope already perhaps. Otherwise, move the `assert_some` to inside the function for consistency

milapsheth: can be done in a separate PR though if we want to add this

contracts/axelar-gateway/src/auth.rs
```diff
@@ -235,3 +235,656 @@ fn validate_signers(env: &Env, weighted_signers: &WeightedSigners) -> Result<(),
...
+        env.as_contract(&contract_id, || {
+            assert_err!(
+                auth::validate_proof(&env, &msg_hash, proof),
+                ContractError::InvalidSignersHash
+            );
+        });
+    }
+
+    #[test]
+    fn test_rotate_signers() {
```
milapsheth, line 540: avoid `test_` prefix
```suggestion
    fn rotate_signers_succeeds() {
```

contracts/axelar-gateway/src/auth.rs
```diff
@@ -235,3 +235,656 @@ fn validate_signers(env: &Env, weighted_signers: &WeightedSigners) -> Result<(),
 
     Ok(())
 }
+
+#[cfg(test)]
+mod tests {
```
milapsheth, line 240: Most of these need to be converted into integration tests. Are you planning to do that in a separate PR?

contracts/axelar-gateway/src/auth.rs
hydrobeam: done in [70f1b26](https://github.com/axelarnetwork/axelar-cgp-soroban/pull/49/commits/70f1b26bdb431ad1e877243c5aa318f0a9d5cf5f)

contracts/axelar-operators/src/contract.rs
hydrobeam: having it look exactly like that would require a proc macro since regular macros are hygenic and can't pull in variables from outside their context. will save it for a future PR since most instances of the `env.as_contract()` pattern occur in `auth.rs` which we're planning to rework (and probably removing most of its uses).

handled the second part in [f04854c](https://github.com/axelarnetwork/axelar-cgp-soroban/pull/49/commits/f04854c5df691ac200ccc91c72ca7b32900e28ea)

PR #52, opened 2024-10-31
test: add expected invoke auth error for unit tests

ahramy:
[AXE-6136](https://axelarnetwork.atlassian.net/browse/AXE-6136)

[AXE-6136]: https://axelarnetwork.atlassian.net/browse/AXE-6136?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelar-gateway/src/testutils.rs
```diff
@@ -192,3 +193,13 @@ pub fn rotate_signers(env: &Env, contract_id: &Address, new_signers: TestSignerS
...
+
+pub fn expect_auth_error<T>(call_result: Result<T, ContractError>) {
+    if let Some(_) = call_result.err() {
+        let val = ScVal::Error(ScError::Context(ScErrorCode::InvalidAction));
+        match ScError::try_from(val) {
+            Ok(ScError::Context(ScErrorCode::InvalidAction)) => {}
+            _ => panic!("Expected ScErrorCode::InvalidAction"),
+        }
+    }
+}
```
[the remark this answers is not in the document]
ahramy: yeah, this applies for all `ScErrorCode::InvalidAction` type of errors. I agree that It's better to change the name to `expect_invalid_action_error`. 

For the upgrade, `update_current_contract_wasm` returns `Infallible`, so there doesn't seem to be a way to catch an error. I believe we have covered a positive test for it in the integration tests, so it should be fine.

contracts/axelar-gateway/src/testutils.rs
```diff
@@ -192,3 +193,13 @@ pub fn rotate_signers(env: &Env, contract_id: &Address, new_signers: TestSignerS
         (),
     );
 }
+
+pub fn expect_invalid_action_error<T>(call_result: Result<T, ContractError>) {
```
milapsheth, line 197: This should exist in the `std` crate, and be made into a macro that avoids the duplication of specifying mock auth and function name for the args (e.g. a simple usage like `assert_invoke_auth!(client.rotate_signers(...))`

milapsheth:
PR scope should be something like `test: ...`

contracts/axelar-gateway/src/testutils.rs
ahramy: make sense. updated.

contracts/axelar-gateway/src/test_gateway.rs
```diff
@@ -414,22 +415,17 @@ fn rotate_signers_bypass_rotation_delay_fail_if_not_operator() {
...
-                    new_signers.signers.clone(),
-                    proof.clone(),
-                    bypass_rotation_delay,
-                )
-                    .into_val(&env),
-                sub_invokes: &[],
-            },
-        }])
-        .rotate_signers(&new_signers.signers, &proof, &bypass_rotation_delay);
+    assert_invoke_auth!(
```
milapsheth, line 418: make the usage more natural similar to what I suggested, the macro can parse the input as a function call and figure out how to add the mock auth based on it. You can even ask Claude to generate the macro for you given the usage and the previous code that needs to be generated

ahramy: I agree that the usage should be more natural. updated the macro and code, please take look.

packages/axelar-soroban-std/src/error.rs
```diff
@@ -104,3 +104,33 @@ macro_rules! assert_some {
...
+            .mock_auths(&[MockAuth {
+                address: &$caller,
+                invoke: &MockAuthInvoke {
+                    contract: &$client.address,
+                    fn_name: stringify!($method),
+                    args: ($($arg.clone(),)*).into_val(&$client.env),
+                    sub_invokes: &[],
+                },
+            }])
+            .$method($($arg),*);
```
milapsheth, line 123:
```suggestion
            .try_$method($($arg),*);
```
can you add the `try` part in the macro, so the usage just looks like `client.rotate_signers(..)`

milapsheth (approved):

contracts/axelar-gateway/src/test_gateway.rs
```diff
@@ -470,28 +459,17 @@ fn transfer_operatorship() {
...
-            invoke: &MockAuthInvoke {
-                contract: &contract_id,
-                fn_name: "transfer_operatorship",
-                args: (&new_operator,).into_val(&env),
-                sub_invokes: &[],
-            },
-        }])
-        .transfer_operatorship(&new_operator);
+
+    assert_invoke_auth!(not_owner, client.try_transfer_operatorship(&not_owner));
```
milapsheth, line 472:
```suggestion
    assert_invoke_auth_err!(not_owner, client.try_transfer_operatorship(&not_owner));
```
Also convert the tests where the auth is meant to be successful to a macro as well

ahramy: got it. updated.

packages/axelar-soroban-std/src/error.rs
ahramy: I’ve tried different approaches, but I’m not sure this will work. Due to the logic comparing the `call_result` to `Err`, it’s producing an error.

```
match call_result {
            Err(_) => {
            }
            Ok(_) => panic!("Expected error, but got Ok result."),
        }
```

ahramy: I’ve tried different approaches, but I’m not sure this will work. Due to the logic matching the `call_result` to `Err`, it’s producing an error if I try to use `try_$method`.

PR #53, opened 2024-10-31
feat: use build scripts to generate binary contract data

hydrobeam:
closes AXE-6324

## Summary

use build scripts to generate the binary contract data used in the `upgrade.rs` test. there are two build scripts, one under `integration_test` and one under `axelar_gateway`

The `integration_test` build script sets up some preamble and calls `stellar contract build` (+ some args) which itself just calls a cargo command. It sets an environment variable which is checked in the `gateway` build script to override the `CARGO_PKG_VERSION` environment variable which is used to set the contract version. It cannot be set directly from an external source as cargo overrides it.

The odd looking `println!()` calls allow us to override the env vars cargo sets during compilation.

## Misc

Currently, the build script generates a new directory called `_artefacts` in the crate root of `integration_tests`, which is a little weird for a build script to do. I don't think it's possible to get the `contractimport()` calls to handle the `$OUT_DIR` env var which is how I'd ideally approach this.


.gitignore
```diff
@@ -3,6 +3,9 @@
 target
 debug
 
+# generated by build script, used for testing
+_artefacts
```
ahramy, line 7: I think just `artifacts` would be fine, no need `_`

ahramy: Do we want to include the initial version of `axelar_gateway_new.wasm` and `axelar_gateway_old.wasm` for testing? Otherwise for someone cloned the repo, running `build.rs` becomes mandatory for `cargo test` to pass, which may be unclear. Alternatively, we could update the `README.md` to have clear instructions

contracts/axelar-gateway/build.rs
```diff
@@ -0,0 +1,8 @@
+fn main() {
+    // trick taken from:
+    // https://github.com/rust-lang/cargo/issues/6583#issuecomment-1259871885
+    if let Ok(val) = std::env::var("GATEWAY_CONTRACT_TEST_VERSION") {
+        println!("cargo:rustc-env=CARGO_PKG_VERSION={}", val);
```
ahramy, line 5: 👍

integration-tests/build.rs
```diff
@@ -0,0 +1,37 @@
+use std::{env, path::Path, process::Command};
+
+// make sure these are in sync with the values in tests/upgrade.rs
+const OLD_CONTRACT_VERSION: &str = "0.1.0";
```
ahramy, line 4: We should have clear instructions in the `README.md` on how to update this version.

ahramy (commented):
Could we combine `build.rs` with `optimize.rs`? Otherwise, it seems like we have to run two scripts, which could be cumbersome. Handling both build and optimization in one script would be ideal.

PR #56, opened 2024-11-06
feat(its): add skeleton code for interchain token

ahramy:
[AXE-6461](https://axelarnetwork.atlassian.net/browse/AXE-6461)

[AXE-6461]: https://axelarnetwork.atlassian.net/browse/AXE-6461?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


AttissNgo (approved):

hydrobeam (changes requested):
The tests should be integration tests under `tests/test.rs` to match the pattern we're following for other contracts so that we're explicitly testing contract endpoints / the public API. If there's a test that depends on internal behaviour then that can be a unit test.

contracts/interchain-token/src/contract.rs
```diff
@@ -0,0 +1,41 @@
...
+        env.storage().instance().set(&DataKey::Initialized, &true);
+
+        env.storage().instance().set(&DataKey::Owner, &owner);
+
+        Ok(())
+    }
+
+    pub fn owner(env: &Env) -> Address {
+        env.storage().instance().get(&DataKey::Owner).unwrap()
+    }
```
hydrobeam, line 31: since Owner is set in `initialize`, this should be fine to avoid a panic
```suggestion
    pub fn owner(env: &Env) -> Result<Address, ContractError> {
        env.storage().instance().get(&DataKey::Owner).ok_or(ContractError::NotInitialized)
    }
```
i've followed this pattern in other locations

contracts/interchain-token/tests/test.rs
```diff
@@ -0,0 +1,86 @@
...
+fn setup_env<'a>() -> (Env, Address, InterchainTokenClient<'a>) {
+    let env = Env::default();
+
+    let contract_id = env.register_contract(None, InterchainToken);
+    let client = InterchainTokenClient::new(&env, &contract_id);
+
+    (env, contract_id, client)
+}
+
+fn initialize(_env: &Env, client: &InterchainTokenClient, owner: Address) {
```
hydrobeam, line 22: Feels a little overkill to have a function for a single line (unless we expect to add more here eventually?)

hydrobeam (changes requested):

hydrobeam (approved):

PR #59, opened 2024-11-08
feat(gas-service): add gas function in gas service

ahramy:
[AXE-6459](https://axelarnetwork.atlassian.net/browse/AXE-6459)
- EVM Reference: https://github.com/axelarnetwork/axelar-cgp-solidity/blob/b323ae613a4286e4849bbf81dd8922e80fe4fdca/contracts/gas-service/AxelarGasService.sol#L329

[AXE-6459]: https://axelarnetwork.atlassian.net/browse/AXE-6459?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/axelar-gas-service/src/contract.rs
+++ b/contracts/axelar-gas-service/src/contract.rs
@@ -66,6 +66,30 @@ impl AxelarGasService {
         Ok(())
     }
 
+    /// Add additional gas payment after initiating a cross-chain call.
+    pub fn add_gas(
+        env: Env,
+        sender: Address,
+        message_id: String,
+        refund_address: Address,
+        token: Token,
+    ) -> Result<(), ContractError> {
+        sender.require_auth();
+
+        ensure!(token.amount > 0, ContractError::InvalidAmount);
+
+        token::Client::new(&env, &token.address).transfer_from(
+            &env.current_contract_address(),
+            &sender,
+            &env.current_contract_address(),
+            &token.amount,
+        );
+
+        event::gas_added(&env, message_id, refund_address, token);
+
+        Ok(())
+    }
+
     /// Allows the `gas_collector` to collect accumulated fees from the contract.
     ///
     /// Only callable by the `gas_collector`.
--- a/contracts/axelar-gas-service/src/contract.rs
+++ b/contracts/axelar-gas-service/src/contract.rs
@@ -71,8 +71,8 @@ impl AxelarGasService {
         env: Env,
         sender: Address,
         message_id: String,
-        refund_address: Address,
         token: Token,
+        refund_address: Address,
     ) -> Result<(), ContractError> {
         sender.require_auth();
 
@@ -85,7 +85,7 @@ impl AxelarGasService {
             &token.amount,
         );
 
-        event::gas_added(&env, message_id, refund_address, token);
+        event::gas_added(&env, message_id, token, refund_address);
 
         Ok(())
     }
--- a/contracts/axelar-gas-service/src/event.rs
+++ b/contracts/axelar-gas-service/src/event.rs
@@ -22,12 +22,12 @@ pub fn gas_paid_for_contract_call(
     );
 }
 
-pub fn gas_added(env: &Env, message_id: String, refund_address: Address, token: Token) {
+pub fn gas_added(env: &Env, message_id: String, token: Token, refund_address: Address) {
     let topics = (
         Symbol::new(env, "gas_added"),
         message_id,
-        refund_address,
         token,
+        refund_address,
     );
     env.events().publish(topics, ());
 }
--- a/contracts/axelar-gas-service/src/event.rs
+++ b/contracts/axelar-gas-service/src/event.rs
@@ -22,6 +22,16 @@ pub fn gas_paid_for_contract_call(
     );
 }
 
+pub fn gas_added(env: &Env, message_id: String, refund_address: Address, token: Token) {
+    let topics = (
+        Symbol::new(env, "gas_added"),
+        message_id,
+        refund_address,
+        token,
+    );
+    env.events().publish(topics, ());
+}
+
 pub fn refunded(env: &Env, message_id: String, receiver: Address, token: Token) {
     let topics = (
         Symbol::new(env, "gas_refunded"),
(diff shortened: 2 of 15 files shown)


contracts/axelar-gas-service/tests/test.rs
```diff
@@ -181,6 +181,91 @@ fn pay_gas_for_contract_call() {
...
+    let (env, contract_id, _gas_collector, client) = setup_env();
+
+    let asset = env.register_stellar_asset_contract_v2(Address::generate(&env));
+    let sender: Address = Address::generate(&env);
+    let message_id = String::from_str(
+        &env,
+        &format!(
+            "{}-{}",
+            "0xfded3f55dec47250a52a8c0bb7038e72fa6ffaae33562f77cd2b629ef7fd424d", 0
+        ),
```
hydrobeam, line 195: would be cleaner to just hardcode the string instead of using `format!()`. 

Also, im just curious how you got to this message id, was it random?

ahramy: Yeah, it's just a random value, I grabbed it from the unit test `fn refund()` in this file. I think the formatting is intended to make it clear that we are explicitly combining the tx hash and index. [In the EVM implementation](https://github.com/axelarnetwork/axelar-cgp-solidity/blob/b323ae613a4286e4849bbf81dd8922e80fe4fdca/contracts/gas-service/AxelarGasService.sol#L329), they were separated, but in Stellar, we combined them into message_id, so I guess we want to explicitly test using hashtx + index.

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -66,6 +66,30 @@ impl AxelarGasService {
         Ok(())
     }
 
+    /// Add additional gas payment after initiating a cross-chain call.
+    pub fn add_gas(
+        env: Env,
+        sender: Address,
+        message_id: String,
+        refund_address: Address,
+        token: Token,
```
milapsheth, line 75:
```suggestion
        token: Token,
        refund_address: Address,
```
match ordering from EVM gas service

contracts/axelar-gas-service/src/event.rs
```diff
@@ -22,6 +22,16 @@ pub fn gas_paid_for_contract_call(
     );
 }
 
+pub fn gas_added(env: &Env, message_id: String, refund_address: Address, token: Token) {
+    let topics = (
+        Symbol::new(env, "gas_added"),
+        message_id,
+        refund_address,
```
milapsheth, line 29: same

contracts/axelar-gas-service/tests/test.rs
```diff
@@ -181,6 +181,91 @@ fn pay_gas_for_contract_call() {
...
+        address: asset.address(),
+        amount: gas_amount,
+    };
+
+    let token_client = TokenClient::new(&env, &asset.address());
+    StellarAssetClient::new(&env, &asset.address()).mint(&sender, &gas_amount);
+
+    let expiration_ledger = &env.ledger().sequence() + 200;
+
+    // approve token spend before invoking `add_gas`
```
milapsheth, line 245:
```suggestion
```
i think it's implied, since it's similar to ERC20

contracts/axelar-gas-service/tests/test.rs
```diff
@@ -181,6 +181,91 @@ fn pay_gas_for_contract_call() {
...
+    token_client.approve(&sender, &contract_id, &gas_amount, &expiration_ledger);
+
+    assert_eq!(token_client.allowance(&sender, &contract_id), gas_amount);
+
```
milapsheth, line 213: not really needed since it'll fail due to validation right away

contracts/axelar-gas-service/tests/test.rs
```diff
@@ -181,6 +181,91 @@ fn pay_gas_for_contract_call() {
...
+    let token_client = TokenClient::new(&env, &asset.address());
+    StellarAssetClient::new(&env, &asset.address()).mint(&sender, &gas_amount);
+
+    let expiration_ledger = &env.ledger().sequence() + 200;
+
+    // approve token spend before invoking `add_gas`
+    token_client.approve(&sender, &contract_id, &gas_amount, &expiration_ledger);
+
+    assert_eq!(token_client.allowance(&sender, &contract_id), gas_amount);
```
milapsheth, line 248: we don't need to test that `token.approve` works in this test

contracts/axelar-gas-service/tests/test.rs
```diff
@@ -181,6 +181,91 @@ fn pay_gas_for_contract_call() {
     );
 }
 
+#[test]
+fn fail_add_gas_zero_gas_amount() {
```
milapsheth, line 185: add unit test for failure if approval wasn't set before calling add_gas

contracts/axelar-gas-service/src/event.rs
ahramy: updated

contracts/axelar-gas-service/tests/test.rs
ahramy: removed comment

contracts/axelar-gas-service/tests/test.rs
ahramy: make sense, updated.

contracts/axelar-gas-service/tests/test.rs
ahramy: make sense, updated.

contracts/axelar-gas-service/tests/test.rs
ahramy: updatd.

contracts/axelar-gas-service/tests/test.rs
```diff
@@ -126,6 +126,39 @@ fn fail_pay_gas_zero_gas_amount() {
     );
 }
 
+#[test]
+fn fail_pay_gas_no_approved_allowance() {
```
milapsheth, line 130:
```suggestion
fn fail_pay_gas_zero_gas_amount() {
```

milapsheth (approved):

PR #64, opened 2024-11-12
feat: upgrade soroban-sdk version to v22.0.0-rc.3

hydrobeam:
closes https://axelarnetwork.atlassian.net/browse/AXE-6571
## Summary

Testnet upgraded to v22 today, we need to keep the version compatible with the network so that contracts can execute

from [the upgrade guide](https://stellar.org/blog/developers/protocol-22-upgrade-guide), stable builds of the software will be available on November 19, so we'll shift off the "rc" version then

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2021"
 rust-version = "1.76.0"
 
 [workspace.dependencies]
-soroban-sdk = { version = "21.7.6" }
+soroban-sdk = { version = "22.0.0-rc.3" }
 axelar-soroban-std = { version = "^0.1.0", path = "packages/axelar-soroban-std" }
 axelar-gas-service = { version = "^0.1.0", path = "contracts/axelar-gas-service" }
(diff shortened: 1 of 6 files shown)


Cargo.toml
```diff
@@ -8,10 +8,10 @@ members = [
 
 [workspace.package]
 edition = "2021"
-rust-version = "1.76.0"
+rust-version = "1.79.0"
```
ahramy, line 11: Why don't we use latest stable version: `1.82.0`?
```
rustup update stable
rustc --version     
rustc 1.82.0 (f6e511eec 2024-10-15)
```

hydrobeam: rust 1.82 won't work because of https://github.com/stellar/rs-soroban-sdk/pull/1371 . We could do 1.81 but i just decided to go with the minimum version that will probably work instead of the newest possible one as long as we don't need any features from the later version.

milapsheth (approved):

ahramy (approved):

PR #65, opened 2024-11-13
feat(ITS): add ITS message ABI encoding/decoding support

AttissNgo:
[AXE-6065](https://axelarnetwork.atlassian.net/browse/AXE-6065)

[AXE-6065]: https://axelarnetwork.atlassian.net/browse/AXE-6065?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,519 @@
...
+        let payload_array = binding.as_slice();
+
+        let message_type = match MessageType::abi_decode(&payload_array[0..32], true) {
+            Ok(value) => value,
+            Err(_) => return Err(MessageError::InvalidMessageType),
+        };
+
+        let message = match message_type {
+            MessageType::InterchainTransfer => {
+                let decoded = match InterchainTransfer::abi_decode_params(payload_array, true) {
```
ahramy, line 110: Could this be written as a one-liner?
```suggestion
               let decoded = InterchainTransfer::abi_decode_params(payload_array, true).map_err(|_| MessageError::AbiDecodeFailed)?;
```

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,519 @@
...
+
+                Self::InterchainTransfer(types::InterchainTransfer {
+                    token_id: BytesN::from_array(env, &decoded.tokenId.into()),
+                    source_address: Bytes::from_slice(env, decoded.sourceAddress.as_ref()),
+                    destination_address: Bytes::from_slice(
+                        env,
+                        decoded.destinationAddress.as_ref(),
+                    ),
+                    amount: convert_to_i128(decoded.amount)?,
+                    data: match decoded.data.len() {
```
ahramy, line 123: Could this be written as a one-liner?
```suggestion
                    data: (!decoded.data.is_empty()).then(|| Bytes::from_slice(env, decoded.data.as_ref())),
```

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,519 @@
...
+                        0 => None,
+                        _ => Some(Bytes::from_slice(env, decoded.data.as_ref())),
+                    },
+                })
+            }
+            MessageType::DeployInterchainToken => {
+                let decoded = match DeployInterchainToken::abi_decode_params(payload_array, true) {
```
ahramy, line 130: Could this be written as a one-liner?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,519 @@
...
+                    Ok(value) => value,
+                    Err(_) => return Err(MessageError::AbiDecodeFailed),
+                };
+
+                Self::DeployInterchainToken(types::DeployInterchainToken {
+                    token_id: BytesN::from_array(env, &decoded.tokenId.into()),
+                    name: String::from_str(env, &decoded.name),
+                    symbol: String::from_str(env, &decoded.symbol),
+                    decimals: decoded.decimals.into(),
+                    minter: match decoded.minter.len() {
```
ahramy, line 140: Could this be written as a one-liner?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,519 @@
...
+        Bytes::from_slice(env, &msg)
+    }
+
+    pub fn abi_decode(env: &Env, payload: &Bytes) -> Result<Self, MessageError> {
+        ensure!(payload.len() >= 32, MessageError::InsufficientMessageLength);
+
+        let binding = payload.to_alloc_vec();
+        let payload_array = binding.as_slice();
+
+        let message_type = match MessageType::abi_decode(&payload_array[0..32], true) {
```
ahramy, line 184: Could this be written as a one-liner?
```suggestion
        let message_type = MessageType::abi_decode(&payload_array[0..32], true).map_err(|_| MessageError::InvalidMessageType)?;
```

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,519 @@
...
+            Ok(value) => value,
+            Err(_) => return Err(MessageError::InvalidMessageType),
+        };
+
+        let message = match message_type {
+            MessageType::SendToHub => {
+                let decoded = match SendToHub::abi_decode_params(payload_array, true) {
```
ahramy, line 191: Could this be written as a one-liner?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,519 @@
...
+                Self::SendToHub(types::SendToHub {
+                    destination_chain: String::from_str(env, &decoded.destination_chain),
+                    message: Message::abi_decode(
+                        env,
+                        &Bytes::from_slice(env, decoded.message.as_ref()),
+                    )?,
+                })
+            }
+            MessageType::ReceiveFromHub => {
+                let decoded = match ReceiveFromHub::abi_decode_params(payload_array, true) {
```
ahramy, line 205: Could this be written as a one-liner?

ahramy (commented):
ABI encoding/decoding could be quite challenging, but you managed it very well. Well done!

ahramy (approved):

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+                    .map_err(|_| MessageError::AbiDecodeFailed)?;
+
+                Self::InterchainTransfer(types::InterchainTransfer {
+                    token_id: BytesN::from_array(env, &decoded.tokenId.into()),
+                    source_address: Bytes::from_slice(env, decoded.sourceAddress.as_ref()),
+                    destination_address: Bytes::from_slice(
+                        env,
+                        decoded.destinationAddress.as_ref(),
+                    ),
+                    amount: convert_to_i128(decoded.amount)?,
```
milapsheth, line 118:
```suggestion
                    amount: to_i128(decoded.amount)?,
```

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+                })
+            }
+            _ => return Err(MessageError::InvalidMessageType),
+        };
+
+        Ok(message)
+    }
+}
+
+fn soroban_string_to_string(soroban_string: String) -> StdString {
```
milapsheth, line 209:
```suggestion
fn to_std_string(soroban_string: String) -> StdString {
```

milapsheth, line 209: does soroban_sdk not provide a From conversion between their own string types?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+    let length = soroban_string.len() as usize;
+    let mut bytes = vec![0u8; length];
+    soroban_string.copy_into_slice(&mut bytes);
+    StdString::from_utf8(bytes).expect("Invalid UTF-8 sequence")
+}
+
+fn convert_to_i128(decoded_value: Uint<256, 4>) -> Result<i128, MessageError> {
+    let slice = decoded_value.as_le_slice();
```
milapsheth, line 217:
```suggestion
    let slice = decoded_value.to_le_bytes();
```
perhaps?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+
+    if !slice[16..].iter().all(|&b| b == 0) {
```
milapsheth, line 219:
```suggestion
    if i128::from_le_bytes(value_bytes[16..]) != 0 {
```
+ use `ensure!`

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+fn soroban_string_to_string(soroban_string: String) -> StdString {
+    let length = soroban_string.len() as usize;
+    let mut bytes = vec![0u8; length];
+    soroban_string.copy_into_slice(&mut bytes);
+    StdString::from_utf8(bytes).expect("Invalid UTF-8 sequence")
+}
+
+fn convert_to_i128(decoded_value: Uint<256, 4>) -> Result<i128, MessageError> {
```
milapsheth, line 216:
```suggestion
fn convert_to_i128(value: Uint<256, 4>) -> Result<i128, MessageError> {
```

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+    let slice = decoded_value.as_le_slice();
+
+    if !slice[16..].iter().all(|&b| b == 0) {
+        return Err(MessageError::InvalidAmount);
+    }
+
+    let mut truncated = [0u8; 16];
+    truncated.copy_from_slice(&slice[..16]);
+
+    Ok(i128::from_le_bytes(truncated))
```
milapsheth, line 226:
```suggestion
    Ok(i128::from_le_bytes(value_bytes[..16]))
```

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloc::vec;
+    use soroban_sdk::{Bytes, BytesN, Env, String};
+    use std::vec::Vec;
+
+    const MAX_I128: i128 = 170141183460469231731687303715884105727 as i128;
```
milapsheth, line 236: use `i128::MAX`

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+    #[test]
+    fn abi_encode_decode() {
+        let env = Env::default();
+
+        let its_messages = vec![
+            types::Message::InterchainTransfer(types::InterchainTransfer {
+                token_id: BytesN::from_array(&env, &[1; 32]),
+                source_address: Bytes::from_slice(&env, &[2; 32]),
+                destination_address: Bytes::from_slice(&env, &[3; 32]),
+                amount: 9_876_543_210_123_456_789,
```
milapsheth, line 252: avoid hardcoding, retrieve it instead

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,503 @@
...
+            }),
+        ];
+
+        for msg in its_messages {
+            let encoded = msg.clone().abi_encode(&env);
+            let decoded = Message::abi_decode(&env, &encoded);
+            assert_eq!(msg, decoded.unwrap());
+        }
+
+        let hub_messages = vec![
```
milapsheth, line 270: use the existing encode_decode tests if more vectors are needed instead of creating another one

contracts/interchain-token-service/src/lib.rs
```diff
@@ -8,6 +9,8 @@ mod types;
 pub mod contract;
 
 #[cfg(test)]
+#[macro_use]
 mod test;
+extern crate std;
```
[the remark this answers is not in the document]
AttissNgo: @milapsheth Chris discovered that the `goldie` macro definition for `assert_json` expects the `std` crate to be found at the root level. Therefore, this has been changed to 
```
#[cfg(test)]
extern crate std;
```
in `lib.rs`

contracts/interchain-token-service/src/types.rs
```diff
@@ -6,20 +6,23 @@ use soroban_sdk::{contracttype, Bytes, BytesN, String};
...
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum Message {
+    InterchainTransfer(InterchainTransfer),
+    DeployInterchainToken(DeployInterchainToken),
+}
+
+#[contracttype]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub enum HubMessage {
+    SendToHub(SendToHub),
```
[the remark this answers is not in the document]
AttissNgo: @milapsheth Soroban only allows unit, tuple, and integer variants in enums. It won't support named fields like the ITS Hub example. I wrapped these in an enum in order to be able to define an `impl` block on the `HubMessage` type for encoding and decoding. Is there a better way to accomplish this without wrapping the message structs?

contracts/interchain-token-service/src/types.rs
```diff
@@ -28,8 +31,8 @@ pub struct InterchainTransfer {
     pub token_id: BytesN<32>,
     pub source_address: Bytes,
     pub destination_address: Bytes,
-    pub amount: u128,
-    pub data: Bytes,
+    pub amount: i128,
```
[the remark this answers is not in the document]
AttissNgo: Will do in separate PR - ticket created.

contracts/interchain-token-service/src/abi.rs
AttissNgo: @milapsheth  There is a `to_string()` method provided, but this is where we run into compiler issues when building the wasm, since implementing it requires using the `Display` trait, which means pulling in `std::fmt`. Is implementing `Display` for Soroban's `String` type preferable to a custom method?

contracts/interchain-token-service/src/abi.rs
AttissNgo: Using `to_le_bytes` requires a type annotation: "type annotations needed for `[u8; _]`"

Is there an advantage to doing it this way?

contracts/interchain-token-service/src/abi.rs
AttissNgo: @milapsheth I've been unable to make this conversion work without using slices. Do we need this to work with an owned value `[u8; 32]`?

contracts/interchain-token-service/src/types.rs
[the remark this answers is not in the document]
milapsheth: I like how using `#[contracttype]` hides the fact that we have dead code lol

contracts/interchain-token-service/src/abi.rs
milapsheth: We can't implement a foreign crate on a foreign type. I think soroban sdk should have gated this under testutils feature since it has an std dependency. Maybe report the issue to them?

contracts/interchain-token-service/src/abi.rs
milapsheth: nah, it's fine if we can't get it to work

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,464 @@
...
+                name,
+                symbol,
+                decimals,
+                minter,
+            }) => DeployInterchainToken {
+                messageType: MessageType::DeployInterchainToken.into(),
+                tokenId: FixedBytes::<32>::new(token_id.into()),
+                name: to_std_string(name),
+                symbol: to_std_string(symbol),
+                decimals: decimals.try_into().expect("failed to convert"),
```
milapsheth, line 89: Why is this needed? `decimals` should be `u8` in both types

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,464 @@
...
+                minter: into_vec(minter).into(),
+            }
+            .abi_encode_params(),
+        };
+        Bytes::from_slice(env, &msg)
```
milapsheth, line 94: Similar to `to_alloc_vec`, can we ask Stellar team to add a `from_alloc_vec` for completeness?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,464 @@
...
+
+        Ok(message)
+    }
+}
+
+fn to_std_string(soroban_string: String) -> StdString {
+    let length = soroban_string.len() as usize;
+    let mut bytes = vec![0u8; length];
+    soroban_string.copy_into_slice(&mut bytes);
+    StdString::from_utf8(bytes).expect("Invalid UTF-8 sequence")
```
milapsheth, line 210: test this failure since it can be due to user input

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,464 @@
...
+    let mut bytes_to_convert = [0; 16];
+    bytes_to_remove.copy_from_slice(&slice[16..]);
+    bytes_to_convert.copy_from_slice(&slice[..16]);
+
+    ensure!(
+        i128::from_le_bytes(bytes_to_remove) == 0,
+        MessageError::InvalidAmount
+    );
+
+    Ok(i128::from_le_bytes(bytes_to_convert))
```
milapsheth, line 226: also ensure this is non-negative and test this behaviour

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,464 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use alloc::vec;
+    use soroban_sdk::{Bytes, BytesN, Env, String};
+    use std::vec::Vec;
+
+    const MAX_I128: i128 = i128::MAX as i128;
```
milapsheth, line 254: use `i128::MAX` directly
```suggestion
```

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,464 @@
...
+
+    fn bytes_from_hex(env: &Env, hex_string: &str) -> Bytes {
+        let bytes_vec: Vec<u8> = hex::decode(hex_string).unwrap();
+        Bytes::from_slice(env, &bytes_vec)
+    }
+
+    #[test]
+    fn uint256_to_i128() {
+        let amount: i128 = 9_876_543_210_123_456_789;
```
milapsheth, line 263: compute this instead of hardcoding. applies everywhere

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,464 @@
...
+    use super::*;
+    use alloc::vec;
+    use soroban_sdk::{Bytes, BytesN, Env, String};
+    use std::vec::Vec;
+
+    const MAX_I128: i128 = i128::MAX as i128;
+
+    fn bytes_from_hex(env: &Env, hex_string: &str) -> Bytes {
+        let bytes_vec: Vec<u8> = hex::decode(hex_string).unwrap();
+        Bytes::from_slice(env, &bytes_vec)
```
milapsheth, line 258: check codecov coverage info for any missing coverage

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,508 @@
...
+    StdString::from_utf8(bytes).expect("Invalid UTF-8 sequence")
+}
+
+fn to_i128(value: Uint<256, 4>) -> Result<i128, MessageError> {
+    let slice = value.as_le_slice();
+
+    let mut bytes_to_remove = [0; 16];
+    let mut bytes_to_convert = [0; 16];
+    bytes_to_remove.copy_from_slice(&slice[16..]);
+    bytes_to_convert.copy_from_slice(&slice[..16]);
```
cgorenflo, line 219: switch the order, so you go `..16` and the `16..`

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,508 @@
...
+}
+
+fn to_std_string(soroban_string: String) -> StdString {
+    let length = soroban_string.len() as usize;
+    let mut bytes = vec![0u8; length];
+    soroban_string.copy_into_slice(&mut bytes);
+    StdString::from_utf8(bytes).expect("Invalid UTF-8 sequence")
+}
+
+fn to_i128(value: Uint<256, 4>) -> Result<i128, MessageError> {
```
cgorenflo, line 213: why to we have to go to i128 instead of u128?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    #[test]
+    #[should_panic(expected = "Invalid UTF-8 sequence")]
+    fn to_std_string_panics_invalid_utf8_bytes() {
+        let env = Env::default();
+        let invalid_sequence = vec![0x80, 0x81, 0x82, 0x83];
+        let invalid_string = String::from_bytes(&env, &invalid_sequence);
+        to_std_string(invalid_string);
+    }
```
cgorenflo, line 270: I would definitely prefer more string tests, conversions of user input is scary

cgorenflo: definitely check some characters with variable length encoding

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,508 @@
...
+        Bytes::from_slice(env, &bytes_vec)
+    }
+
+    #[test]
+    #[should_panic(expected = "Invalid UTF-8 sequence")]
+    fn to_std_string_panics_invalid_utf8_bytes() {
+        let env = Env::default();
+        let invalid_sequence = vec![0x80, 0x81, 0x82, 0x83];
+        let invalid_string = String::from_bytes(&env, &invalid_sequence);
+        to_std_string(invalid_string);
```
cgorenflo, line 269: make sure it actually panics on this line, not the invalid_string creation

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,508 @@
...
+    }
+
+    #[test]
+    fn to_i128_fails_overflow() {
+        let overflow: Uint<256, 4> = Uint::from(MAX_I128) + Uint::from(1);
+
+        let result = to_i128(overflow);
+
+        assert!(matches!(result, Err(MessageError::InvalidAmount)));
+    }
```
cgorenflo: it's kind of covered, but just to be certain add a test for `u128::Max` as well

contracts/interchain-token-service/src/types.rs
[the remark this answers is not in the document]
cgorenflo: what's the reason for i128 in the first place?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,508 @@
...
+    InvalidMessageType = 1,
+    AbiDecodeFailed = 2,
+    InvalidAmount = 3,
+}
+
+sol! {
+    enum MessageType {
+        InterchainTransfer,
+        DeployInterchainToken,
+        DeployTokenManager, // note, this case is not supported by the ITS hub
```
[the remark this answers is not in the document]
AttissNgo: Yes, exactly. The issue is that the golden tests for ABI encoding are no longer "golden" if we change the enum order.

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,508 @@
...
+                })
+            }
+            _ => return Err(MessageError::InvalidMessageType),
+        };
+
+        Ok(message)
+    }
+}
+
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
AttissNgo: The compiler complains about unused code since these have not been implemented in the ITS contract yet. I have created a ticket to remove the `#[allow(dead_code)]` once the ITS contract is actually using the HubMessage encode/decode methods.

contracts/interchain-token-service/src/abi.rs
AttissNgo: i128 is the standard type for token amounts on Stellar. It has to do with asset clawbacks/burning, but I agree that it's confusing.

contracts/interchain-token-service/src/types.rs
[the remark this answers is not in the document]
AttissNgo: No, this was just (perhaps incorrect) intuition on my part. I'm fine to define the message types and the ABI encode/decode methods in the same file. @milapsheth is this okay?

Cargo.toml
```diff
@@ -17,8 +13,12 @@ axelar-gas-service = { version = "^0.1.0", path = "contracts/axelar-gas-service"
...
 axelar-operators = { version = "^0.1.0", path = "contracts/axelar-operators" }
 interchain-token-service = { version = "^0.1.0", path = "contracts/interchain-token-service" }
-interchain-token = { version = "^0.1.0", path = "contracts/interchain-token" }
-example = { version = "^0.1.0", path = "contracts/example" }
+alloy-primitives = { version = "0.7.6", default-features = false, features = [
+  "std",
+] }
+alloy-sol-types = { version = "0.7.6", default-features = false, features = [
+  "std",
+] }
```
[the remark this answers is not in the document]
AttissNgo: The stellar build command will build the wasm, but haven't tried deploying yet. If this does cause an issue, we'll have to rethink the approach to encoding/decoding here. @milapsheth how can we verify that the `std` features here are okay to use?

contracts/interchain-token-service/src/abi.rs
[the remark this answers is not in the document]
cgorenflo: but if it's just about internal tests I would vote for removing unsupported types, adding them back in later and adjusting the tests. If it's about compatibility with other platforms that's a different matter. On that note, why do we need solidity types here in the first place?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -0,0 +1,508 @@
...
+
+        Ok(message)
+    }
+}
+
+fn to_std_string(soroban_string: String) -> StdString {
+    let length = soroban_string.len() as usize;
+    let mut bytes = vec![0u8; length];
+    soroban_string.copy_into_slice(&mut bytes);
+    StdString::from_utf8(bytes).expect("Invalid UTF-8 sequence")
```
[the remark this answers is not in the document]
AttissNgo: Since the `copy_into_slice` method panics if the output slice and string are of different lengths, is that guarantee enough that the conversion is lossless since it's basically a byte-level copy?

contracts/interchain-token-service/src/abi.rs
```diff
@@ -258,12 +259,67 @@ mod tests {
...
+
+        for sequence in invalid_sequences {
+            let result = to_std_string(sequence);
+            assert!(matches!(result, Err(MessageError::InvalidUtf8)));
+        }
+    }
+
+    #[test]
+    #[should_panic]
+    fn soroban_string_copy_to_slice_panics_on_missized_array() {
```
milapsheth, line 314: this test isn't really relevant since it's testing the behaviour of `copy_into_slice` and not a function that we wrote

contracts/interchain-token-service/src/abi.rs
```diff
@@ -391,7 +448,7 @@ mod tests {
         goldie::assert_json!(encoded);
 
         for original in cases {
-            let encoded = original.clone().abi_encode(&env);
+            let encoded = original.clone().abi_encode(&env).unwrap();
```
milapsheth, line 451:
```suggestion
            let encoded = assert_ok!(original.clone().abi_encode(&env));
```
use assert_ok from axelar-soroban-std to emphasize that we're checking that this function doesn't error

contracts/interchain-token-service/src/abi.rs
```diff
@@ -478,6 +535,7 @@ mod tests {
                     original
                         .clone()
                         .abi_encode(&env)
+                        .unwrap()
```
milapsheth, line 538: same everywhere. `.unwrap` can be used when dropping error for some helper call we know will succeed

contracts/interchain-token-service/src/abi.rs
```diff
@@ -171,35 +169,33 @@ impl HubMessage {
...
                         &Bytes::from_slice(env, decoded.message.as_ref()),
                     )?,
-                }
+                })
             }
-            _ => return Err(MessageError::InvalidMessageType),
-        };
-
-        Ok(message)
+            _ => Err(MessageError::InvalidMessageType),
```
milapsheth, line 198: is this case being covered in tests?

milapsheth (approved):

contracts/interchain-token-service/src/abi.rs
AttissNgo: I realize this. Chris asked me to run a test to confirm that this works as expected, but I can remove.

PR #69, opened 2024-11-14
feat(operators): add contract constructor

hydrobeam:
closes AXE-6572

## Summary

v22 of `soroban-sdk` added support for using constructors.

Most of this was based on the [migration guide](https://docs.rs/soroban-sdk/22.0.0-rc.3.2/soroban_sdk/_migrating/index.html) and the document which outlines how to define constructors (https://github.com/stellar/stellar-protocol/blob/5ef1b45cbab54f57c5ddfc98adb964f26b68bb3b/core/cap-0058.md)

`client.register_contract()` (which generates an address which you then initialize at) is deprecated in favour of `client.register(contract, constructor_args)` which instead runs the contract's `__constructor()`. In turn, `initialize()` was renamed to `__constructor()`

This makes it so that `NotInitialized` and `AlreadyInitialized` errors can no longer occur when contracts are invoked this way since `__constructor()` can only be called by the host environment and you cannot generate a `client` to work with without running `register` and thus initializing.

 So I removed those variants and their corresponding tests.

-- from review
- removed DataKey::Initialized
- removing `DataKey::Initialized` no longer forces us to depend on the contract's private API, so the unit test was moved to be an integration test


contracts/axelar-operators/src/contract.rs
```diff
@@ -11,11 +11,7 @@ pub struct AxelarOperators;
...
 impl AxelarOperators {
     pub fn transfer_ownership(env: Env, new_owner: Address) -> Result<(), ContractError> {
-        let owner: Address = env
-            .storage()
-            .instance()
-            .get(&DataKey::Owner)
-            .ok_or(ContractError::NotInitialized)?;
+        let owner: Address = assert_some!(env.storage().instance().get(&DataKey::Owner));
 
         owner.require_auth();
```
milapsheth, line 16:
```suggestion
        Self::owner(&env).require_auth();
```
everywhere

contracts/axelar-operators/src/contract.rs
```diff
@@ -26,30 +22,17 @@ impl AxelarOperators {
         Ok(())
     }
 
-    pub fn owner(env: &Env) -> Result<Address, ContractError> {
-        env.storage()
-            .instance()
-            .get(&DataKey::Owner)
-            .ok_or(ContractError::NotInitialized)
+    pub fn owner(env: &Env) -> Address {
+        assert_some!(env.storage().instance().get(&DataKey::Owner))
```
milapsheth, line 26:
```suggestion
        env.storage().instance().get(&DataKey::Owner).expect("owner not found")
```
`assert`'s are meant for tests, use `expect` to make it clear that we can unwrap, and it's fine to panic otherwise

contracts/axelar-operators/src/contract.rs
```diff
@@ -26,30 +22,17 @@ impl AxelarOperators {
...
-    pub fn initialize(env: Env, owner: Address) -> Result<(), ContractError> {
-        ensure!(
-            env.storage()
-                .instance()
-                .get::<DataKey, bool>(&DataKey::Initialized)
-                .is_none(),
-            ContractError::AlreadyInitialized
-        );
-
+    pub fn __constructor(env: Env, owner: Address) {
```
milapsheth, line 33: create a PR to update deployment script to use the constructor

contracts/axelar-operators/tests/test.rs
```diff
@@ -222,33 +194,32 @@ fn execute() {
...
+            &client.address,
+            &symbol_short!("method"),
+            &Vec::new(&env)
+        ),
         ContractError::NotAnOperator
     );
 }
 
 #[test]
 #[should_panic(expected = "This method should fail")]
```
milapsheth, line 217: use try_* and check result?

contracts/axelar-operators/src/contract.rs
```diff
@@ -26,30 +22,17 @@ impl AxelarOperators {
...
 }
 
 #[contractimpl]
 impl AxelarOperators {
     /// Initialize the operators contract with an owner.
-    pub fn initialize(env: Env, owner: Address) -> Result<(), ContractError> {
-        ensure!(
-            env.storage()
-                .instance()
-                .get::<DataKey, bool>(&DataKey::Initialized)
```
milapsheth, line 44: remove `DataKey::Initialized`

contracts/axelar-operators/tests/test.rs
```diff
@@ -38,6 +41,16 @@ fn setup_env<'a>() -> (Env, AxelarOperatorsClient<'a>, Address) {
     (env, client, target_id)
 }
 
+#[test]
+fn initialize_operators() {
```
milapsheth, line 45:
```suggestion
fn register_operators() {
```

milapsheth (approved):

PR #73, opened 2024-11-15
feat(gateway): add contract constructor

hydrobeam:
closes https://axelarnetwork.atlassian.net/browse/AXE-6581

## Summary

similar to #69

most of the changes are to the tests

- `testutils::initialize` now returns the `contract_id` it generated since it doesn't depend on a `client` anymore (`contract_id` is needed  to make a `client`) 
- remove instances of initializing after setup_env (they are handled in `setup_env` now)
   - setup_env now requires passing in values for `num_signers` and `minimum_rotation_delay`. Where there was not already a value defined (i.e. tests that did not require initializing), defaulted to `setup_env(1, 5)`
- setup_env in `auth.rs` and `tests/test.rs` doesn't return `contract_id`, instead prefer using `client.address` where possible
- update `validate_message` to expect 1 event because of the event emitted while initializing

### DIFF
--- a/contracts/example/tests/test.rs
+++ b/contracts/example/tests/test.rs
@@ -5,7 +5,7 @@ use axelar_gas_service::contract::AxelarGasService;
 use axelar_gas_service::AxelarGasServiceClient;
 use axelar_gateway::testutils::{generate_proof, get_approve_hash, initialize, TestSignerSet};
 use axelar_gateway::types::Message;
-use axelar_gateway::{AxelarGateway, AxelarGatewayClient};
+use axelar_gateway::AxelarGatewayClient;
 use axelar_soroban_std::assert_last_emitted_event;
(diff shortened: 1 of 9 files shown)


contracts/example/tests/test.rs
```diff
@@ -16,14 +16,11 @@ use soroban_sdk::{
 };
 use soroban_sdk::{Bytes, Symbol};
 
-fn setup_gateway<'a>(env: &Env) -> (AxelarGatewayClient<'a>, Address, TestSignerSet) {
-    let gateway_id = env.register_contract(None, AxelarGateway);
+fn setup_gateway<'a>(env: &Env) -> (TestSignerSet, AxelarGatewayClient<'a>) {
```
milapsheth, line 19: export setup_gateway instead of initialize, so you can merge the two

PR #74, opened 2024-11-15
feat(interchain-token-service,interchain-token): add contract constructor

hydrobeam:
closes https://axelarnetwork.atlassian.net/browse/AXE-6583

## Summary

similar to #69 

only thing of note was that the `interchain-token` `ContractError` type only had two variants, both of which were removed. So I just commented it out (could just be deleted instead?)

### DIFF
--- a/contracts/interchain-token-service/src/contract.rs
+++ b/contracts/interchain-token-service/src/contract.rs
@@ -10,24 +10,12 @@ pub struct InterchainTokenService;
 
 #[contractimpl]
 impl InterchainTokenService {
-    pub fn initialize_its(env: Env, owner: Address) -> Result<(), ContractError> {
-        ensure!(
-            env.storage()
-                .instance()
-                .get::<DataKey, bool>(&DataKey::Initialized)
-                .is_none(),
-            ContractError::AlreadyInitialized
-        );
-
-        env.storage().instance().set(&DataKey::Initialized, &true);
-
+    pub fn __constructor(env: Env, owner: Address) {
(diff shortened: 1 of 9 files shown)


contracts/interchain-token-service/src/contract.rs
```diff
@@ -10,24 +10,15 @@ pub struct InterchainTokenService;
```
milapsheth, line 10: PR title scope should be `feat(interchain-token-service,interchain-token): ..`

contracts/interchain-token-service/src/test.rs
```diff
@@ -9,59 +9,38 @@ use soroban_sdk::testutils::{MockAuth, MockAuthInvoke};
...
-    (env, contract_id, client)
-}
-
-fn initialize(_env: &Env, client: &InterchainTokenServiceClient, owner: Address) {
-    client.initialize_its(&owner);
+    (env, client)
 }
 
 #[test]
 fn initialize_succeeds() {
```
milapsheth, line 22:
```suggestion
fn register_interchain_token_service() {
```
alternatively `register_contract_succeeds`

contracts/interchain-token/src/error.rs
```diff
@@ -1,9 +1,6 @@
-use soroban_sdk::contracterror;
+// use soroban_sdk::contracterror;
```
milapsheth, line 1: debug

milapsheth (approved):

PR #77, opened 2024-11-15
feat(upgrader): add generalized atomic upgrade and migration capabilities

cgorenflo:
AXE-6099


contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+        let contract_id = env.register(Upgrader, ());
+        let gateway_id = env.register(AxelarGateway, ());
+        let client = UpgraderClient::new(&env, &contract_id);
+        client.upgrade(
+            &gateway_id,
+            &String::from_str(&env, "1.0"),
+            &BytesN::from_array(&env, &[0; 32]),
+            &soroban_sdk::vec![&env],
+        );
+    }
```
cgorenflo, line 98: WIP

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
+use crate::error::ContractError;
+use soroban_sdk::{contract, contractimpl, Address, BytesN, Env, String, Symbol, Val};
+
```
ahramy, line 3: It would be great to convert the string to constants for maintainability and readability.

```
const MIGRATE: &str = "migrate";
const VERSION: &str = "version";
const UPGRADE: &str = "upgrade";

Symbol::new(&env, MIGRATE)
Symbol::new(&env, VERSION)
Symbol::new(&env, UPGRADE)
```

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+fn version_fn(env: &Env) -> Symbol {
+    Symbol::new(&env, "version")
+}
+
+fn assert_new_version_is_different(
+    env: &Env,
+    contract_address: &Address,
+    new_version: &String,
+) -> Result<(), ContractError> {
+    let cur_version: String = env.invoke_contract(
```
ahramy, line 47: perhaps define a function to get `current_version` of a contract and call it, as there is duplicated logic.
```
fn get_current_version(
    env: &Env,
    contract_address: &Address,
) -> Result<String, ContractError> {
    env.invoke_contract(
        contract_address,
        &version_fn(env),
        soroban_sdk::vec![env],
    ).map_err(|_| ContractError::InvocationError)
}

let cur_version = get_current_version(env, contract_address)?;
```

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+        Err(ContractError::SameVersion)
+    }
+}
+
+fn assert_new_version_matches_expected(
+    env: &Env,
+    contract_address: &Address,
+    new_version: &String,
+) -> Result<(), ContractError> {
+    let cur_version: String = env.invoke_contract(
```
ahramy, line 65: ```let cur_version = get_current_version(env, contract_address)?;```

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::contract::{Upgrader, UpgraderClient};
+    use axelar_gateway::AxelarGateway;
+    use soroban_sdk::{BytesN, Env, String};
+
+    const _WASM: &[u8] = include_bytes!("testdata/axelar_gateway.wasm");
+    #[test]
```
ahramy, line 85: nit: new line before #[test]

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::contract::{Upgrader, UpgraderClient};
+    use axelar_gateway::AxelarGateway;
+    use soroban_sdk::{BytesN, Env, String};
+
+    const _WASM: &[u8] = include_bytes!("testdata/axelar_gateway.wasm");
```
ahramy, line 84: `testdata/axelar_gateway.optimized.wasm` or `testdata/axelar_gateway.wasm`?

contracts/upgrader/src/error.rs
```diff
@@ -0,0 +1,8 @@
+use soroban_sdk::contracterror;
+
+#[contracterror]
+#[repr(u32)]
+pub enum ContractError {
+    SameVersion = 1,
+    VersionMismatch = 2,
```
ahramy, line 7:
```suggestion
    VersionMismatch = 2,
    InvocationError = 3,
```

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+#[contractimpl]
+impl Upgrader {
+    pub fn __constructor(_env: Env) {}
+    pub fn upgrade(
+        env: Env,
+        contract_address: Address,
+        new_version: String,
+        new_wasm_hash: BytesN<32>,
+        migration_data: soroban_sdk::Vec<Val>,
+    ) -> Result<(), ContractError> {
```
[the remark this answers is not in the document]
cgorenflo: I figured that would be part of the `upgrade` functions of the contracts, otherwise we would need to specify a unique interface that all contracts implement to make the `owner` available. That being said I still have to test if the auth call inside the upgrade function even works as expected

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+    contract_address: &Address,
+    new_version: &String,
+) -> Result<(), ContractError> {
+    let cur_version: String = env.invoke_contract(
+        &contract_address,
+        &version_fn(&env),
+        soroban_sdk::vec![&env],
+    );
+
+    if cur_version != *new_version {
```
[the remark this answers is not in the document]
cgorenflo: I considered this (that goes for all places where you suggested this), but decided that it's clearer to directly do the conversion from condition to Result, instead of transforming the bool into and Option and then from there into a Result. That being said I could abstract away the `then_some().ok_or()` into a `then_ok()` function if the more functional approach is preferred here.

contracts/axelar-gateway/src/contract.rs
```diff
@@ -258,10 +258,36 @@ impl AxelarGateway {
...
+        Self::assert_is_migrating(&env)?;
+
+        // Add migration logic here as needed
+
+        // DO NOT REMOVE THIS LINE, IT PREVENTS THE CONTRACT FROM BEING MIGRATED MULTIPLE TIMES
+        Self::complete_migration(&env);
+        Ok(())
+    }
+
+    fn assert_is_migrating(env: &Env) -> Result<(), ContractError> {
```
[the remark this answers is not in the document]
cgorenflo: see detailed response about bool to result

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,99 @@
...
+    contract_address: &Address,
+    new_version: &String,
+) -> Result<(), ContractError> {
+    let cur_version: String = env.invoke_contract(
+        &contract_address,
+        &version_fn(&env),
+        soroban_sdk::vec![&env],
+    );
+
+    if cur_version == *new_version {
```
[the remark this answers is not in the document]
cgorenflo: see detailed response about bool to result

contracts/upgrader/src/contract.rs
cgorenflo: tests are still completely in progress, I just wanted to check in what I had on Friday to showcase the logic

contracts/axelar-gateway/src/contract.rs
```diff
@@ -324,4 +344,12 @@ impl AxelarGateway {
             .instance()
             .extend_ttl(INSTANCE_TTL_THRESHOLD, INSTANCE_TTL_EXTEND_TO);
     }
+
+    fn start_migrating(env: &Env) {
```
milapsheth, line 348:
```suggestion
    fn start_migration(env: &Env) {
```
for consistency

contracts/axelar-gateway/src/contract.rs
```diff
@@ -258,10 +258,32 @@ impl AxelarGateway {
...
+        Self::start_migrating(&env);
+
+        Ok(())
+    }
+
+    pub fn migrate(env: Env, _migration_data: ()) -> Result<(), ContractError> {
+        // DO NOT REMOVE THIS LINE, IT PREVENTS THE CONTRACT FROM BEING MIGRATED MULTIPLE TIMES
+        Self::assert_is_migrating(&env)?;
+
+        // Add migration logic here as needed
```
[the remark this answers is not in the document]
cgorenflo: good idea

contracts/upgrader/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "upgrader"
+version = "0.1.0"
+edition = { workspace = true }
+
+[lib]
+crate-type = ["cdylib", "rlib"]
+
+[dependencies]
+axelar-gateway = { workspace = true }
```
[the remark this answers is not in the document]
cgorenflo: pay the test stuff no heed, nothings been cleaned up there yet

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,82 @@
+use crate::error::ContractError;
+use axelar_soroban_std::traits::ThenOk;
+use soroban_sdk::{contract, contractimpl, Address, BytesN, Env, String, Symbol, Val};
+
+const MIGRATE: &'static str = "migrate";
+const UPGRADE: &'static str = "upgrade";
+const VERSION: &'static str = "version";
```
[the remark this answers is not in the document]
cgorenflo: not sure there is a lot of benefit to it and will lead to unnecessary dependencies. In my opinion the contracts don't need to know that there is an upgrader contract calling them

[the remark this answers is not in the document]
milapsheth: This interface only defines that the contract is upgradable, not that there is an upgrader contract. We can put this interface in the axelar-soroban-std instead, and the upgrader can use it instead of hardcoding function interfaces

contracts/axelar-gateway/src/contract.rs
```diff
@@ -21,6 +21,24 @@ const INSTANCE_TTL_EXTEND_TO: u32 = 60 * LEDGERS_PER_DAY;
...
+#[contractimpl]
+impl UpgradeableInterface for AxelarGateway {
+    type Error = ContractError;
+
+    fn version(env: Env) -> String {
+        String::from_str(&env, CONTRACT_VERSION)
+    }
+
+    fn upgrade(env: Env, new_wasm_hash: BytesN<32>) -> Result<(), ContractError> {
+        Self::owner(&env)?.require_auth();
```
ahramy, line 33:
```suggestion
        Self::owner(&env).require_auth();
```

contracts/upgrader/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+
+[dependencies]
+axelar-soroban-std = { workspace = true }
+soroban-sdk = { workspace = true }
+
+[dev-dependencies]
+soroban-sdk = { workspace = true, features = ["testutils"] }
+
+[features]
+
```
ahramy, line 17: Empty line, is it intended?

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,138 @@
...
+}
+
+fn ensure_new_version_matches_expected(
+    env: &Env,
+    contract_address: &Address,
+    new_version: &String,
+) -> Result<(), ContractError> {
+    let versions_match = current_version(env, contract_address) == *new_version;
+    versions_match.then_ok((), ContractError::VersionMismatch)
+}
```
ahramy, line 59: `ensure_new_version_is_different` and `ensure_new_version_matches_expected` have similar logic but different number of lines: 3 vs 2 lines.

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,408 @@
...
+            // it's safe to map the true return value of the upgrade and migrate functions to (),
+            // because we don't care about it, and in case of failure the contract will panic anyway
+
+            env.invoke_contract::<()>(
+                &contract_address,
+                &Symbol::new(&env, UPGRADE),
+                soroban_sdk::vec![&env, new_wasm_hash.into()],
+            );
+
+            env.invoke_contract::<()>(
```
milapsheth, line 43: import the client so you can avoid using invoke_contract, e.g.
https://github.com/axelarnetwork/axelar-cgp-soroban/blob/527c2fb63b9ad1596d131fdb5baf0b0c5e7e8a95/contracts/example/src/contract.rs#L54

contracts/upgrader/src/contract.rs
cgorenflo: good catch, the `string` variable was a leftover from some debugging

packages/axelar-soroban-std/src/upgrade.rs
```diff
@@ -0,0 +1,7 @@
+use soroban_sdk::{BytesN, Env, String};
+
+pub trait UpgradeableInterface {
```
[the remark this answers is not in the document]
cgorenflo: won't work due to the associate type

contracts/upgrader/src/contract.rs
cgorenflo: as discussed, this has to be generic (not in the generic type sense), because the upgrader doesn't know which concrete error type it has to cast to

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,317 @@
...
+
+        // ensure migration was successful
+        env.as_contract(&contract_address, || {
+            let data: String = env.storage().instance().get(&DataKey::Data).unwrap();
+            assert_eq!(data, expected_data);
+        });
+    }
+
+    #[test]
+    #[should_panic]
```
ahramy, line 205: It would be good to specify the expected panic reason
```suggestion
    #[should_panic(expected = "HostError: Error(Auth, InvalidAction)")]
```

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,317 @@
...
+        UpgraderClient::new(&env, &upgrader_address).upgrade(
+            &contract_address,
+            &expected_version,
+            &hash_after_upgrade,
+            &soroban_sdk::vec![&env, expected_data.to_val()],
+        );
+    }
+
+    #[test]
+    #[should_panic]
```
ahramy, line 244:
```suggestion
    #[should_panic(expected = "HostError: Error(Auth, InvalidAction)")]
```

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,317 @@
...
+    const WASM_AFTER_UPGRADE: &[u8] = include_bytes!("testdata/dummy.wasm");
+
+    #[test]
+    fn upgrade_and_migrate_are_atomic() {
+        let env = Env::default();
+
+        let contract_address = env.register(DummyContract, ());
+        let upgrader_address = env.register(Upgrader, ());
+
+        let owner = set_owner(&env, &contract_address);
```
ahramy, line 158: The owner is being set in a separate function for `DummyContract`. Ideally, the ownership should be established in the `__constructor` function or transferred from the previous owner to the new owner using `transfer_ownership` function, gated with `owner.require_auth()`. Would this approach align with `UpgradeableInterface` or is it beyond the scope of this PR? I'm opting to be cautious.

cgorenflo: I can move setting the owner to the constructor, but I would like to keep the `DummyContract` as simple as I can, because I don't want the upgrader to make assumption about what's available beyond the upgrade, version and migrate functions. The actual upgrade and migration functions need to be tested in each contract themselves, again without making any assumptions about the upgrader

ahramy (approved):

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,312 @@
...
+
+        // add the owner to the set of authenticated addresses
+        env.mock_auths(&[
+            MockAuth {
+                address: &owner,
+                invoke: &upgrade_auth,
+            },
+            MockAuth {
+                address: &owner,
+                invoke: &migrate_auth,
```
milapsheth, line 182: Try using `assert_invoke_auth_err` instead? You can add another pattern for it where it takes a list of callers and invocation calls and creates mock auths for them, and then calls the final method to test the error?

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,312 @@
...
+                address: &owner,
+                invoke: &upgrade_auth,
+            },
+            MockAuth {
+                address: &caller,
+                invoke: &migrate_auth,
+            },
+        ]);
+
+        UpgraderClient::new(&env, &upgrader_address).upgrade(
```
milapsheth, line 276: add a test where the mock auth is missing entirely for each of the invocation?

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,317 @@
...
+    ///     }
+    /// }
+    const WASM_AFTER_UPGRADE: &[u8] = include_bytes!("testdata/dummy.wasm");
+
+    #[test]
+    fn upgrade_and_migrate_are_atomic() {
+        let env = Env::default();
+
+        let contract_address = env.register(DummyContract, ());
+        let upgrader_address = env.register(Upgrader, ());
```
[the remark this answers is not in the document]
cgorenflo: I need it for the upgraded version anyway because I don't have the source code for it around

contracts/upgrader/src/contract.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    #[test]
+    #[should_panic]
+    fn upgrade_fails_if_caller_is_authenticated_but_not_owner() {
+        let env = Env::default();
+
+        let contract_address = env.register(DummyContract, ());
+        let upgrader_address = env.register(Upgrader, ());
+
+        let _owner = set_owner(&env, &contract_address);
```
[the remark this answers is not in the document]
cgorenflo: I need the owner, and contract addresses throughout the code. I find 
```
let owner = Address::generate(&env);
let contract_address = env.register(DummyContract, (&owner,));
let upgrader_address = env.register(Upgrader, ());
```
cleaner than

```
let (owner, contract_address, upgrader_address) = setup(&env)

...

fn setup(env: &Env){
   let owner = Address::generate(env);
   let contract_address = env.register(DummyContract, (&owner,));
   let upgrader_address = env.register(Upgrader, ());
}
```

contracts/upgrader/src/contract.rs
cgorenflo: the auth that I need to mock here is very specific for this scenario: I call function A, but need to authenticate functions B and C that have very specific subsets of the input parameters. Also, the tests are modulating what is authenticated and what is not, so there is no good macro solution, and it wouldn't be reusable anyway

PR #80, opened 2024-11-19
feat: allow exporting soroban contract crates are libraries

milapsheth:
[AXE-6666]

This PR adds support for a `library` feature that allows soroban contracts to be exported as libraries, i.e without entrypoints to prevent issues with entrypoint name collisions.

`soroban contract build` now works without name collision issues.

## Notes

- A `testutils` feature is now also required, since Rust doesn't seem to allow declaring a dependency with `features = ["library"]` but a dev dependency with `features = []`, which I believe is because cargo features need to be additive (although need to dive more into). And the actual contract needs to be available in tests.
- Another consequence of this is that it enforces that private methods don't get included as contract entrypoints, since the trait implementation won't allow it
- The Interface trait can also extend other Interfaces (e.g. Upgradable, etc.)

[AXE-6666]: https://axelarnetwork.atlassian.net/browse/AXE-6666?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/axelar-gas-service/src/lib.rs
+++ b/contracts/axelar-gas-service/src/lib.rs
@@ -1,9 +1,20 @@
 #![no_std]
 
-mod event;
-mod storage_types;
-
-pub mod contract;
 pub mod error;
 
-pub use contract::AxelarGasServiceClient;
+mod interface;
+
+#[cfg(all(target_family = "wasm", feature = "testutils"))]
+compile_error!("'testutils' feature is not supported on 'wasm' target");
+
+cfg_if::cfg_if! {
+    if #[cfg(all(feature = "library", not(feature = "testutils")))] {
+        pub use interface::{AxelarGasServiceClient, AxelarGasServiceInterface};
+    } else {
+        mod event;
+        mod storage_types;
+
+        pub mod contract;
+        pub use contract::{AxelarGasService, AxelarGasServiceClient};
+    }
+}
--- a/contracts/axelar-gateway/src/lib.rs
+++ b/contracts/axelar-gateway/src/lib.rs
@@ -1,23 +1,31 @@
 #![no_std]
 
-mod auth;
-mod event;
-mod storage_types;
-
-pub mod contract;
-pub mod error;
-pub mod executable;
-pub mod types;
-
-#[cfg(all(target_family = "wasm", feature = "testutils"))]
-compile_error!("'testutils' feature is not supported on 'wasm' target");
-
-#[cfg(any(test, feature = "testutils"))]
-pub mod testutils;
-
 // Allows using std (and its macros) in test modules
 #[cfg(test)]
 #[macro_use]
 extern crate std;
 
-pub use contract::{AxelarGateway, AxelarGatewayClient};
+pub mod error;
+pub mod types;
+
+pub mod executable;
+mod interface;
+
+#[cfg(all(target_family = "wasm", feature = "testutils"))]
+compile_error!("'testutils' feature is not supported on 'wasm' target");
+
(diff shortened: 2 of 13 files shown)


contracts/axelar-gas-service/src/lib.rs
```diff
@@ -1,9 +1,20 @@
...
+    if #[cfg(all(feature = "library", not(feature = "testutils")))] {
+        pub use interface::{AxelarGasServiceClient, AxelarGasServiceInterface};
+    } else {
+        mod event;
+        mod storage_types;
+
+        pub mod contract;
+        pub use contract::{AxelarGasService, AxelarGasServiceClient};
+    }
+}
```
cgorenflo, line 20: should this be part of the build.rs instead?

contracts/axelar-gateway/src/lib.rs
```diff
@@ -1,23 +1,31 @@
...
-pub mod contract;
 pub mod error;
-pub mod executable;
 pub mod types;
 
+pub mod executable;
+mod interface;
+
 #[cfg(all(target_family = "wasm", feature = "testutils"))]
 compile_error!("'testutils' feature is not supported on 'wasm' target");
```
cgorenflo, line 15: same question here

contracts/axelar-gas-service/src/lib.rs
cgorenflo: The second part probably needs to be here, so it's about the compile_error macro

contracts/axelar-gas-service/src/lib.rs
```diff
@@ -1,9 +1,20 @@
...
+
+cfg_if::cfg_if! {
+    if #[cfg(all(feature = "library", not(feature = "testutils")))] {
+        pub use interface::{AxelarGasServiceClient, AxelarGasServiceInterface};
+    } else {
+        mod event;
+        mod storage_types;
+
+        pub mod contract;
+        pub use contract::{AxelarGasService, AxelarGasServiceClient};
```
cgorenflo, line 18: are both exports necessary?

contracts/axelar-gateway/src/lib.rs
```diff
@@ -1,23 +1,31 @@
...
+    if #[cfg(all(feature = "library", not(feature = "testutils")))] {
+        pub use interface::{AxelarGatewayClient, AxelarGatewayInterface};
+    } else {
+        mod auth;
+        mod event;
+        mod storage_types;
 
-pub use contract::{AxelarGateway, AxelarGatewayClient};
+        pub mod contract;
+        pub use contract::{AxelarGateway, AxelarGatewayClient};
```
cgorenflo, line 29: are both exports necessary?

contracts/axelar-gas-service/src/lib.rs
milapsheth: i'm thinking of keeping only the latter. But want to do that in a separate PR for all contracts

contracts/axelar-gas-service/src/lib.rs
milapsheth: makes sense for `compile_error`

cgorenflo (approved):

contracts/axelar-gas-service/src/lib.rs
milapsheth: will be investigated separately

Foivos (approved):

hydrobeam (approved):

PR #85, opened 2024-11-21
feat(axelar-gateway): emit event when upgrade is completed

cgorenflo:


contracts/axelar-gateway/src/event.rs
```diff
@@ -51,3 +51,8 @@ pub fn transfer_ownership(env: &Env, previous_owner: Address, new_owner: Address
     );
     env.events().publish(topics, ());
 }
+
+pub fn upgrade_complete(env: &Env, version: &String) {
+    env.events()
+        .publish((Symbol::new(env, "upgrade_completed"),), version.to_val());
```
milapsheth, line 57:
```suggestion
pub fn upgrade(env: &Env, version: &String) {
    env.events()
        .publish((Symbol::new(env, "upgraded"),), (version,).into());
```
There's no upgrade started event, so I think this is simpler (and don't really need a started event since it's atomic unless you think it's better) 

Also, stellar team said to use tuple for data, so more values can be added later without breaking the event

milapsheth (approved):

contracts/axelar-gateway/src/event.rs
cgorenflo: They said to use a `Vec`, didn't they? I think a tuple would have the same effect as a regular value because it has a defined length, so it will break anyone parsing it if we add a new value. With a vector we can always extend it.

PR #87, opened 2024-11-22
feat: add upgrade capabilities to all contracts

cgorenflo:
[AXE-6100]

[AXE-6100]: https://axelarnetwork.atlassian.net/browse/AXE-6100?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Overview:
- `OwnershipInterface`, `UpgradeableInterface` and `MigratableInterface` are defined in packages/axelar-soroban-std/src/shared_interfaces.rs and implemented by all contracts
- migration is customizable for each contract and can change from version to version
- migration is only allowed after an upgrade and only callable once
- both upgrade and migrate require authentication of the owner account
- once migration is complete, an event is emitted that contains the new version of the contract


contracts/axelar-gateway/src/storage_types.rs
```diff
@@ -19,11 +19,8 @@ pub enum MessageApprovalValue {
 #[derive(Clone, Debug)]
 pub enum DataKey {
     /// Gateway
-    Owner,
```
milapsheth, line 22: as part of doing code reuse, we wanted to understand the impact of using different DataKey types for the same contract. If XDR encoding is used, then unique enum variant names should be enough to avoid collisions. But if something else is done internally with the key, then we'd need to investigate

packages/axelar-soroban-std/src/upgrade.rs
```diff
@@ -1,10 +1,77 @@
...
     fn version(env: &Env) -> String;
 
+    /// The owner that is allowed to upgrade the contract.
+    fn owner(env: &Env) -> Address {
+        env.storage()
+            .instance()
+            .get(&DataKey::Owner)
+            .expect("owner not found")
+    }
+
```
milapsheth, line 16: add a default trait fn `set_owner` to set the owner. The contract constructors can then use `Self::set_owner` to set the owner instead of accessing the `upgrade::DataKey` directly. Default trait functions aren't included in contractimpl entrypoints last time I checked. It works here, although `owner` method above can't be called to query the owner on-chain

contracts/axelar-gas-service/src/contract.rs
```diff
@@ -1,23 +1,44 @@
...
+        env.storage()
+            .instance()
+            .set(&upgrade::DataKey::Owner, &owner);
         env.storage()
             .instance()
             .set(&DataKey::GasCollector, &gas_collector);
     }
+
+    pub fn migrate(env: &Env, migration_data: ()) -> Result<(), ContractError> {
+        standardized_migrate::<Self>(env, || Self::run_migration(env, migration_data))
```
milapsheth, line 26:
```suggestion
        axelar_soroban_std::migrate::<Self>(env, || Self::run_migration(env, migration_data))
```
perhaps?

packages/axelar-soroban-std/src/shared_interfaces.rs
```diff
@@ -0,0 +1,398 @@
...
+        assert!(client.migration_data().is_none());
+
+        assert_invoke_auth_ok!(owner, client.try_migrate(&()));
+
+        assert_eq!(
+            client.migration_data(),
+            Some(String::from_str(&env, "migrated"))
+        );
+
+        let event = env
```
milapsheth, line 349: This is a nice pattern over the current usage of `assert_last_emitted_event` to test events. Can you create a helper that converts the event emitted at some index to the arbitrary event type `T`, could add the last_* version of this function as well, since it's commonly used. I like specifying the index better than filtering over all events, so we know if the order changes as well

packages/axelar-soroban-std/src/shared_interfaces.rs
```diff
@@ -0,0 +1,398 @@
...
+        env,
+        UpgradedEvent {
+            version: T::version(env),
+        },
+    );
+
+    Ok(())
+}
+
+fn emit_event_upgraded(env: &Env, event: UpgradedEvent) {
```
milapsheth, line 81: Maybe create an `Event` trait that has `topic`, `data` methods, and a default `emit` method that uses the other methods? So usage becomes something like:
```
UpgradedEvent {
    version: T::version(env),
}.emit(env);
```

packages/axelar-soroban-std/src/shared_interfaces.rs
```diff
@@ -0,0 +1,398 @@
...
+        assert_eq!(
+            client.migration_data(),
+            Some(String::from_str(&env, "migrated"))
+        );
+
+        let event = env
+            .events()
+            .all()
+            .iter()
+            .find_map(|event| UpgradedEvent::try_from_val(&env, &event).ok());
```
milapsheth, line 353: Instead of having to implement try_from_val for each event, create a version of `assert_last_emitted_event` helper, that takes a type `T` implementing the `Event` trait, and checks that the emitted event matches the expected one at the specific index. This also ensure ordering doesn't change

packages/axelar-soroban-std/src/testdata/contract.rs
```diff
@@ -0,0 +1,62 @@
...
+        shared_interfaces::owner(env)
+    }
+}
+
+#[contractimpl]
+impl UpgradeableInterface for Contract {
+    fn version(env: &Env) -> String {
+        String::from_str(env, "0.1.0")
+    }
+    fn upgrade(env: &Env, new_wasm_hash: BytesN<32>) {
```
milapsheth, line 48:
```suggestion

    fn upgrade(env: &Env, new_wasm_hash: BytesN<32>) {
```

packages/axelar-soroban-std/src/testdata/contract.rs
```diff
@@ -0,0 +1,62 @@
...
+            .instance()
+            .set(&DataKey::Data, &String::from_str(env, "migrated"));
+    }
+}
+
+#[contractimpl]
+impl MigratableInterface for Contract {
+    type MigrationData = ();
+    type Error = ContractError;
+    fn migrate(env: &Env, migration_data: ()) -> Result<(), ContractError> {
```
milapsheth, line 30:
```suggestion

    fn migrate(env: &Env, migration_data: ()) -> Result<(), ContractError> {
```

contracts/upgrader/tests/utils/dummy_contract.rs
```diff
@@ -43,32 +47,43 @@ pub enum ContractError {
 
 // Dummy contract logic after upgrade is available as testdata/dummy.wasm
```
milapsheth, line 48: define this in dummy_contract_0_2.rs maybe?

contracts/axelar-operators/src/contract.rs
```diff
@@ -103,4 +93,37 @@ impl AxelarOperators {
 
         Ok(res)
     }
+
+    // Modify this function to add migration logic
+    const fn run_migration(_env: &Env, _migration_data: ()) {}
```
milapsheth, line 98: Define this outside of a `contractimpl` impl block

milapsheth: once we add the interface for this contract, it should help

packages/axelar-soroban-std/src/shared_interfaces.rs
```diff
@@ -0,0 +1,398 @@
...
+}
+
+pub trait MigratableInterface: UpgradeableInterface {
+    /// Data needed during the migration. Each contract can define its own data type.
+    type MigrationData: FromVal<Env, Val>;
+    /// Error type returned if the migration fails.
+    type Error: Into<soroban_sdk::Error>;
+
+    /// Migrates contract state after upgrading the contract code.
+    fn migrate(env: &Env, migration_data: Self::MigrationData) -> Result<(), Self::Error>;
```
milapsheth, line 28: I'm ok with not returning a result since the caller doesn't do anything meaningful with the error. We can just use `panic_with_error` internally instead to keep this simpler

packages/axelar-soroban-std/src/shared_interfaces.rs
milapsheth: Actually, might be better to define the event to include a reference to Env
```
UpgradedEvent {
    env,
    version: T::version(env),
}.emit();
```
since constructing the topic/data might require `env` (e.g. hashing a payload to emit in the topic). Alternatively, passing `env` to `topic()` and `data()` from `emit`

packages/axelar-soroban-std/src/shared_interfaces.rs
cgorenflo: I prefer to keep it as is. Migration can go wrong for a variety of reasons and I don't want to take the choice away from contracts to return a specific error, just because it's likely they will get called from the upgrader contract.
I'm not a fan of how much soroban makes use of panics

milapsheth: the upgrader will panic if migrate errors anyways, so doesn't make too much difference until soroban adds error stacking ability. but i'm fine to keep it as-is

milapsheth (approved):

packages/axelar-soroban-std/src/shared_interfaces.rs
cgorenflo: follow-up PR

packages/axelar-soroban-std/src/shared_interfaces.rs
cgorenflo: follow-up PR

cgorenflo:
Overview:
- OwnershipInterface, UpgradeableInterface and MigratableInterface are defined in packages/axelar-soroban-std/src/shared_interfaces.rs and implemented by all contracts
- migration is customizable for each contract and can change from version to version
- migration is only allowed after an upgrade and only callable once
- both upgrade and migrate require authentication of the owner account
- once migration is complete, an event is emitted that contains the new version of the contract

PR #89, opened 2024-11-25
fix(axelar-gas-service)!: fix axlear gas service add_gas

ahramy:
- Add `spender` in add_gas
- Use `transfer` instead of `transfer_from`

### DIFF
--- a/contracts/axelar-gas-service/src/interface.rs
+++ b/contracts/axelar-gas-service/src/interface.rs
@@ -10,12 +10,9 @@ pub trait AxelarGasServiceInterface {
     ///
     /// This function is called on the source chain before calling the gateway to send a message.
     ///
-    /// # Notes
-    /// - The `sender` is distinct from the `spender`. The `sender` initiates the action
-    ///   requiring gas payment but does not directly pay for the gas. Instead, the
-    ///   `spender` is responsible for authorizing and funding the payment.
-    /// - The `spender` can also serve as the `refund_address`, receiving any unused gas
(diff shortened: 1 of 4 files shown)


contracts/axelar-gas-service/src/interface.rs
```diff
@@ -28,12 +28,18 @@ pub trait AxelarGasServiceInterface {
     ) -> Result<(), ContractError>;
 
     /// Add additional gas payment after initiating a cross-chain message.
+    /// # Notes
+    /// - The `sender` is distinct from the `spender`. The `sender` initiates the action
+    ///   requiring gas payment but does not directly pay for the gas. Instead, the
+    ///   `spender` is responsible for authorizing and funding the payment.
+    /// - The `spender` can also serve as the `refund_address`, receiving any unused gas
+    ///   or reimbursement as applicable.
```
milapsheth, line 36:
```suggestion
    /// `sender` is the address that sent the cross-chain message via the `axelar_gateway`.
    /// The `spender` pays the gas, but might not be the same as the `sender`, e.g. `sender` is a contract, but `spender` can be the user signing the tx.
```

milapsheth: update `pay_gas` doc accordingly

milapsheth (approved):

contracts/axelar-gas-service/src/interface.rs
ahramy: updated

PR #90, opened 2024-11-26
feat(interchain-token-service): add message routing

AttissNgo:
[AXE-6067](https://axelarnetwork.atlassian.net/browse/AXE-6067)

[AXE-6067]: https://axelarnetwork.atlassian.net/browse/AXE-6067?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/src/contract.rs
```diff
@@ -238,3 +328,381 @@ impl AxelarExecutableInterface for InterchainTokenService {
         event::executed(&env, source_chain, message_id, source_address, payload);
     }
 }
+
+#[cfg(test)]
```
ahramy, line 332: should we move tests to `test.rs` rather than having in `contract.rs` ?

contracts/interchain-token-service/src/contract.rs
```diff
@@ -238,3 +328,381 @@ impl AxelarExecutableInterface for InterchainTokenService {
...
+
+        let chain = String::from_str(&env, HUB_CHAIN);
+        client.set_trusted_address(&chain, &client.its_hub_routing_identifier());
+
+        let chain = client.its_hub_chain_name();
+        let addr = String::from_str(&env, HUB_ADDRESS);
+        client.set_trusted_address(&chain, &addr);
+    }
+
+    fn bytes_from_hex(env: &Env, hex_string: &str) -> Bytes {
```
ahramy, line 371: This `bytes_from_hex` function is defined twice, the other one is in [abi.rs](https://github.com/axelarnetwork/axelar-cgp-soroban/blob/ce8d3af09bff15b85d94b5a2502806411a62374e/contracts/interchain-token-service/src/abi.rs#L257).

Could we move this to test utils to avoid defining the same function twice?

contracts/interchain-token-service/src/contract.rs
```diff
@@ -238,3 +328,381 @@ impl AxelarExecutableInterface for InterchainTokenService {
...
+
+    #[test]
+    fn get_call_params_hub_message() {
+        let (env, client) = setup_env();
+        register_chains(&env, &client);
+
+        let msg = Message::InterchainTransfer(InterchainTransfer {
+            token_id: BytesN::from_array(&env, &[255u8; 32]),
+            source_address: bytes_from_hex(&env, "4F4495243837681061C4743b74B3eEdf548D56A5"),
+            destination_address: bytes_from_hex(&env, "4F4495243837681061C4743b74B3eEdf548D56A5"),
```
ahramy, line 384: `source_address` and `destination_address` are the same. Is this intentional? If not, it might be better to use different addresses.

contracts/interchain-token-service/src/contract.rs
```diff
@@ -152,63 +258,47 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
     ) {
         todo!()
     }
 
     fn deploy_remote_interchain_token(
-        env: &Env,
-        caller: Address,
-        destination_chain: String,
+        _env: &Env,
+        _caller: Address,
```
milapsheth, line 274: `caller` becomes the `source_address` in the deploy Message

contracts/interchain-token-service/src/interface.rs
```diff
@@ -19,32 +19,31 @@ pub trait InterchainTokenServiceInterface: AxelarExecutableInterface {
...
-        _token_id: String,
-        _source_address: Bytes,
+        _token_id: BytesN<32>,
         _destination_chain: String,
-        _destination_address: Bytes,
-        _amount: i128,
-        _metadata: Bytes,
+        _name: String,
+        _symbol: String,
+        _decimals: u32,
```
milapsheth, line 26: take in a TokenMetadata object instead

contracts/interchain-token-service/src/contract.rs
```diff
@@ -152,63 +258,47 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
-            caller,
-            destination_chain,
-            String::from_val(env, &destination_address.to_val()),
-            payload,
-            gas_token,
-        );
+        // EXAMPLE implementation only to compile
+        let message = Message::InterchainTransfer(InterchainTransfer {
+            token_id,
+            source_address: Bytes::from_slice(env, &[0]),
```
milapsheth, line 295: caller is the sender of the transfer / source address

contracts/interchain-token-service/src/contract.rs
```diff
@@ -329,380 +183,146 @@ impl AxelarExecutableInterface for InterchainTokenService {
...
-            assert_eq!(source_chain_name, String::from_str(&env, P2P_CHAIN));
-            assert_eq!(message, msg);
+fn get_execute_params(
+    env: &Env,
+    source_chain: String,
+    payload: &Bytes,
+) -> Result<(String, Message), ContractError> {
+    let message_type =
+        get_message_type(&payload.to_alloc_vec()).map_err(|_| ContractError::InvalidPayload)?;
+
```
milapsheth, line 265: use `ensure!(message_type == ..)` so it doesn't need to be nested under a `match ...`

contracts/interchain-token-service/src/contract.rs
```diff
@@ -329,380 +183,146 @@ impl AxelarExecutableInterface for InterchainTokenService {
...
-            let (source_chain_name, message) = assert_ok!(env.as_contract(&client.address, || {
-                InterchainTokenService::get_execute_params(
-                    &env,
-                    String::from_str(&env, P2P_CHAIN),
-                    encoded_msg,
-                )
-            }));
-            assert_eq!(source_chain_name, String::from_str(&env, P2P_CHAIN));
-            assert_eq!(message, msg);
+fn get_execute_params(
```
milapsheth, line 258: same

contracts/interchain-token-service/src/contract.rs
```diff
@@ -329,380 +183,146 @@ impl AxelarExecutableInterface for InterchainTokenService {
...
+    destination_chain: String,
+    message: Message,
+) -> Result<Bytes, ContractError> {
+    // Note: ITS Hub chain as the actual destination chain for the messsage isn't supported
+    ensure!(
+        destination_chain != InterchainTokenService::its_hub_chain_name(env),
+        ContractError::UntrustedChain
+    );
+
+    match InterchainTokenService::trusted_address(env, destination_chain.clone()) {
```
milapsheth, line 310:
```suggestion
    let Some(destination_address) = InterchainTokenService::trusted_address(env, destination_chain.clone()) else { return Err(...) };
```
use let else instead of match

contracts/interchain-token-service/src/contract.rs
```diff
@@ -183,146 +183,145 @@ impl AxelarExecutableInterface for InterchainTokenService {
...
+            destination_address == Self::its_hub_routing_identifier(env),
+            ContractError::UntrustedChain
+        );
+
+        let payload = HubMessage::SendToHub {
+            destination_chain,
+            message,
         }
-        _ => Err(ContractError::UntrustedChain),
+        .abi_encode(env)
```
milapsheth, line 322: make `abi_encode` use a `ContractError` so you don't have to map it

contracts/interchain-token-service/tests/interchain_transfer.rs
```diff
@@ -0,0 +1,130 @@
...
+        &sender,
+        &token_id,
+        &destination_chain,
+        &destination_address,
+        &amount,
+        &data,
+        &gas_token,
+    );
+
+    assert_last_emitted_event(
```
milapsheth, line 56: this should be switched to using the new style event test with goldie that Chris introduced in a follow up PR

milapsheth (approved):

contracts/interchain-token-service/src/abi.rs
```diff
@@ -1,24 +1,15 @@
 use alloy_primitives::{FixedBytes, Uint, U256};
 use alloy_sol_types::{sol, SolValue};
 use axelar_soroban_std::ensure;
-use soroban_sdk::{contracterror, Bytes, BytesN, Env, String};
+use soroban_sdk::{Bytes, BytesN, Env, String};
 
+use crate::abi::alloc::{string::String as StdString, vec};
+use crate::error::ContractError;
 use crate::types::{self, HubMessage, Message};
 extern crate alloc;
```
milapsheth, line 9: add a comment for why this is needed

contracts/interchain-token-service/src/contract.rs
```diff
@@ -186,52 +113,61 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
+        _destination_chain: String,
+        _gas_token: Token,
+    ) -> Result<BytesN<32>, ContractError> {
+        // TODO: implementation
+        // deploy salt
+        // minter approval
+        // registered token address - get token metadata
+        // abi encode with MessageType.DeployInterchainToken
+        // token id
+        // pay_gas_and_call_contract
```
milapsheth, line 142: these comments should be in the task or some other place instead of the code itself

contracts/interchain-token-service/src/contract.rs
```diff
@@ -186,52 +113,61 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
-            env.clone(),
-            caller,
-            destination_chain,
-            String::from_val(env, &destination_address.to_val()),
-            payload,
-            gas_token,
-        );
+        let message = Message::InterchainTransfer(InterchainTransfer {
+            token_id,
+            source_address: caller.clone().to_xdr(env),
```
milapsheth, line 162: How does Address when encoded in XDR look like? is it just the raw bytes, or is there other metadata. We would like to encode the raw address bytes, it's easier to debug and reason about then some custom data format, similar to EVM accounts

contracts/interchain-token-service/src/lib.rs
```diff
@@ -6,9 +6,12 @@ pub mod error;
 mod event;
 mod interface;
 mod storage_types;
-mod types;
+pub mod types;
 
 #[cfg(test)]
 extern crate std;
 
+#[cfg(any(test, feature = "testutils"))]
```
milapsheth, line 14: add library export feature in a follow up PR

contracts/interchain-token-service/src/testutils.rs
```diff
@@ -0,0 +1,85 @@
...
+
+    let chain = String::from_str(&env, HUB_CHAIN);
+    client.set_trusted_address(&chain, &client.its_hub_routing_identifier());
+
+    let chain = client.its_hub_chain_name();
+    let addr = String::from_str(&env, HUB_ADDRESS);
+    client.set_trusted_address(&chain, &addr);
+}
+
+pub fn bytes_from_hex(env: &Env, hex_string: &str) -> Bytes {
```
milapsheth, line 82: Here's a cleaner way to implement this.
1. Create a `BytesExt` trait in the axelar-soroban-std crate, which defines a `from_hex` method
2. Implement this trait for `Bytes` in the same crate
3. Import the trait, and use it via `Bytes::from_hex(env, ...)`
Similar example here: https://github.com/axelarnetwork/axelar-amplifier/blob/main/packages/axelar-wasm-std/src/vec.rs

you can try this in a small followup PR

contracts/interchain-token-service/tests/interchain_transfer.rs
```diff
@@ -0,0 +1,130 @@
...
+    let data_hash = get_approve_hash(&env, messages.clone());
+    let proof = generate_proof(&env, data_hash, signers);
+
+    gateway_client.approve_messages(&messages, &proof);
+
+    client.execute(&source_chain, &message_id, &source_address, &payload);
+
+    assert_emitted_event(
+        &env,
+        -2,
```
milapsheth, line 118: why isn't this last event? Create a ticket to move the executed event to be emitted within `validate_message` so it's emitted first for consistency with EVM and for code reuse. you can try tackling it if you want

milapsheth (approved):

PR #92, opened 2024-11-28
feat(axelar-soroban-std): allow typed matching of emitted events

cgorenflo:
AXE-6758


contracts/interchain-token-service/src/testdata/interchain_transfer_encode_decode.golden
```diff
@@ -3,4 +3,4 @@
...
   "0000000000000000000000000000000000000000000000000000000000000004000000000000000000000000000000000000000000000000000000000000006000000000000000000000000000000000000000000000000000000000000000a00000000000000000000000000000000000000000000000000000000000000005636861696e00000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000001800000000000000000000000000000000000000000000000000000000000000000ffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff00000000000000000000000000000000000000000000000000000000000000c00000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000007fffffffffffffffffffffffffffffff000000000000000000000000000000000000000000000000000000000000014000000000000000000000000000000000000000000000000000000000000000144f4495243837681061c4743b74b3eedf548d56a500000000000000000000000000000000000000000000000000000000000000000000000000000000000000144f4495243837681061c4743b74b3eedf548d56a50000000000000000000000000000000000000000000000000000000000000000000000000000000000000002abcd000000000000000000000000000000000000000000000000000000000000"
-]
+]
```
cgorenflo, line 6: I don't know why this had a newline at the end to begin with, goldie never generates one. So after regenerating all goldie files it's gone again

packages/axelar-soroban-std/src/shared_interfaces.rs
```diff
@@ -105,35 +102,23 @@ fn complete_migration(env: &Env) {
...
-    pub fn topic() -> impl Topics {
+impl Event for UpgradedEvent {
+    fn topics(&self) -> impl Topics + Debug {
         (symbol_short!("upgraded"),)
     }
 
-    pub fn data(&self) -> impl IntoVal<Env, Val> {
-        (self.version.to_val(),)
+    fn data(&self) -> impl IntoVal<Env, Val> + Debug {
+        (self.version.clone(),)
```
milapsheth, line 116: not efficient to have to create a copy of the field to create the topic and data. Should we consider combining the two into something like `fn topic_and_data(self) -> (impl Topics + Debug, impl IntoVal<Env, Val> + Debug)` so it can take `self` instead of a reference

packages/axelar-soroban-std/src/shared_interfaces.rs
```diff
@@ -105,35 +102,23 @@ fn complete_migration(env: &Env) {
...
-    ) -> Result<Self, Self::Error> {
-        ensure!(topics.eq(&Self::topic().into_val(env)), ConversionError);
-
-        let v: Vec<Val> = Vec::try_from_val(env, data)?;
-        String::try_from_val(env, &v.first().ok_or(ConversionError)?)
-            .map(|version| Self { version })
-    }
-}
+#[cfg(any(test, feature = "testutils"))]
+impl_event_testutils!(UpgradedEvent, (soroban_sdk::Symbol), (String));
```
milapsheth, line 121:
```suggestion
impl_event_testutils!(UpgradedEvent, (Symbol), (String));
```
import Symbol

packages/axelar-soroban-std/src/events.rs
```diff
@@ -0,0 +1,155 @@
...
+mod testutils {
+    use crate::events::Event;
+    use soroban_sdk::testutils::Events;
+    use soroban_sdk::{Env, Val};
+
+    pub trait EventTestutils: Event {
+        fn matches(
+            &self,
+            env: &Env,
+            event: &(soroban_sdk::Address, soroban_sdk::Vec<Val>, Val),
```
milapsheth, line 27:
```suggestion
            event: &(Address, Vec<Val>, Val),
```
don't need to use qualified naming for sdk types

packages/axelar-soroban-std/src/events.rs
```diff
@@ -0,0 +1,155 @@
...
+                            .expect("given topic value does not match the expected type")));
+
+                        i += 1;
+                    )*
+
+                    let data = soroban_sdk::Vec::<soroban_sdk::Val>::try_from_val(env, data).expect("data should be defined as a vector-compatible type");
+
+                    let mut data_output = std::vec![];
+
+
```
milapsheth, line 85:
```suggestion
```

packages/axelar-soroban-std/src/events.rs
```diff
@@ -0,0 +1,155 @@
...
+                    let mut i = 0;
+                    $(
+                        let data_entry = data.get(i).expect("the number of data entries does not match this function's definition");
+                        data_output.push(std::format!("{:?}", <$data_type>::try_from_val(env, &data_entry)
+                            .expect("given data value does not match the expected type")));
+
+                        i += 1;
+                    )*
+
+                    std::format!("contract: {:?}, topics: ({}), data: ({})", contract_id, topics_output.join(", "), data_output.join(", "))
```
milapsheth, line 95: print on different lines
```suggestion
                    std::format!("contract: {:?}\n\ntopics: ({})\n\ndata: ({})", contract_id, topics_output.join(", "), data_output.join(", "))
```

packages/axelar-soroban-std/src/events.rs
```diff
@@ -0,0 +1,155 @@
...
+                        i += 1;
+                    )*
+
+                    let data = soroban_sdk::Vec::<soroban_sdk::Val>::try_from_val(env, data).expect("data should be defined as a vector-compatible type");
+
+                    let mut data_output = std::vec![];
+
+
+                    let mut i = 0;
+                    $(
```
milapsheth, line 87: use a for loop instead of this?

cgorenflo: does not work, this is pattern matched on the number of arguments the macro receives

packages/axelar-soroban-std/src/shared_interfaces.rs
cgorenflo: that clone slipped in from some experimentation. The data has to be converted into a val anyway, so there isn't a good reason to muddle up the interface

packages/axelar-soroban-std/src/events.rs
cgorenflo: it's not a loop during runtime, but the compiler inserts this block n times

PR #93, opened 2024-12-02
test(axelar-gateway): rework auth unit tests to integration tests

hydrobeam:
closes [AXE-6475](https://axelarnetwork.atlassian.net/browse/AXE-6475)

# Summary
Rework the auth unit tests to depend only on the valid contract entry points instead of calling private `auth::` functions (moving them to `tests/auth.rs`). None of the expected variants were changed, since they're all reachable via the entry points exposed in `axelar_gateway/contract.rs`. 

The commit history shows each test that was removed/replaced. 

A test was removed if its scope was already being addressed in `test.rs` (like `auth::test::rotate_signers()`) and replaced with the appropriate changes otherwise. The replacements were mostly straightforward to handle and didn't require that much intervention (beyond adding a few lines)  + changing the way we expect errors. 

For the validate_proof() tests, the ways to access `auth::validate_proof` are via `rotate_signers()` or `approve_messages()`, so those were used as replacements. 

All tests except `multi_rotate_signers` were handled. This one was left because it explicitly calls `validate_proof` several times and replacing it with `approve_messages` just to use a public entry point feels like it would obfuscate the point of the test and make it harder to understand.

[AXE-6475]: https://axelarnetwork.atlassian.net/browse/AXE-6475?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+
+    let (message, _) = generate_test_message(&env);
+    let messages = vec![&env, message.clone()];
+
+    let msg_hash: BytesN<32> = BytesN::random(&env);
+    let proof = generate_proof(&env, msg_hash, signers);
+
+    // should panic, proof is for different message hash
+    // NOTE: panic occurs in std function, cannot handle explicitly
+    client.approve_messages(&messages, &proof);
```
milapsheth, line 538: instead of panicking, maybe test that the returned error when formatted contains the expected string (maybe extend `assert_contract_err` macro to support a rule that takes a string literal). Also, is it possible to check the typed error instead?
https://docs.rs/soroban-sdk/latest/soroban_sdk/testutils/ed25519/enum.Error.html
https://docs.rs/soroban-env-common/latest/soroban_env_common/struct.Error.html

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
     assert_invoke_auth_err!(not_owner, client.try_upgrade(&new_wasm_hash));
     assert_invoke_auth_err!(client.operator(), client.try_upgrade(&new_wasm_hash));
 }
+
+#[test]
+#[should_panic]
```
milapsheth, line 497: add expected panic message

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+            signer: signer.signer,
+            signature: ProofSignature::Unsigned,
+        });
+    }
+    proof.signers = new_signers;
+
+    let new_signers = generate_signers_set(&env, 5, signers.domain_separator.clone());
+
+    assert_contract_err!(
+        client.try_rotate_signers(&new_signers.signers, &proof, &true),
```
milapsheth, line 561: test with approve_messages to keep the test simpler

contracts/axelar-gateway/src/contract.rs
```diff
@@ -76,9 +76,7 @@ impl AxelarGateway {
             minimum_rotation_delay,
             previous_signers_retention,
             initial_signers,
-        )?;
-
-        Ok(())
+        )
```
milapsheth, line 79: you can add validate_proof as an entry point to the contract to make it easier to test authentication, might be helpful for on-chain debugging as well

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+
+    // Modify signatures to make them invalid
+    let mut new_signers = Vec::new(&env);
+    for signer in proof.signers.iter() {
+        new_signers.push_back(ProofSigner {
+            signer: signer.signer,
+            signature: ProofSignature::Unsigned,
+        });
+    }
+    proof.signers = new_signers;
```
milapsheth, line 556:
```suggestion
    proof.signers.iter().for_each(|mut signer| signer.signature = ProofSignature::Unsigned);
```
maybe something like this instead?

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+fn fail_validate_proof_threshold_not_met() {
+    let (env, signers, client) = setup_env(randint(0, 10), randint(1, 10));
+
+    let mut total_weight = 0u128;
+
+    let (message, _) = generate_test_message(&env);
+    let messages = vec![&env, message.clone()];
+    let msg_hash = get_approve_hash(&env, messages.clone());
+
+    // let msg_hash: BytesN<32> = BytesN::random(&env);
```
milapsheth, line 576: not needed

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+    let mut proof = generate_proof(&env, msg_hash.clone(), signers);
+
+    // Modify signatures to make them invalid
```
milapsheth, line 579: not needed

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+        total_weight += signer.weight;
+
+        if total_weight < proof.threshold {
+            new_signers.push_back(ProofSigner { signer, signature });
+        } else {
+            new_signers.push_back(ProofSigner {
+                signer,
+                signature: ProofSignature::Unsigned,
+            });
+        }
```
milapsheth, line 591:
```suggestion
        let signature = signature if total_weight < proof.threshold else ProofSignature::Unsigned;
        new_signers.push_back(ProofSigner { signer, signature });
```
maybe

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+
+    let new_proof = generate_proof(&env, msg_hash.clone(), new_signers.clone());
+
+    proof.signers = new_proof.signers;
+
+    // validate_proof should panic, signatures do not match signers
+    assert_contract_err!(
+        client.try_rotate_signers(&new_signers.signers, &proof, &true),
+        ContractError::InvalidSignersHash
+    );
```
milapsheth, line 619:
```suggestion
    let invalid_proof = generate_proof(&env, msg_hash.clone(), new_signers.clone());

    assert_contract_err!(
        client.try_rotate_signers(&new_signers.signers, &invalid_proof, &true),
        ContractError::InvalidSignersHash
    );
```
avoid mutation, and use descriptive names so comments aren't needed

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+        signers.signers.signers.set(last_index, last_signer);
+    }
+
+    let msg_hash: BytesN<32> = BytesN::random(&env);
+    let proof = generate_proof(&env, msg_hash.clone(), signers.clone());
+
+    let (message, _) = generate_test_message(&env);
+    let messages = vec![&env, message];
+
+    // should panic, as modified signer wouldn't match the epoch
```
milapsheth, line 640: not needed

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+fn fail_validate_proof_threshold_overflow() {
+    let (env, mut signers, client) = setup_env(randint(0, 10), randint(1, 10));
+
+    let last_index = signers.signers.signers.len() - 1;
+
+    // get last signer and modify its weight to max u128 - 1
+    if let Some(mut last_signer) = signers.signers.signers.get(last_index) {
+        last_signer.weight = u128::MAX - 1;
+        signers.signers.signers.set(last_index, last_signer);
+    }
```
milapsheth, line 632:
```suggestion
    let ProofSigner { signer, weight } = signers.signers.signers.pop_back_unchecked() {
    signers.signers.signers.push_back(ProofSigner { signer, weight: u128::MAX });
```
use helper methods to make it cleaner. use u128::MAX itself to test overflow

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+#[test]
+fn rotate_signers_fail_zero_weight() {
+    let (env, signers, client) = setup_env(1, randint(1, 10));
+
+    let mut new_signers = generate_signers_set(&env, randint(1, 10), BytesN::random(&env));
+
+    let last_index = new_signers.signers.signers.len() - 1;
+
+    // get last signer and modify its weight to zero
+    if let Some(mut last_signer) = new_signers.signers.signers.get(last_index) {
```
milapsheth, line 676: same

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
...
+        .signers
+        .signers
+        .iter()
+        .map(|WeightedSigner { weight, .. }| weight)
+        .reduce(|acc, weight| acc + weight)
+        .expect("Empty signers");
+
+    let new_threshold = total_weight + 1;
+
+    // set the threshold to zero
```
milapsheth, line 748: remove comment. also, it's incorrect

contracts/axelar-gateway/tests/test.rs
```diff
@@ -490,3 +492,346 @@ fn upgrade_unauthorized() {
     assert_invoke_auth_err!(not_owner, client.try_upgrade(&new_wasm_hash));
     assert_invoke_auth_err!(client.operator(), client.try_upgrade(&new_wasm_hash));
 }
+
+#[test]
+#[should_panic]
+/// TODO: figure out how to detect error in constructor failure
+fn fail_initialization_with_empty_signer_set() {
+    let env = Env::default();
+    env.mock_all_auths();
```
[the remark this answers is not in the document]
hydrobeam: nope, removed

contracts/axelar-gateway/tests/test.rs
hydrobeam: there isn't an `iter_mut` exposed so ultimately new `ProofSigner`s have to be created. I tried using `map` to do this functionally but I couldn't `collect()` into a `sorban_sdk::Vec` so I think the way it is acceptable for now

contracts/axelar-gateway/tests/test.rs
hydrobeam: applied a similar change

contracts/axelar-gateway/tests/test.rs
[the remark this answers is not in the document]
hydrobeam: went with just `gateway.rs` and `auth.rs` for now

contracts/axelar-gateway/src/interface.rs
```diff
@@ -40,4 +40,10 @@ pub trait AxelarGatewayInterface: AxelarGatewayMessagingInterface + UpgradableIn
 
     /// Returns the signers hash by epoch.
     fn signers_hash_by_epoch(env: &Env, epoch: u64) -> Result<BytesN<32>, ContractError>;
+
+    fn validate_proof(
```
milapsheth, line 44: Add docstring

contracts/axelar-gateway/src/testutils.rs
```diff
@@ -53,6 +53,17 @@ pub fn setup_gateway<'a>(
     (signer_set, client)
 }
 
+pub fn setup_env<'a>(
```
milapsheth, line 56: This isn't a util being exported to other crates. Define this in `tests/util.rs`instead

contracts/axelar-gateway/src/testutils.rs
```diff
@@ -53,6 +53,17 @@ pub fn setup_gateway<'a>(
...
+    previous_signers_retention: u32,
+    num_signers: u32,
+) -> (Env, TestSignerSet, AxelarGatewayClient<'a>) {
+    let env = Env::default();
+    env.mock_all_auths();
```
milapsheth, line 61: Mocking auth here might affect tests that check the auth. Double check the use

contracts/axelar-gateway/tests/auth.rs
```diff
@@ -0,0 +1,359 @@
...
+    let owner = Address::generate(&env);
+    let operator = Address::generate(&env);
+
+    let empty_signer_set = Vec::<WeightedSigners>::new(&env);
+    let domain_separator: BytesN<32> = BytesN::random(&env);
+    let previous_signers_retention = randint(0, 10) as u64;
+    let minimum_rotation_delay: u64 = 0;
+    let initial_signers = empty_signer_set;
+
+    // should panic because of empty signer set
```
milapsheth, line 24: Remove all unnecessary comments in tests

milapsheth (approved):

contracts/axelar-gateway/tests/test.rs
hydrobeam: done. the test actually would have returned info on the error but i was passing in a i32 where a u64 was needed (defaulted to that) so that's why it was failing.

contracts/axelar-gateway/src/testutils.rs
hydrobeam: set it up as `utils/mod.rs` to prevent it from being interpreted as a test file (https://doc.rust-lang.org/rust-by-example/testing/integration_testing.html)

> Each Rust source file in the tests directory is compiled as a separate crate. In order to share some code between integration tests we can make a module with public functions, importing and using it within tests.

contracts/axelar-gateway/src/testutils.rs
hydrobeam: there are tests that depend on auth being mocked (although it's not necessary for all of them), so I think it's fine to keep

milapsheth: let's mock it explicitly for the tests that need it. you can use `assert_invoke_auth_ok` to avoid having to mock auth completely instead for some of the tests that need auth as well. This can make the test more precise and avoids slipping in an auth bug due to mocking

contracts/axelar-gateway/src/interface.rs
```diff
@@ -41,6 +41,7 @@ pub trait AxelarGatewayInterface: AxelarGatewayMessagingInterface + UpgradableIn
     /// Returns the signers hash by epoch.
     fn signers_hash_by_epoch(env: &Env, epoch: u64) -> Result<BytesN<32>, ContractError>;
 
+    /// Returns a boolean indicating if provided signers are the latest ones
```
milapsheth, line 44:
```suggestion
    /// Validate the `proof` for `data_hash` created by the signers. Returns a boolean indicating if the proof was created by the latest signers.
```

contracts/axelar-gateway/tests/gateway.rs
```diff
@@ -201,8 +203,7 @@ fn approve_messages_skip_duplicate_message() {
     client.approve_messages(&messages, &proof);
 
     // should not throw an error, should just skip
```
milapsheth, line 205: remove comments everywhere

contracts/axelar-gateway/tests/gateway.rs
```diff
@@ -201,8 +203,7 @@ fn approve_messages_skip_duplicate_message() {
...
-    let res = client.try_approve_messages(&messages, &proof);
-    assert!(res.is_ok());
+    assert!(client.try_approve_messages(&messages, &proof).is_ok());
 
     // should not emit any more events (2 total because of rotate signers in auth)
```
milapsheth, line 208: same

contracts/axelar-gateway/tests/gateway.rs
```diff
@@ -201,8 +203,7 @@ fn approve_messages_skip_duplicate_message() {
...
     assert_eq!(env.events().all().len(), 2);
```
milapsheth, line 209: this is brittle, check that the difference between event counts before and after second call is the same

milapsheth (approved):

contracts/axelar-gateway/tests/auth.rs
```diff
@@ -149,7 +149,10 @@ fn rotate_signers_fail_empty_signers() {
     let proof = generate_proof(&env, data_hash, signers);
 
     assert_contract_err!(
-        client.try_rotate_signers(&empty_signers, &proof, &true),
+        invoke_auth!(
```
milapsheth, line 152: what's the point of using this macro? This test isn't checking auth, but a specific contract failure, so auth check isn't really relevant here, and can be checked separately

contracts/axelar-gateway/tests/auth.rs
```diff
@@ -170,7 +173,10 @@ fn rotate_signers_fail_zero_weight() {
     let proof = generate_proof(&env, data_hash, signers);
 
     assert_contract_err!(
-        client.try_rotate_signers(&new_signers.signers, &proof, &true),
+        invoke_auth!(
```
milapsheth, line 176: same everywhere

contracts/axelar-gateway/tests/auth.rs
milapsheth: it's more useful for successful executions where we want to ensure that the call would NOT fail due to an auth issue as well. So your usage of `assert_invoke_auth_ok` later is useful

contracts/axelar-gateway/tests/gateway.rs
```diff
@@ -75,12 +78,15 @@ fn validate_message() {
 
     let prev_event_count = env.events().all().len();
 
-    let approved = client.validate_message(
-        &contract_address,
-        &source_chain,
-        &message_id,
-        &source_address,
-        &payload_hash,
+    let approved = invoke_auth!(
```
milapsheth, line 81: make `assert_invoke_auth_ok` return the unwrapped result instead of creating a new macro

contracts/axelar-gateway/tests/auth.rs
milapsheth: Use `client.mock_all_auths().try_rotate_signers(&empty_signers, &proof, &true)` instead to explicitly mock it. This is more precise than the generic `env.mock_all_auths()` so it doesn't ignore legitimate auth checks

contracts/axelar-gateway/tests/gateway.rs
hydrobeam: having `assert_invoke_auth_ok` return makes it so that all its current uses need to handle the return value (or else a warning for an unused `Result` pops up).  So I think `invoke_auth` serves some purpose for this use case.

milapsheth (approved):

packages/axelar-soroban-std/src/error.rs
```diff
@@ -108,14 +108,19 @@ macro_rules! assert_some {
...
 
         match call_result {
-            Ok(_) => {}
+            Ok(outer) => {
+                match outer {
+                    Ok(inner) => {inner},
+                    Err(_) => panic!("Expected Ok result, but got an error."),
+                }
+            }
             Err(_) => panic!("Expected Ok result, but got an error."),
```
milapsheth, line 124:
```suggestion
                    Err(err) => panic!("Expected Ok result, but got an error {}", err),
                }
            }
            Err(err) => panic!("Expected Ok result, but got an error {}", err),
```

PR #100, opened 2024-12-06
ci: upload contracts to CF R2 from main

hydrobeam:
closes AXE-6843

the new workflow is mostly adapted from `release.yaml`

Test: https://github.com/axelarnetwork/axelar-cgp-soroban-testing/actions/runs/13063535201


.github/workflows/release-hash.yaml
```diff
@@ -0,0 +1,55 @@
...
+      package-version: "${{ github.sha }}"
+
+      # CF Bucket related variables
+      cf-bucket-name: "${{ vars.CF_BUCKET_NAME }}"
+
+      # The root key to be used for accessing the configs. (ex: `test-root-key` puts releases in `test-root-key/*`)
+      cf-config-bucket-root-key: "${{ vars.CF_BUCKET_ROOT_KEY }}"
+
+      # we just want to upload to CF, do not run a github release
+      github-release: "false"
```
ahramy, line 49: Not sure if `github-release: "false"` is the correct usage. Since we just want to upload to CF and don't want to run a github release, wouldn't it be something like `cf-upload-wasm-to-bucket: "true"`? This is somewhat confusing to me.

.github/workflows/reusable-build-upload.yaml
```diff
@@ -33,6 +33,12 @@ on:
         required: true
         type: string
 
+      github-release:
+        description: "Whether to upload the build as a github release"
+        required: false
+        default: "true"
```
ahramy, line 39: Just wondering, why is it not required but defaulted to true?

.github/workflows/release-hash.yaml
hydrobeam: The idea is that we always upload to CF but only sometimes want to do a github release. For regular releases, we do want to upload to github, so I just added the option to disable it and disabled it when we "release" based on the hash.

.github/workflows/reusable-build-upload.yaml
hydrobeam: Mentioned it above, but, for regular releases we /do/ want to run a github release, so I had the default set to true so that I didn't have to change `release.yaml`.

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,67 @@
...
+
+  # Publishes a release in case the release isn't published
+  build-and-upload:
+    name: Build artifacts for ${{ matrix.releases.package_name }}-${{ matrix.releases.version }}
+    needs: define-matrix
+
+    # Once a release is done for a package, we iterate on each of these packages and build its corresponding artifacts and upload them
+    strategy:
+      matrix:
+        releases: ${{ fromJson(needs.define-matrix.outputs.releases) }}
```
milapsheth, line 41: Very inefficient, we're rebuilding each contract although they're all on the same commit. Separate the build and cloudflare upload into their own reusable workflows, so the build can be run once, and cloudflare push can be done for each contract

ahramy: make sense, updated

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,65 @@
+# Workflow to fetch the latest commit hash on the main branch and upload artifacts to CF storage.
+name: Build and upload from main
+on:
+  workflow_dispatch:
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  define-matrix:
+    name: Define Matrix
```
milapsheth, line 10:
```suggestion
  stellar-contract-names:
    name: Get all Stellar contracts
```
define-matrix doesn't make much sense

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,65 @@
...
+          echo "releases=$(echo "$RELEASES_JSON" | jq -c)" >> "$GITHUB_OUTPUT"
+
+  build:
+    needs: define-matrix
+    uses: ./.github/workflows/reusable-build.yaml
+    with:
+      commit-hash: ${{ needs.define-matrix.outputs.commit_hash }}
+
+  upload:
+    needs: [define-matrix, build]
```
milapsheth, line 44: This is still a confusing approach. There's a single commit being built on. And then each contract needs to be pushed to R2. So the first job should be `build` itself. Second action can be to list out all contracts, and then third is to publish.

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,65 @@
...
+        run: echo "hash=$(git rev-parse --short HEAD)" >> "$GITHUB_OUTPUT"
+
+      - name: Prepare JSON output for matrix
+        id: prepare-matrix
+        run: |
+          RELEASES_JSON=$(find contracts -maxdepth 1 -mindepth 1 -type d | sed 's|contracts/||' | jq -R . | jq -s --arg commit "${{ steps.get-commit-hash.outputs.hash }}" 'map({
+            package_name: .,
+            version: $commit,
+            package_git_tag: "\(.)_\($commit)"
+          })')
```
milapsheth, line 34: this is overkill, the commit is the same now, we just need to list all contract names

.github/workflows/reusable-build.yaml
```diff
@@ -0,0 +1,69 @@
...
+  build:
+    runs-on: blacksmith-8vcpu-ubuntu-2204
+    outputs:
+      artifact-name: ${{ steps.set-artifact-name.outputs.name }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
```
milapsheth, line 25:
```suggestion
```
is this needed?

.github/workflows/reusable-build.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Checkout specific commit
+        run: git checkout ${{ inputs.commit-hash }}
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          toolchain: 1.81.0
```
milapsheth, line 33: create a separate ticket to read this from the root Cargo.toml for all workflows. Avoids the need to hardcode

.github/workflows/reusable-build.yaml
```diff
@@ -0,0 +1,69 @@
...
+
+      - name: Set artifact name
+        id: set-artifact-name
+        run: |
+          echo "name=wasm-builds-${{ inputs.commit-hash }}" >> $GITHUB_OUTPUT
+
+      - name: Build all contracts
+        run: |
+          # 21.1.1 is the latest rust 1.76.0 compatible version, the version after is 21.2.0 which is compatible with 1.79.0 (different than the workspace's version)
+          cargo install --locked stellar-cli --version 22.2.0 --features opt
```
milapsheth, line 44:
```suggestion
          cargo install --locked stellar-cli --version ^22 --features opt
```
just fix the major version?

.github/workflows/reusable-build.yaml
```diff
@@ -0,0 +1,69 @@
...
+          targets: wasm32-unknown-unknown
+
+      - name: Set artifact name
+        id: set-artifact-name
+        run: |
+          echo "name=wasm-builds-${{ inputs.commit-hash }}" >> $GITHUB_OUTPUT
+
+      - name: Build all contracts
+        run: |
+          # 21.1.1 is the latest rust 1.76.0 compatible version, the version after is 21.2.0 which is compatible with 1.79.0 (different than the workspace's version)
```
milapsheth, line 43:
```suggestion
          # Install Stellar CLI compatible with the soroban-sdk version in Cargo.toml
```
could read this dynamically as well from Cargo.toml in the future

.github/workflows/reusable-build.yaml
```diff
@@ -0,0 +1,69 @@
...
+          cargo install --locked stellar-cli --version 22.2.0 --features opt
+
+          # Build all contracts
+          cargo wasm
```
milapsheth, line 47:
```suggestion
          stellar contract build
```

.github/workflows/reusable-upload.yaml
```diff
@@ -48,36 +56,22 @@ on:
         required: true
 
 jobs:
-  build-and-upload:
-    runs-on: blacksmith-8vcpu-ubuntu-2204
+  upload:
+    name: upload ${{ inputs.package-git-tag }}
```
milapsheth, line 60:
```suggestion
    name: Upload ${{ inputs.package-git-tag }}
```
add contract name as well?

.github/workflows/reusable-upload.yaml
```diff
@@ -1,9 +1,8 @@
-name: "Publish specific rust package"
+name: "Upload Contract to Cloudflare"
```
milapsheth, line 1:
```suggestion
name: "Upload Contract wasm to Cloudflare R2"
```

.github/workflows/reusable-upload.yaml
```diff
@@ -48,36 +56,22 @@ on:
...
-
-      - name: Build artifacts for ${{ inputs.package-name }}-v${{ inputs.package-version }}
-        run: |
-          echo "Building wasm for '${{ inputs.package-name }}-v${{ inputs.package-version }}'";
-          cargo install --locked stellar-cli --version 22.2.0 --features opt
-          cargo wasm -p ${{ inputs.package-name }}
-          stellar contract build
-          ./optimize.sh
+          name: ${{ inputs.artifact-name }}
+          path: wasm-builds
```
milapsheth, line 67: a caller might use a different folder, this should be an input

.github/workflows/reusable-build.yaml
```diff
@@ -0,0 +1,69 @@
...
+          # Remove unoptimized files and rename optimized ones
+          # This ensures we only keep the optimized versions
+          find . -type f -name "*.wasm" ! -name "*.optimized.wasm" -maxdepth 1 -delete
+          find . -name "*.optimized.wasm" -maxdepth 1 -exec sh -c 'mv "$0" "${0%.optimized.wasm}.wasm"' {} \;
+
+          # Move all optimized WASM files to the builds directory
+          mv *.wasm ../../../wasm-builds/
+
+      - name: Upload artifact
+        uses: actions/upload-artifact@v4
```
milapsheth, line 65: needs to be uploaded just to communicate between jobs?

.github/workflows/reusable-build.yaml
```diff
@@ -0,0 +1,69 @@
...
+          # Process in the release directory
+          cd target/wasm32-unknown-unknown/release
+
+          # Remove unoptimized files and rename optimized ones
+          # This ensures we only keep the optimized versions
+          find . -type f -name "*.wasm" ! -name "*.optimized.wasm" -maxdepth 1 -delete
+          find . -name "*.optimized.wasm" -maxdepth 1 -exec sh -c 'mv "$0" "${0%.optimized.wasm}.wasm"' {} \;
+
+          # Move all optimized WASM files to the builds directory
+          mv *.wasm ../../../wasm-builds/
```
milapsheth, line 62:
```suggestion
          # Process in the release directory
          cd target/wasm32-unknown-unknown/release

          # Remove unoptimized files and rename optimized ones
          # This ensures we only keep the optimized versions
          find target/wasm32-unknown-unknown/release -type f -name "*" ! -name "*.optimized.wasm" -maxdepth 1 -delete
          find target/wasm32-unknown-unknown/release -name "*.optimized.wasm" -maxdepth 1 -exec sh -c 'mv "$0" "${0%.optimized.wasm}.wasm"' {} \;
```
reuse the existing folder?

milapsheth (approved):

.github/workflows/pre-release.yaml
```diff
@@ -0,0 +1,66 @@
+# Workflow to fetch the latest commit hash on the main branch and upload artifacts to CF storage.
```
milapsheth, line 1: `# Build Stellar contracts on every commit to main and upload the wasm code to Cloudflare R2

.github/workflows/reusable-upload.yaml
```diff
@@ -1,37 +1,43 @@
...
 on:
   workflow_call:
     inputs:
-      # Package related variables
       package-name:
         description: "The package name to use (ex: gz-srv)"
         type: string
         required: true
         default: ""
-
```
milapsheth, line 12: why remove newlines? it's more readable that way

.github/workflows/pre-release.yaml
milapsheth: But we aren't releasing to crates.io here. Take a look at how the `version` and `package_git_tag` are being used below. The commit hash can be used there directly instead of creating this confusing JSON

.github/workflows/release.yaml
```diff
@@ -6,15 +6,14 @@ on:
   pull_request:
     branches:
       - main
-      - 'releases/**'
+      - "releases/**"
```
milapsheth, line 9: Just `releases/**` is fine, consistent with other actions

.github/workflows/release.yaml
```diff
@@ -56,35 +59,33 @@ jobs:
...
           echo "releases=$(echo '${{ steps.publish-release.outputs.releases }}' | jq -c '.')" >> $GITHUB_OUTPUT
 
-  # Creates other artifacts needed (`wasm` files)
-  build-and-upload:
-    name: Build artifacts for ${{ matrix.releases.package_name }}-v${{ matrix.releases.version }}
+  build:
     needs: publish-release
+    uses: ./.github/workflows/reusable-build.yaml
+    with:
+      commit-hash: ${{ needs.publish-release.outputs.commit_hash }}
```
milapsheth, line 66: Is this correct? This is creating a build on the current commit hash instead of using the release tags created for every contract by release-plz. So, the built wasm files won't correspond to the release tag? The release matrix needs to be applied for both build and upload, unless all release tags are the same commit

PR #101, opened 2024-12-06
refactor: move interface storage into mod.rs

cgorenflo:


packages/axelar-soroban-std/src/interfaces/mod.rs
```diff
@@ -7,3 +7,31 @@ mod upgradable;
...
+
+    use soroban_sdk::contracttype;
+
+    // add a separate data key type for each interface. Using a single enum could lead to unintentionally breaks
+    // of unrelated interfaces, because the key serialization is variant order dependent.
+
+    #[contracttype]
+    pub enum OperatorDataKey {
+        Interfaces_Operator,
+    }
```
milapsheth, line 26:
```suggestion
    mod operator {
        #[contracttype]
        pub enum DataKey {
            Interfaces_Operator,
        }
    }
```
Perhaps?

packages/axelar-soroban-std/src/interfaces/mod.rs
```diff
@@ -7,3 +7,31 @@ mod upgradable;
...
+/// If a contract implements a contract type that would result in a collision with a key defined here,
+/// the linter will complain about it. So as long as contracts follow regular naming conventions,
+/// there is no risk of collisions.
+mod storage {
+    #![allow(non_camel_case_types)]
+
+    use soroban_sdk::contracttype;
+
+    // add a separate data key type for each interface. Using a single enum could lead to unintentionally breaks
+    // of unrelated interfaces, because the key serialization is variant order dependent.
```
milapsheth, line 21: Is it order dependent? I thought it only depends on the variant name. Still separate types are nicer

milapsheth (approved):

PR #105, opened 2024-12-08
feat: add macros for shared interfaces

milapsheth:
[AXE-6852]

[AXE-6852]: https://axelarnetwork.atlassian.net/browse/AXE-6852?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

Shared interfaces can now be implemented via derive macros:
```rust
use axelar_soroban_std::{Operatable, Ownable, Upgradable};

#[contract]
#[derive(Operatable, Ownable, Upgradable)]
pub struct AxelarGateway;
```


packages/axelar-soroban-std-derive/tests/testdata/contract.rs
cgorenflo, line 1: how does this compile? Upgradable requires Ownable, and now it's not implemented

cgorenflo (commented):
I do like the approach, but a few notes:
- I'd prefer if contracts don't have to depend directly on the derive crate, instead have an indirect import (potentially with feature flag "derive") in the axelar-soroban-sdk. See the amplifier repo for how to do that
- remove the leading `::` in paths, it makes the code much more brittle, like with goldie, and you have to define the import at the root level, it won't look at the local relative context
- you only showed the trivial migration data example, how does it work with more complex data?
- Are you sure that the contract proc macro actually picks up on the generated code and creates entrypoints for it?

packages/axelar-soroban-std-derive/src/lib.rs
```diff
@@ -0,0 +1,193 @@
...
+/// }
+/// # }
+/// ```
+#[proc_macro_attribute]
+pub fn ownable(_attr: TokenStream, input: TokenStream) -> TokenStream {
+    let input = parse_macro_input!(input as DeriveInput);
+    let name = &input.ident;
+
+    let expanded = quote! {
+        use ::axelar_soroban_std::interfaces::OwnableInterface;
```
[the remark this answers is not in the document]
milapsheth: The method seems to be used in the soroban generated Client code, which requires importing the trait
```
error[E0599]: no function or associated item named `migrate` found for struct `upgradable::Contract` in the current scope
  --> packages/axelar-soroban-std-derive/tests/test.rs:56:5
   |
56 |     #[upgradable]
   |     ^^^^^^^^^^^^^ function or associated item not found in `Contract`
57 |     #[contract]
58 |     pub struct Contract;
   |     ------------------- function or associated item `migrate` not found for this struct
   |
   = help: items from traits can only be used if the trait is in scope
   = note: this error originates in the attribute macro `upgradable` (in Nightly builds, run with -Z macro-backtrace for more info)
help: trait `MigratableInterface` which provides `migrate` is implemented but not in scope; perhaps you want to import it
   |
56 +     use axelar_soroban_std::interfaces::MigratableInterface;
```

milapsheth:
> remove the leading :: in paths, it makes the code much more brittle, like with goldie, and you have to define the import at the root level, it won't look at the local relative context
The goal was to ensure it doesn't check the local context to avoid potential name collisions. The Rust [doc](https://doc.rust-lang.org/reference/procedural-macros.html#procedural-macro-hygiene) mentioned the `::` prefix as well, although collision is unlikely in practice. So, I don't have a strong opinion

packages/axelar-soroban-std-derive/tests/testdata/contract.rs
milapsheth: there's a separate `#[ownable]` applied to the contract

milapsheth:
> I'd prefer if contracts don't have to depend directly on the derive crate, instead have an indirect import (potentially with feature flag "derive") in the axelar-soroban-sdk. See the amplifier repo for how to do that

Yeah, forgot about that. Refactored

packages/axelar-soroban-std-derive/.cargo/config.toml
```diff
@@ -0,0 +1,2 @@
+[alias]
+wasm = "build --release --target wasm32-unknown-unknown"
```
ahramy, line 2: How does this alias work? if it's something like `cargo wasm`, could you update README as well?

cgorenflo (approved):

PR #111, opened 2024-12-10
chore: fix docstring and update cargo lint

ahramy:


packages/axelar-soroban-std/src/address.rs
```diff
@@ -10,7 +10,7 @@ impl AddressExt for Address {
     /// Returns Stellar's "dead" address, represented by the constant `ZERO_ADDRESS`.
     ///
     /// # Reference
-    /// - Stellar GitHub: https://github.com/stellar/js-stellar-base/blob/master/test/unit/address_test.js
+    /// - Stellar GitHub: <https://github.com/stellar/js-stellar-base/blob/master/test/unit/address_test.js>
```
milapsheth:
```suggestion
    /// - Stellar [GitHub](https://github.com/stellar/js-stellar-base/blob/master/test/unit/address_test.js)
```
use markdown syntax

milapsheth (approved):

packages/axelar-soroban-std/src/address.rs
```diff
@@ -8,7 +8,7 @@ pub trait AddressExt {
 
 impl AddressExt for Address {
     /// Returns Stellar's "dead" address, represented by the constant `ZERO_ADDRESS`.
-    ///
+    /// test
```
milapsheth, line 11: bad merge?

ahramy: I was testing it. removed

packages/axelar-soroban-std/src/address.rs
ahramy: updated

PR #114, opened 2024-12-12
feat(its): update event test with new event test utils

AttissNgo:
[AXE-6848](https://axelarnetwork.atlassian.net/browse/AXE-6848)

[AXE-6848]: https://axelarnetwork.atlassian.net/browse/AXE-6848?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/tests/utils/mod.rs
```diff
@@ -1,14 +1,56 @@
...
+    testutils::Address as _, token::StellarAssetClient, Address, Bytes, BytesN, Env, IntoVal,
+    String, Symbol, Val,
+};
 
 pub const HUB_CHAIN: &str = "hub_chain";
 
 const INTERCHAIN_TOKEN_WASM_HASH: &[u8] = include_bytes!("../testdata/interchain_token.wasm");
 
+#[derive(Debug, PartialEq, Eq)]
+pub struct InterchainTransferReceivedEvent {
```
ahramy, line 21: Just wondering, is specifying `topic1, topic2, ...` necessary? It seems not, for example, OperatorshipTransferredEvent doesn't use it: https://github.com/axelarnetwork/axelar-cgp-soroban/blob/a90ac0b9d2d66533f32df8a9e5e15e483ec59c5d/packages/axelar-soroban-std/src/interfaces/operatable.rs#L48

contracts/interchain-token-service/src/contract.rs
```diff
@@ -260,15 +260,15 @@ impl AxelarExecutableInterface for InterchainTokenService {
     ) {
         let _ = Self::validate_message(&env, &source_chain, &message_id, &source_address, &payload);
 
-        let _ = Self::execute_message(
+        event::executed(
```
milapsheth, line 263: actually, this isn't needed. the gateway `validate_message` already emits the event, which is meant to be consistent with EVM gateway
https://github.com/axelarnetwork/axelar-cgp-soroban/blob/60b818d31aa250f1d57e2ef259780a2f3d2ed40e/contracts/axelar-gateway/src/contract.rs#L123

there's no need for the app to emit this event (EVM ITS doesn't do it as well)

contracts/interchain-token-service/src/contract.rs
```diff
@@ -260,15 +260,15 @@ impl AxelarExecutableInterface for InterchainTokenService {
...
             &env,
             source_chain.clone(),
             message_id.clone(),
             source_address.clone(),
             payload.clone(),
         );
 
-        event::executed(&env, source_chain, message_id, source_address, payload);
+        let _ = Self::execute_message(&env, source_chain, message_id, source_address, payload);
```
milapsheth, line 271: `let _ ` is a bug. It's ignoring the error being returned by `execute_message` which will prevent tx from reverting. Instead, return the error. Since this is an interface/trait method, the error type needs to be parameterized on the trait. See `MigratableInterface` for reference

contracts/interchain-token-service/tests/interchain_transfer.rs
```diff
@@ -107,20 +112,17 @@ fn interchain_transfer_receive() {
...
-            Symbol::new(&env, "interchain_transfer_received"),
-            String::from_str(&env, HUB_CHAIN),
-            token_id,
-            sender,
-            recipient,
-            amount,
-        ),
-        (data,),
-    );
+    assert!(expected_event.matches(&env, &env.events().all().last().unwrap()));
```
milapsheth, line 127: see existing usage with `fmt_last_emitted_event` for how we should test the event

contracts/interchain-token-service/tests/utils/mod.rs
milapsheth, line 21: The event should be defined within the contract crate and emitted within the contract. It's not meant to just be defined in the test. See existing usage

contracts/interchain-token-service/src/contract.rs
AttissNgo: Handled in separate PR

contracts/interchain-token-service/src/event.rs
```diff
@@ -1,47 +1,78 @@
...
-    let topics = (
-        Symbol::new(env, "executed"),
-        source_chain,
-        message_id,
-        source_address,
-    );
-    env.events().publish(topics, (payload,));
+#[derive(Debug, PartialEq, Eq)]
+pub struct InterchainTransferReceivedEvent {
+    pub original_source_chain: String,
```
milapsheth, line 19:
```suggestion
    pub source_chain: String,
```

contracts/interchain-token-service/src/event.rs
```diff
@@ -1,47 +1,78 @@
...
+            self.original_source_chain.as_val(),
+            self.token_id.to_val(),
+            self.source_address.to_val(),
+            self.destination_address.to_val(),
+            self.amount,
+        )
+    }
+
+    fn data(&self, _env: &Env) -> impl IntoVal<Env, Val> + Debug {
+        (self.data.clone(),)
```
milapsheth, line 63:
```suggestion
        (self.data.to_val(),)
```

contracts/interchain-token-service/tests/utils/mod.rs
```diff
@@ -1,14 +1,22 @@
...
 use axelar_soroban_std::types::Token;
+// use interchain_token_service::event::InterchainTransferReceivedEvent;
 use interchain_token_service::{InterchainTokenService, InterchainTokenServiceClient};
 use soroban_sdk::{testutils::Address as _, token::StellarAssetClient, Address, Env, String};
 
 pub const HUB_CHAIN: &str = "hub_chain";
 
 const INTERCHAIN_TOKEN_WASM_HASH: &[u8] = include_bytes!("../testdata/interchain_token.wasm");
 
+// impl_event_testutils!(
```
milapsheth, line 14: remove commented code

milapsheth (approved):

contracts/interchain-token-service/src/event.rs
AttissNgo: Can't use `to_val` here as data is `Option` type

PR #118, opened 2024-12-13
feat(interchain-token-service): remote interchain token

ahramy:
[AXE-6892](https://axelarnetwork.atlassian.net/browse/AXE-6892)

[AXE-6892]: https://axelarnetwork.atlassian.net/browse/AXE-6892?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/src/contract.rs
```diff
@@ -238,16 +240,39 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
-        _env: &Env,
-        _caller: Address,
-        _salt: BytesN<32>,
-        _minter: Option<Bytes>,
-        _destination_chain: String,
-        _gas_token: Token,
+        env: &Env,
+        caller: Address,
+        salt: BytesN<32>,
+        minter: Option<Address>,
```
milapsheth, line 246: Let's remove the minter in this method. We'll just add a separate method in another PR that allows specifying the minter because this needs to be bytes since other chains have a different address format

contracts/interchain-token-service/src/contract.rs
```diff
@@ -238,16 +240,39 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
+        let token_id = Self::interchain_token_id(env, Address::zero(env), deploy_salt);
+
+        if let Some(ref minter) = minter {
+            ensure!(
+                Self::check_token_minter(env, token_id.clone(), minter.clone()),
+                ContractError::InvalidMinter
+            );
+        }
+
+        let registered_token_address = Self::token_id_config(env, token_id.clone()).token_address;
```
milapsheth, line 262:
```suggestion
        let registered_token_address = Self::token_id_config(env, token_id.clone()).token_address;
```
Use the token_address query instead?

contracts/interchain-token-service/src/contract.rs
```diff
@@ -238,16 +240,39 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
+        let token = token::Client::new(env, &registered_token_address);
+
+        let message = Message::DeployInterchainToken(DeployInterchainToken {
+            token_id: token_id.clone(),
+            name: token.name(),
+            symbol: token.symbol(),
+            decimals: token.decimals() as u8,
```
milapsheth, line 269: Implicit lossy cast. Check and revert if decimals exceeds u8 MAX. In general, we should have explicit checks for things like this

contracts/interchain-token-service/src/contract.rs
```diff
@@ -239,16 +252,49 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
+        let registered_token_address = Self::token_address(env, token_id.clone());
+        let token = token::Client::new(env, &registered_token_address);
+
+        let token_name = token.name();
+        let token_symbol = token.symbol();
+        let token_decimals = token.decimals();
 
-        todo!()
+        ensure!(
+            token_decimals <= u8::MAX as u32,
```
milapsheth, line 273: create token metadata from token, and call validate_token_metadata on it instead?

contracts/interchain-token-service/src/event.rs
```diff
@@ -56,6 +76,42 @@ impl Event for TrustedChainRemovedEvent {
...
+
+    fn data(&self, env: &Env) -> impl IntoVal<Env, Val> + Debug {
+        Vec::<Val>::new(env)
+    }
+}
+
+impl Event for InterchainTokenDeploymentStartedEvent {
+    fn topics(&self, env: &Env) -> impl Topics + Debug {
+        (
+            String::from_str(env, "interchain_token_deployment_started"),
```
milapsheth, line 100:
```suggestion
            Symbol::new(env, "interchain_token_deployment_started"),
```

contracts/interchain-token-service/src/event.rs
```diff
@@ -13,6 +13,26 @@ pub struct TrustedChainRemovedEvent {
...
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub struct InterchainTokenDeployedEvent {
+    pub token_id: BytesN<32>,
+    pub token_address: Address,
+    pub minter: Option<Address>,
+    pub name: String,
+    pub symbol: String,
+    pub decimals: u32,
```
milapsheth, line 23:
```suggestion
    pub decimals: u8,
```
since this is not a contracttype?

contracts/interchain-token-service/src/event.rs
```diff
@@ -13,6 +13,26 @@ pub struct TrustedChainRemovedEvent {
...
+}
+
+#[derive(Debug, PartialEq, Eq)]
+pub struct InterchainTokenDeploymentStartedEvent {
+    pub token_id: BytesN<32>,
+    pub name: String,
+    pub symbol: String,
+    pub decimals: u32,
+    pub minter: Option<Address>,
+    pub destination_chain: String,
```
milapsheth, line 33:
```suggestion
    pub token_id: BytesN<32>,
    pub token_address: Address,
    pub destination_chain: String,
    pub name: String,
    pub symbol: String,
    pub decimals: u32,
    pub minter: Option<Address>,
```
add token_address for indexing, and move destination chain above

contracts/interchain-token-service/src/event.rs
```diff
@@ -13,6 +13,26 @@ pub struct TrustedChainRemovedEvent {
     pub chain: String,
 }
 
+#[derive(Debug, PartialEq, Eq)]
+pub struct InterchainTokenDeployedEvent {
+    pub token_id: BytesN<32>,
+    pub token_address: Address,
+    pub minter: Option<Address>,
```
milapsheth, line 20: keep minter as the last field maybe, since it's optional

contracts/interchain-token-service/src/event.rs
milapsheth: it's fine to deviate a bit from the EVM event since it's extending it

contracts/interchain-token-service/src/interface.rs
```diff
@@ -42,12 +42,11 @@ pub trait InterchainTokenServiceInterface: AxelarExecutableInterface {
     ) -> Result<BytesN<32>, ContractError>;
 
     fn deploy_remote_interchain_token(
```
milapsheth, line 44: add docstring

contracts/interchain-token-service/tests/deploy_remote_interchain_token.rs
```diff
@@ -0,0 +1,112 @@
...
+use axelar_soroban_std::assert_contract_err;
+use axelar_soroban_std::events;
+use interchain_token_service::error::ContractError;
+use interchain_token_service::event::InterchainTokenDeploymentStartedEvent;
+use soroban_sdk::{testutils::Address as _, Address, BytesN, Env, IntoVal, String};
+use soroban_token_sdk::metadata::TokenMetadata;
+use utils::setup_env;
+use utils::setup_gas_token;
+
+fn setup_token_metadata(env: &Env, name: &str, symbol: &str, decimal: u32) -> TokenMetadata {
```
milapsheth, line 12: not a very useful helper, use TokenMetaData directly

milapsheth: Or implement a TokenMetaDataExt trait with a `new` method in `utils/mod.rs` for a nicer usage

contracts/interchain-token-service/tests/deploy_remote_interchain_token.rs
```diff
@@ -0,0 +1,112 @@
...
+        .mock_all_auths()
+        .set_trusted_chain(&destination_chain);
+
+    let deployed_token_id = client.mock_all_auths().deploy_remote_interchain_token(
+        &sender,
+        &salt,
+        &destination_chain,
+        &gas_token,
+    );
+
```
milapsheth, line 51:
```suggestion
```
keep the assert right after the call

contracts/interchain-token-service/tests/deploy_remote_interchain_token.rs
```diff
@@ -0,0 +1,112 @@
...
+        &sender,
+        &salt,
+        &token_meta_data,
+        &initial_supply,
+        &minter,
+    );
+
+    let destination_chain = String::from_str(&env, "ethereum");
+
+    client
```
milapsheth, line 41:
```suggestion
    client
```

contracts/interchain-token-service/tests/deploy_remote_interchain_token.rs
```diff
@@ -0,0 +1,112 @@
...
+        .mock_all_auths()
+        .set_trusted_chain(&destination_chain);
+
+    let deployed_token_id = client.mock_all_auths().deploy_remote_interchain_token(
```
milapsheth, line 45: check auth as well, since this test is about a successful call to this method

contracts/interchain-token-service/src/contract.rs
```diff
@@ -239,16 +252,49 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
+        salt: BytesN<32>,
+        destination_chain: String,
+        gas_token: Token,
     ) -> Result<BytesN<32>, ContractError> {
-        // TODO: implementation
+        caller.require_auth();
+
+        let deploy_salt = Self::interchain_token_deploy_salt(env, caller.clone(), salt);
+        let token_id = Self::interchain_token_id(env, Address::zero(env), deploy_salt);
+        let registered_token_address = Self::token_address(env, token_id.clone());
```
milapsheth, line 265: since the token address might not exist, query the token config manually, and return an explicit error if it doesn't exist?

contracts/interchain-token-service/src/event.rs
milapsheth: Let's use a smaller symbol then. The first topic should always be a symbol for consistency. String looks different in XDR

Maybe `token_deployment_started`

contracts/interchain-token-service/src/event.rs
milapsheth: Ah, I see

contracts/interchain-token-service/src/contract.rs
```diff
@@ -239,17 +252,80 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
         Ok(token_id)
     }
 
+    /// Deploys an interchain token to a remote chain.
+    ///
+    /// This function initiates the deployment of an interchain token to a specified
+    /// destination chain. It validates the token metadata, emits a deployment event,
+    /// and triggers the necessary cross-chain call.
+    ///
+    /// # Parameters
```
milapsheth, line 261: Is this the convention used by rustdoc to document params? If not, look into the rust convention

contracts/interchain-token-service/src/contract.rs
```diff
@@ -239,17 +252,80 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
+        let token_id = Self::interchain_token_id(env, Address::zero(env), deploy_salt);
+
+        ensure!(
+            env.storage()
+                .persistent()
+                .has(&DataKey::TokenIdConfigKey(token_id.clone())),
+            ContractError::InvalidTokenId
+        );
+
+        let token_address = Self::token_address(env, token_id.clone());
```
milapsheth, line 297: Get the token id config instead and return error if it's None or get it's token address instead of using two different approaches to check and retrieve. You can do this in a functional way.

contracts/interchain-token-service/src/contract.rs
```diff
@@ -239,17 +252,80 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
+        let token = token::Client::new(env, &token_address);
+        let token_meta_data = TokenMetadata {
+            name: token.name(),
+            decimal: token.decimals(),
+            symbol: token.symbol(),
+        };
+
+        InterchainTokenClient::new(env, &token_address).validate_token_metadata(&token_meta_data);
```
milapsheth, line 305: This shouldn't be an entry point of a contract. It's a helper. Move it to token.rs in std crate

contracts/interchain-token-service/src/event.rs
ahramy: got it, updated to use `token_deployment_started`

milapsheth (approved):

contracts/interchain-token-service/src/contract.rs
ahramy: updated.

contracts/interchain-token-service/src/contract.rs
ahramy: need to re-visit token_address retrieval and error handling. I will address it in a separate PR. created a small ticket for this: https://axelarnetwork.atlassian.net/browse/AXE-6985

contracts/interchain-token-service/src/contract.rs
ahramy: yeah, I looked up rustdoc conventions, it seems using `///` is a convention.
- https://doc.rust-lang.org/rust-by-example/meta/doc.html
- https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html
```
There are two doc comments: /// and //!.

/// documents the item following it whereas //! documents the parent item.
```

contracts/interchain-token-service/tests/deploy_remote_interchain_token.rs
```diff
@@ -0,0 +1,196 @@
...
+        minter: None,
+    });
+
+    let payload = HubMessage::SendToHub {
+        destination_chain: destination_chain.clone(),
+        message,
+    }
+    .abi_encode(&env);
+
+    let transfer_auth = auth_invocation!(&env,
```
milapsheth, line 108: Can you make the usage for this simpler, similar to assert_invoke_auth_ok so you just need to pass in the function call as is? This way the function call is typed by using a client

Also after this change, the succeeds and auth check shouldn't be separate tests, auth should be tested as part of success

ahramy: Created a ticket to track this, and will be working on it: https://axelarnetwork.atlassian.net/browse/AXE-6986

PR #119, opened 2024-12-14
Feat/its register canonical token

AttissNgo:
[AXE-6745](https://axelarnetwork.atlassian.net/browse/AXE-6745)


milapsheth:
PR scope should be `feat(interchain-token-service)` for consistency

contracts/interchain-token-service/src/contract.rs
```diff
@@ -145,6 +146,13 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
             .into()
     }
 
+    fn canonical_token_deploy_salt(env: &Env, token_address: Address) -> BytesN<32> {
```
milapsheth, line 149: this can be an internal function

contracts/interchain-token-service/tests/canonical_token.rs
```diff
@@ -0,0 +1,42 @@
...
+#[test]
+fn register_canonical_token_fails_if_already_registered() {
+    let (env, client, _, _) = setup_env();
+    let token_address = Address::generate(&env);
+
+    client.register_canonical_token(&token_address);
+
+    let result = client.try_register_canonical_token(&token_address);
+
+    assert_contract_err!(result, ContractError::TokenAlreadyRegistered);
```
milapsheth, line 41:
```suggestion
    assert_contract_err!(client.try_register_canonical_token(&token_address), ContractError::TokenAlreadyRegistered);
```

contracts/interchain-token-service/tests/canonical_token.rs
```diff
@@ -0,0 +1,42 @@
...
+
+#[test]
+fn register_canonical_token_succeeds() {
+    let (env, client, _, _) = setup_env();
+    let token_address = Address::generate(&env);
+
+    let expected_deploy_salt = client.canonical_token_deploy_salt(&token_address);
+    let expected_id = client.interchain_token_id(&Address::zero(&env), &expected_deploy_salt);
+
+    assert_eq!(client.register_canonical_token(&token_address), expected_id);
```
milapsheth, line 18: use assert_invoke_auth_ok to verify caller auth

contracts/interchain-token-service/src/interface.rs
```diff
@@ -60,4 +60,9 @@ pub trait InterchainTokenServiceInterface: AxelarExecutableInterface {
         metadata: Option<Bytes>,
         gas_token: Token,
     ) -> Result<(), ContractError>;
+
+    fn register_canonical_token(
```
milapsheth, line 64: add docstring

contracts/interchain-token-service/tests/canonical_token.rs
```diff
@@ -0,0 +1,57 @@
...
+        .keccak256(
+            &(
+                PREFIX_CANONICAL_TOKEN_SALT,
+                chain_name_hash,
+                token_address.clone(),
+            )
+                .to_xdr(&env),
+        )
+        .into();
+    let expected_id = client.interchain_token_id(&Address::zero(&env), &expected_deploy_salt);
```
milapsheth, line 30: use goldie to test the token_id derivation matches. Basically, `goldie` can be used to create expectations that a certain computed data doesn't change. We can use this for events, token ids, salts, hashes etc.

goldie can save one file per test, so you can create a separate test to check token_id is derived correctly.

Another option is to do something like:
https://github.com/axelarnetwork/axelar-amplifier/blob/fa798314a2035aeeba06cd4905ae4a9ed71b27b8/contracts/interchain-token-service/tests/execute.rs#L366

But I prefer the former here since they're distinct tests

contracts/interchain-token-service/src/contract.rs
```diff
@@ -431,4 +465,16 @@ impl InterchainTokenService {
             .get(&DataKey::TokenIdConfigKey(token_id))
             .expect("token id config not found")
     }
+
+    fn chain_name_hash(env: &Env) -> BytesN<32> {
+        let chain_name = Self::chain_name(env);
+        env.crypto().keccak256(&(chain_name).to_xdr(env)).into()
```
milapsheth, line 471:
```suggestion
        env.crypto().keccak256(&chain_name.to_xdr(env)).into()
```
a bit clearer, since it read like we're encoding a tuple

milapsheth (approved):

PR #120, opened 2024-12-15
feat(interchain-token-service): add interchain token executable

milapsheth:
[AXE-6871]

[AXE-6871]: https://axelarnetwork.atlassian.net/browse/AXE-6871?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/interchain-token-service/tests/executable.rs
+++ b/contracts/interchain-token-service/tests/executable.rs
@@ -2,14 +2,13 @@ mod utils;
 
 use axelar_gateway::testutils::{generate_proof, get_approve_hash};
 use axelar_gateway::types::Message as GatewayMessage;
-use axelar_soroban_std::events;
 use axelar_soroban_std::traits::BytesExt;
-use interchain_token_service::event::InterchainTransferReceivedEvent;
+use axelar_soroban_std::{assert_invoke_auth_err, events};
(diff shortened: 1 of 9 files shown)


contracts/interchain-token-service/tests/executable.rs
```diff
@@ -0,0 +1,203 @@
...
+    #[derive(Debug, PartialEq, Eq)]
+    pub struct ExecutedEvent {
+        pub source_chain: String,
+        pub message_id: String,
+        pub source_address: Bytes,
+        pub payload: Bytes,
+        pub token_id: BytesN<32>,
+        pub token_address: Address,
+        pub amount: i128,
+    }
```
ahramy, line 41: Since the `payload` will be in the data section and can be very large, would it be better to have the payload as the last parameter for the ExecutedEvent?

```suggestion
    #[derive(Debug, PartialEq, Eq)]
    pub struct ExecutedEvent {
        pub source_chain: String,
        pub message_id: String,
        pub source_address: Bytes,
        pub token_id: BytesN<32>,
        pub token_address: Address,
        pub amount: i128,
        pub payload: Bytes,
    }
```

ahramy (approved):

PR #121, opened 2024-12-17
fix(interchain-token-service): update error handling for token id config

ahramy:
[AXE-6985](https://axelarnetwork.atlassian.net/browse/AXE-6985)

[AXE-6985]: https://axelarnetwork.atlassian.net/browse/AXE-6985?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/interchain-token-service/src/contract.rs
+++ b/contracts/interchain-token-service/src/contract.rs
@@ -161,27 +161,16 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
     }
 
     /// Retrieves the address of the token associated with the specified token ID.
-    ///
-    /// # Arguments
-    /// * `env` - A reference to the environment in which the function operates.
-    /// * `token_id` - A 32-byte identifier for the token.
-    ///
-    /// # Returns
-    /// * `Address` - The address of the token associated with the given token ID.
-    fn token_address(env: &Env, token_id: BytesN<32>) -> Address {
-        Self::token_id_config(env, token_id).token_address
+    fn token_address(env: &Env, token_id: BytesN<32>) -> Result<Address, ContractError> {
+        Self::token_id_config(env, token_id).map(|config| config.token_address)
     }
 
     /// Retrieves the type of the token manager type associated with the specified token ID.
-    ///
-    /// # Arguments
-    /// * `env` - A reference to the environment in which the function operates.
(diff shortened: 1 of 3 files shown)


contracts/interchain-token-service/src/contract.rs
```diff
@@ -161,27 +161,16 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
-    ///
-    /// # Returns
-    /// * `TokenManagerType` - The type of the token manager associated with the given token ID.
-    fn token_manager_type(env: &Env, token_id: BytesN<32>) -> TokenManagerType {
-        Self::token_id_config(env, token_id).token_manager_type
+    fn token_manager_type(
+        env: &Env,
+        token_id: BytesN<32>,
+    ) -> Result<TokenManagerType, ContractError> {
+        Self::token_id_config(env, token_id).map(|config| config.token_manager_type)
```
milapsheth, line 173:
```suggestion
    fn token_manager_type(
        env: &Env,
        token_id: BytesN<32>,
    ) -> TokenManagerType {
        Self::token_id_config(env, token_id).expect("token id config not found").token_manager_type
```
these entrypoints should still return the type as before

contracts/interchain-token-service/src/contract.rs
```diff
@@ -509,11 +493,11 @@ impl InterchainTokenService {
                 token_handler::give_token(
                     env,
                     &destination_address,
-                    Self::token_id_config(env, token_id.clone()),
+                    Self::token_id_config(env, token_id.clone())?,
                     amount,
                 )?;
 
-                let token_address = Self::token_address(env, token_id.clone());
+                let token_address = Self::token_address(env, token_id.clone())?;
```
milapsheth, line 500: we're querying the token id twice. query it once, and retrieve the token address from within it

contracts/interchain-token-service/src/contract.rs
ahramy: updated

contracts/interchain-token-service/src/contract.rs
ahramy: yeah.. that's true. updated.

PR #122, opened 2024-12-17
feat(interchain-token-service): process deploy interchain token message

AttissNgo:
[AXE-6952](https://axelarnetwork.atlassian.net/browse/AXE-6952)

[AXE-6952]: https://axelarnetwork.atlassian.net/browse/AXE-6952?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/src/contract.rs
```diff
@@ -539,8 +523,59 @@ impl InterchainTokenService {
...
+
+                let token_metadata = TokenMetadata {
+                    name,
+                    symbol,
+                    decimal: decimals as u32,
+                };
+
+                let minter_address = if let Some(minter) = minter {
+                    let addr = Address::from_xdr(env, &minter)
+                        .map_err(|_| ContractError::InvalidMinter)?;
```
AttissNgo, line 552: Uncertain if `from_xdr` is the correct method here. I think `from_string_bytes` might be correct, because the Bytes should have already been deserialized from XDR when the message was decoded. 
https://docs.rs/soroban-sdk/22.0.0-rc.3.2/soroban_sdk/struct.Address.html#method.from_string_bytes

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -51,3 +61,372 @@ fn execute_fails_with_invalid_message() {
...
+        &source_chain,
+        &message_id_invalid_minter,
+        &source_address,
+        &payload_invalid_minter,
+    );
+}
+
+#[test]
+// #[should_panic(expected = "Error(Contract, #12)")] // ContractError::InvalidMinter
+#[should_panic]
```
AttissNgo, line 324: This is panicking at a lower level of the Host conversion. This leads me to believe that the conversion approach in `execute_message` is incorrect.

contracts/interchain-token-service/src/contract.rs
```diff
@@ -539,8 +523,59 @@ impl InterchainTokenService {
...
-            Message::DeployInterchainToken(_) => {
-                todo!()
+            Message::DeployInterchainToken(DeployInterchainToken {
+                token_id,
+                name,
+                symbol,
+                decimals,
+                minter,
+            }) => {
+                ensure!(!name.is_empty(), ContractError::EmptyTokenName);
```
ahramy, line 533: We can use `validate_token_metadata`. to validate name, symbol, decimals. reference: https://github.com/axelarnetwork/axelar-cgp-soroban/blob/6ec26221bd6a7583b65bde93a2f69a7abb4dacb9/contracts/interchain-token-service/src/contract.rs#L295-L301

contracts/interchain-token-service/src/error.rs
```diff
@@ -21,4 +21,7 @@ pub enum ContractError {
     TokenAlreadyRegistered = 15,
     InvalidTokenMetaData = 16,
     InvalidTokenId = 17,
+    TokenAlreadyDeployed = 18,
+    EmptyTokenName = 19,
```
ahramy, line 25: we don't need `EmptyTokenName` and `EmptyTokenSymbol` error here if we use `validate_token_metadata`

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -51,3 +61,372 @@ fn execute_fails_with_invalid_message() {
...
+    client.execute(
+        &source_chain,
+        &message_id_empty_symbol,
+        &source_address,
+        &payload_empty_symbol,
+    );
+}
+
+#[test]
+#[should_panic(expected = "Error(Contract, #12)")] // ContractError::InvalidMinter
```
ahramy, line 274: we can use `assert_contract_err!` for this case

contracts/interchain-token-service/tests/execute.rs
ahramy, line 324: Please specify the expected lower level panic error. i.e. `#[should_panic(expected = "HostError: Error(Context, InvalidAction)")]`

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -51,3 +61,372 @@ fn execute_fails_with_invalid_message() {
...
+    client.execute(
+        &source_chain,
+        &message_id_invalid_minter,
+        &source_address,
+        &payload_invalid_minter,
+    );
+}
+
+#[test]
+#[should_panic(expected = "Error(Contract, #18)")] // ContractError::TokenAlreadyDeployed
```
ahramy, line 375: Please use `assert_contract_err!`

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -51,3 +61,372 @@ fn execute_fails_with_invalid_message() {
...
+            name: token_metadata.name.clone(),
+            symbol: token_metadata.symbol.clone(),
+            decimals: token_metadata.decimal as u8,
+            minter: Some(sender),
+        }),
+    };
+    let payload = msg.abi_encode(&env).unwrap();
+    let payload_hash: BytesN<32> = env.crypto().keccak256(&payload).into();
+
+    let first_message_id = String::from_str(&env, "first_deployment");
```
ahramy, line 404: nit: does `_` required in the message?

contracts/interchain-token-service/tests/execute.rs
AttissNgo: This is being called by `execute`, which uses `panic_with_error!` in the case that `execute_message` returns an error, so `assert_contract_err!` doesn't work here.

contracts/interchain-token-service/tests/execute.rs
AttissNgo: The commented-out `should_panic` is the behavior I'd like here. The intention of this test is to catch an invalid minter address. I left this in for you to have a look, but the test should not remain if it doesn't properly catch an invalid address.

contracts/interchain-token-service/src/contract.rs
```diff
@@ -539,8 +523,57 @@ impl InterchainTokenService {
...
-                todo!()
+            Message::DeployInterchainToken(DeployInterchainToken {
+                token_id,
+                name,
+                symbol,
+                decimals,
+                minter,
+            }) => {
+                ensure!(
+                    !env.storage()
```
milapsheth, line 534: use helper

contracts/interchain-token-service/src/contract.rs
```diff
@@ -539,8 +523,57 @@ impl InterchainTokenService {
...
+                    ContractError::TokenAlreadyDeployed
+                );
+
+                let token_metadata = TokenMetadata {
+                    name,
+                    symbol,
+                    decimal: decimals as u32,
+                };
+
+                assert_ok!(validate_token_metadata(token_metadata.clone()));
```
milapsheth, line 546: same as the other PR. fail explicitly instead of using assert with panics

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -51,3 +61,376 @@ fn execute_fails_with_invalid_message() {
...
+            contract_address: client.address.clone(),
+            payload_hash: payload_hash_empty_symbol,
+        },
+    ];
+    let data_hash = get_approve_hash(&env, messages.clone());
+    let proof = generate_proof(&env, data_hash, signers);
+
+    gateway_client.approve_messages(&messages, &proof);
+
+    client.execute(
```
milapsheth, line 269: same, not tied down

contracts/interchain-token-service/src/contract.rs
```diff
@@ -526,8 +510,57 @@ impl InterchainTokenService {
...
+                    symbol,
+                    decimal: decimals as u32,
+                };
+
+                ensure!(
+                    validate_token_metadata(token_metadata.clone()).is_ok(),
+                    ContractError::InvalidTokenMetaData
+                );
+
+                let minter_address = if let Some(minter) = minter {
```
milapsheth, line 536: Make this functional, use `minter.and_then(..)?` instead of `if let ...` to transform to the address

contracts/interchain-token-service/src/contract.rs
```diff
@@ -526,8 +510,57 @@ impl InterchainTokenService {
...
+                    let addr = Address::from_xdr(env, &minter).expect("address conversion failed");
```
milapsheth, line 537: This can fail due to a user input, should raise an error instead. Panics should only be used for unexpected scenarios, i.e due to coding errors (e.g. invalid initializations)

contracts/interchain-token-service/src/contract.rs
```diff
@@ -526,8 +510,57 @@ impl InterchainTokenService {
...
+
+                    ensure!(
```
milapsheth, line 539: Do we have this check on the receiving side for EVM ITS?

contracts/interchain-token-service/src/contract.rs
milapsheth, line 536:
```suggestion
                let minter = if let Some(minter) = minter {
```
Avoid using type names as suffixes. You can redefine the variable instead

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -51,3 +62,374 @@ fn execute_fails_with_invalid_message() {
...
+            message_id: message_id.clone(),
+            source_address: source_address.clone(),
+            contract_address: client.address.clone(),
+            payload_hash,
+        },
+    ];
+    let data_hash = get_approve_hash(&env, messages.clone());
+    let proof = generate_proof(&env, data_hash, signers);
+
+    gateway_client.approve_messages(&messages, &proof);
```
milapsheth, line 162: Create a helper to do this given a msg to reduce duplication in tests

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -51,3 +62,374 @@ fn execute_fails_with_invalid_message() {
...
+    assert_eq!(token.symbol(), token_metadata.symbol);
+    assert_eq!(token.decimals(), token_metadata.decimal);
+    assert_eq!(
+        client.token_manager_type(&token_id),
+        TokenManagerType::NativeInterchainToken
+    );
+}
+
+#[test]
+#[should_panic(expected = "Error(Contract, #16)")] // ContractError::InvalidTokenMetadata
```
milapsheth, line 181: Use assert_contract_err? Should avoid testing panics since it's not easy to refactor/maintain

milapsheth (changes requested):

contracts/interchain-token-service/src/contract.rs
AttissNgo: I had this as `map_err` before to return a `ContractError`, but I discovered that when a byte string which does not represent a valid Soroban address is inputted, instead of returning a `ConversionError` as I expected, the Host simply panics with `Error(Value, InvalidInput)`. I'm not sure how to get this to return a meaningful error. Please advise.

contracts/interchain-token-service/tests/execute.rs
AttissNgo: Since `execute` doesn't return a Result, but rather uses `panic_with_error` to return the error to the caller, `assert_contract_err` doesn't work here. Is there a better way to test this?

contracts/interchain-token-service/src/contract.rs
AttissNgo: I don't see it in the EVM contracts. I took it from the `deploy_interchain_token` in the Stellar implementation: https://github.com/axelarnetwork/axelar-cgp-soroban/blob/8430b939fe3a330161ee7318218a85ee6721254d/contracts/interchain-token-service/src/contract.rs#L189.

Seemed to make sense. Should I remove?

contracts/interchain-token-service/tests/execute.rs
milapsheth: makes sense, we'll need to fix the executable interface

contracts/interchain-token-service/src/contract.rs
```diff
@@ -526,8 +510,49 @@ impl InterchainTokenService {
...
+
+                ensure!(
+                    validate_token_metadata(token_metadata.clone()).is_ok(),
+                    ContractError::InvalidTokenMetaData
+                );
+
+                // Note: attempt to convert a byte string which doesn't represent a valid Soroban address fails at the Host level
+                let minter = minter
+                    .map(|m| Address::from_xdr(env, &m).map_err(|_| ContractError::InvalidMinter))
+                    .transpose()?;
```
milapsheth, line 539:
```suggestion
                    .map(|m| Address::from_xdr(env, &m))
                    .transpose()
                    .map_err(|_| ContractError::InvalidMinter)?;
```
a bit more readable

milapsheth (approved):

contracts/interchain-token-service/tests/execute.rs
```diff
@@ -51,3 +60,313 @@ fn execute_fails_with_invalid_message() {
...
+    let messages = vec![
+        &env,
+        GatewayMessage {
+            source_chain: source_chain.clone(),
+            message_id: message_id_empty_symbol.clone(),
+            source_address: source_address.clone(),
+            contract_address: client.address.clone(),
+            payload_hash: payload_hash_empty_symbol,
+        },
+    ];
```
milapsheth, line 255: This should be part of the test helper as well

PR #125, opened 2024-12-23
chore: simplify auth test utils

AttissNgo:
(AXE-6986)[https://axelarnetwork.atlassian.net/browse/AXE-6986]


packages/axelar-soroban-std/src/testutils.rs
```diff
@@ -73,41 +73,32 @@ where
 
 #[macro_export]
```
milapsheth, line 74: Add doc on how to use this. You can add an executable Rust example in the doc, see the derive macro crate for an example

milapsheth (approved):

PR #133, opened 2025-01-13
fix: Rename Stellar assert auth macros for brevity

TanvirDeol:
[AXE-7063](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-7063)

[AXE-7063]: https://axelarnetwork.atlassian.net/browse/AXE-7063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -371,6 +371,7 @@ version = "0.1.0"
 dependencies = [
  "axelar-soroban-std",
  "cfg-if",
+ "paste",
  "soroban-sdk",
 ]
 
@@ -385,6 +386,7 @@ dependencies = [
  "goldie",
  "hex",
  "hex-literal",
+ "paste",
  "rand",
  "rand_chacha",
  "soroban-sdk",
@@ -395,6 +397,7 @@ name = "axelar-operators"
 version = "0.1.0"
 dependencies = [
  "axelar-soroban-std",
+ "paste",
  "soroban-sdk",
 ]
 
@@ -405,6 +408,7 @@ dependencies = [
  "axelar-soroban-std-derive",
  "goldie",
  "hex",
+ "paste",
  "soroban-sdk",
  "soroban-token-sdk",
 ]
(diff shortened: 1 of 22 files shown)


packages/axelar-soroban-std/src/error.rs
```diff
@@ -106,13 +106,15 @@ macro_rules! assert_some {
...
 #[macro_export]
-macro_rules! assert_invoke_auth_ok {
+macro_rules! assert_auth {
     ($caller:expr, $client:ident . $method:ident ( $($arg:expr),* $(,)? )) => {{
         use soroban_sdk::IntoVal;
 
-        let call_result = $client
-            .mock_auths($crate::mock_auth!($caller, $client, $method, $($arg),*))
-            .$method($($arg),*);
+        paste! {
```
milapsheth, line 113: is there no way to do this without adding an external dependency?

milapsheth, line 113:
```suggestion
        paste::paste! {
```
to avoid local imports everywhere it's used

contracts/axelar-gas-service/Cargo.toml
```diff
@@ -13,6 +13,7 @@ crate-type = ["cdylib", "rlib"]
 axelar-soroban-std = { workspace = true }
 cfg-if = { workspace = true }
 soroban-sdk = { workspace = true }
+paste = { workspace = true }
```
milapsheth, line 16: this is only needed for tests so it should be under dev-dependencies

Cargo.lock
milapsheth, line 1: see https://www.conventionalcommits.org/en/v1.0.0/#summary for PR titles and available scopes

Also see previous commits for the pattern we use. This PR is a `refactor: ..`. Also, use lower case for the first word

Cargo.toml
```diff
@@ -25,6 +25,7 @@ alloy-sol-types = { version = "0.8.14", default-features = false, features = [
 ] }
 goldie = "0.5.0"
 hex = { version = "0.4" }
+paste = "1.0"
 
 [workspace.lints.clippy]
```
milapsheth, line 30: gh actions are failing on the PR

PR #138, opened 2025-01-14
refactor: rename assert_auth macros

TanvirDeol:
[AXE-7063](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-7063)

[AXE-7063]: https://axelarnetwork.atlassian.net/browse/AXE-7063?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


TanvirDeol:
@milapsheth 
Based on my research I don't think it's possible to combine string literals `"try_"` with identifiers `$method` without external dependencies like `paste`. There were other solutions which also didn't seem feasible:

- using `concat_idents!`: a macro to concat identifiers but its only availiable in nightly Rust versions
- using mapping (e.g `[method] => $try_method`): doesn't seem practical since theres many different identifiers
- Creating new identifiers (`Ident::new()`), procedural macros: all of these also require external dependencies

Let me know if you think otherwise. Maybe there is something in the existing dependencies that is relevant.

packages/axelar-soroban-std/src/error.rs
```diff
@@ -127,14 +129,15 @@ macro_rules! assert_invoke_auth_ok {
 }
 
 #[macro_export]
-macro_rules! assert_invoke_auth_err {
+macro_rules! assert_auth_err {
     ($caller:expr, $client:ident . $method:ident ( $($arg:expr),* $(,)? )) => {{
         use soroban_sdk::{IntoVal, xdr::{ScError, ScErrorCode, ScVal}};
 
+        paste::paste! {
         let call_result = $client
```
milapsheth, line 137: indent the paste body?

packages/axelar-soroban-std/src/error.rs
```diff
@@ -106,13 +106,15 @@ macro_rules! assert_some {
...
 #[macro_export]
-macro_rules! assert_invoke_auth_ok {
+macro_rules! assert_auth {
     ($caller:expr, $client:ident . $method:ident ( $($arg:expr),* $(,)? )) => {{
         use soroban_sdk::IntoVal;
 
-        let call_result = $client
-            .mock_auths($crate::mock_auth!($caller, $client, $method, $($arg),*))
-            .$method($($arg),*);
+        paste::paste! {
```
milapsheth, line 113: add a comment on why paste is needed

milapsheth (approved):

Cargo.toml
```diff
@@ -25,6 +25,7 @@ alloy-sol-types = { version = "0.8.14", default-features = false, features = [
 ] }
 goldie = "0.5.0"
 hex = { version = "0.4" }
+paste = "1.0"
 
 [workspace.lints.clippy]
```
milapsheth, line 30: after updating the PR from main, some more usages of the macro need to be fixed

packages/axelar-soroban-std/src/error.rs
```diff
@@ -106,13 +106,15 @@ macro_rules! assert_some {
...
+            let call_result = $client
+                .mock_auths($crate::mock_auth!($caller, $client, $method, $($arg),*))
```
milapsheth, line 115: update `mock_auth` macro to remove the `.replace("try_", "")` which isn't needed anymore

packages/axelar-soroban-std/src/error.rs
```diff
@@ -110,10 +110,11 @@ macro_rules! assert_auth {
     ($caller:expr, $client:ident . $method:ident ( $($arg:expr),* $(,)? )) => {{
         use soroban_sdk::IntoVal;
 
+        // Paste is used concatenate the method name with the try_ prefix
```
milapsheth, line 113:
```suggestion
        // Paste is used to concatenate the method name with the `try_` prefix
```

milapsheth (approved):

PR #142, opened 2025-01-14
feat(interchain-token-service): add associated error types for ITS executable interface

cgorenflo:


contracts/interchain-token-service/src/executable.rs
```diff
@@ -24,10 +25,11 @@ pub trait InterchainTokenExecutableInterface {
         token_id: BytesN<32>,
         token_address: Address,
         amount: i128,
-    );
+    ) -> Result<(), <Self as InterchainTokenExecutableInterface>::Error>;
 
     /// Ensure that only the interchain token service can call [`execute_with_interchain_token`].
-    fn validate(env: &Env) {
+    fn validate(env: &Env) -> Result<(), <Self as InterchainTokenExecutableInterface>::Error> {
```
milapsheth, line 31: This doesn't need a Result type. The app's business logic will only live in the execute method, after validate is called

contracts/interchain-token-service/tests/executable.rs
```diff
@@ -84,8 +91,8 @@ mod test {
             token_id: BytesN<32>,
             token_address: Address,
             amount: i128,
-        ) {
-            Self::validate(env);
+        ) -> Result<(), ContractError> {
+            let _ = Self::validate(env);
```
milapsheth, line 95:
```suggestion
            Self::validate(env);
```
The auth should still pass

contracts/interchain-token-service/tests/executable.rs
```diff
@@ -84,8 +91,8 @@ mod test {
...
```
milapsheth, line 96: Maybe return a custom error here on some condition (e.g. amount being less than some value etc.) to demonstrate usage and test the error

contracts/interchain-token-service/src/executable.rs
```diff
@@ -7,14 +7,42 @@
...
+        payload: Bytes,
+        token_id: BytesN<32>,
+        token_address: Address,
+        amount: i128,
+    ) -> Result<(), Self::Error>;
+}
+
+/// Interface for an Interchain Token Executable app. Use the [axelar_soroban_std::Executable] derive macro to implement this interface.
+///
+/// DO NOT IMPLEMENT THIS MANUALLY!
```
ahramy, line 40: Could you maybe emphasize this comment like this if it's important?
```suggestion
/// ***********************************************************************
/// * INTERFACE FOR AN INTERCHAIN TOKEN EXECUTABLE APP                   *
/// *                                                                    *
/// * Use the [axelar_soroban_std::Executable] derive macro to           *
/// * implement this interface.                                          *
/// *                                                                    *
/// * >>>>> DO NOT IMPLEMENT THIS MANUALLY! <<<<<                        *
/// ***********************************************************************
```

contracts/interchain-token-service/tests/executable.rs
```diff
@@ -99,6 +106,7 @@ mod test {
                 amount,
             }
             .emit(env);
+            Ok(())
```
ahramy, line 109: nit: new line

```suggestion
            .emit(env);
            
            Ok(())
```

contracts/interchain-token-service/src/executable.rs
cgorenflo: That get's rendered in an unreadable manner as a doctest, but I made the last line bold now

packages/axelar-std/src/interfaces/mod.rs
```diff
@@ -8,6 +8,12 @@ pub use operatable::*;
 pub use ownable::*;
 pub use upgradable::*;
 
+/// Marker trait for interfaces that should not be implemented by using .
```
milapsheth, line 11:
```suggestion
/// Marker trait for interfaces that should not be implemented by using `contractimpl`.
```
missing word i think

packages/axelar-std-derive/src/lib.rs
```diff
@@ -417,3 +428,45 @@ fn event_struct_fields(input: &DeriveInput) -> (EventStructFields, EventStructFi
 
     ((topic_idents, topic_types), (data_idents, data_types))
 }
+
+#[proc_macro_derive(Executable)]
+pub fn derive_executable(input: TokenStream) -> TokenStream {
```
milapsheth, line 433:
```suggestion
pub fn derive_interchain_token_executable(input: TokenStream) -> TokenStream {
```

contracts/interchain-token-service/src/executable.rs
```diff
@@ -7,14 +7,42 @@
 
 use soroban_sdk::{contractclient, Address, Bytes, BytesN, Env, String};
```
milapsheth, line 8: Re-export `Executable` macro from this crate? Feels a bit weird for apps to import ITS specific macro from `stellar-axelar-std`

contracts/interchain-token-service/src/executable.rs
```diff
@@ -7,14 +7,42 @@
...
+/// For security purposes and convenience, sender authorization and other commonly shared code necessary to implement that trait can be automatically generated with the [`axelar_soroban_std::Executable`] derive macro.
+/// All parts that are specific to an individual ITS executable contract are collected in this [`CustomExecutable`] trait and must be implemented by the contract to be compatible with the [`InterchainTokenExecutableInterface`] trait.
+///
+/// Do not add the implementation of [`CustomExecutable`] to the public interface of the contract, i.e. do not annotate the `impl` block with `#[contractimpl]`
+pub trait CustomExecutable {
+    /// The type of error the [`CustomExecutable::execute`] function returns. Generally matches the error type of the whole contract.
+    type Error: Into<soroban_sdk::Error>;
+
+    /// Returns the address of the interchain token service contract that is authorized to execute arbitrary payloads on this contract
     fn interchain_token_service(env: &Env) -> Address;
```
milapsheth, line 22: We want this to be exposed as an entrypoint for querying, so auto implement this on the InterchainTokenExecutableInterface as well?
```suggestion
    fn internal_interchain_token_service(env: &Env) -> Address;
```

contracts/interchain-token-service/src/executable.rs
```diff
@@ -7,14 +7,42 @@
...
 
+    /// The custom execution logic that takes in an arbitrary payload and a token.
+    /// At the time this function is called, the calling address has already been verified as the correct interchain token service contract.
+    fn execute(
```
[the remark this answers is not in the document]
cgorenflo: calling something internal and then having some external user implement it seems strange

contracts/interchain-token-service/tests/executable.rs
```diff
@@ -84,8 +93,8 @@ mod test {
             token_id: BytesN<32>,
             token_address: Address,
             amount: i128,
-        ) {
-            Self::validate(env);
+        ) -> Result<(), ContractError> {
+            ensure!(payload.len() != 1, ContractError::PayloadLenOne);
```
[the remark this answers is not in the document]
cgorenflo: not feasible because ITS doesn't execute empty data

contracts/interchain-token-service/src/executable.rs
```diff
@@ -6,24 +6,25 @@
...
     type Error: Into<soroban_sdk::Error>;
 
     /// Returns the address of the interchain token service contract that is authorized to execute arbitrary payloads on this contract
-    fn interchain_token_service(env: &Env) -> Address;
+    fn __interchain_token_service(env: &Env) -> Address;
 
     /// The custom execution logic that takes in an arbitrary payload and a token.
     /// At the time this function is called, the calling address has already been verified as the correct interchain token service contract.
-    fn execute(
+    fn __authorized_execute_with_token(
```
milapsheth, line 27: :+1:

milapsheth (approved):

PR #147, opened 2025-01-16
feat(interchain-token-service): encode source/recipient addresses as strings instead of XDR

TanvirDeol:
[AXE-7097](https://axelarnetwork.atlassian.net/browse/AXE-7097)

[AXE-7097]: https://axelarnetwork.atlassian.net/browse/AXE-7097?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/src/contract.rs
```diff
@@ -532,9 +532,7 @@ impl InterchainTokenService {
                 amount,
                 data,
             }) => {
-                let destination_address = Address::from_xdr(env, &destination_address)
-                    .map_err(|_| ContractError::InvalidDestinationAddress)?;
-
+                let destination_address = Address::from_string_bytes(&destination_address);
```
milapsheth, line 535: update the docstring to mention this expectation

contracts/interchain-token-service/src/contract.rs
```diff
@@ -597,10 +595,7 @@ impl InterchainTokenService {
                 );
 
                 // Note: attempt to convert a byte string which doesn't represent a valid Soroban address fails at the Host level
-                let minter = minter
-                    .map(|m| Address::from_xdr(env, &m))
-                    .transpose()
-                    .map_err(|_| ContractError::InvalidMinter)?;
+                let minter = minter.map(|m| Address::from_string_bytes(&m));
```
milapsheth, line 599: Also need to change `interchain_transfer` to encode caller address as a string converted to bytes instead of `XDR`

`source_address: caller.clone().to_xdr(env),`

packages/axelar-soroban-std/src/testutils.rs
```diff
@@ -71,6 +72,15 @@ where
     assert_emitted_event(env, -1, contract_id, topics, data);
 }
 
+// Converts Stellar address (soroban_sdk::Address) to soroban_sdk::Bytes
+pub fn address_to_bytes(env: &Env, address: &Address) -> Bytes {
```
milapsheth, line 76: See `address.rs`. Add a [to_bytes](https://github.com/axelarnetwork/axelar-cgp-soroban/blob/8239e4126cdccb4156f737dd6e20fad5c2bfc239/packages/axelar-soroban-std/src/address.rs#L9) trait method to use this in a cleaner way

contracts/interchain-token-service/src/contract.rs
```diff
@@ -534,11 +532,8 @@ impl InterchainTokenService {
                     ContractError::InvalidTokenMetaData
                 );
 
-                // Note: attempt to convert a byte string which doesn't represent a valid Soroban address fails at the Host level
-                let minter = minter
-                    .map(|m| Address::from_xdr(env, &m))
-                    .transpose()
-                    .map_err(|_| ContractError::InvalidMinter)?;
+                // Note: convert from Option<Bytes> to Option<Address> type
```
milapsheth, line 535:
```suggestion
                // Note: converting a byte string which doesn't represent a valid Soroban address fails at the Host level
```
the prior note is still valid since there's no explicit Result type being returned on a failure from this method

packages/axelar-std/src/address.rs
```diff
@@ -1,9 +1,11 @@
-use soroban_sdk::{Address, Env, String};
+use soroban_sdk::{Address, Bytes, Env, String};
 
 const ZERO_ADDRESS: &str = "GAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAWHF";
+const STELLAR_ADDRESS_LEN: usize = 56;
```
milapsheth, line 4:
```suggestion
const STELLAR_ADDRESS_LEN = ZERO_ADDRESS.len();
```

packages/axelar-std/src/address.rs
```diff
@@ -13,4 +15,12 @@ impl AddressExt for Address {
     fn zero(env: &Env) -> Address {
         Self::from_string(&String::from_str(env, ZERO_ADDRESS))
     }
+
+    // Converts Stellar address (soroban_sdk::Address) to soroban_sdk::Bytes
+    fn to_bytes(&self, env: &Env) -> Bytes {
+        let address_str = self.to_string();
+        let mut str_bytes = [0u8; STELLAR_ADDRESS_LEN];
+        address_str.copy_into_slice(&mut str_bytes);
+        Bytes::from_slice(env, &str_bytes)
```
milapsheth, line 24:
```suggestion
        let mut address_string_bytes = [0u8; STELLAR_ADDRESS_LEN];
        self.to_string().copy_into_slice(&mut address_string_bytes);
        Bytes::from_slice(env, &address_string_bytes)
```

packages/axelar-std/src/address.rs
```diff
@@ -13,4 +15,12 @@ impl AddressExt for Address {
...
+    }
 }
```
milapsheth, line 26: add unit test to convert address to bytes and back to ensure it works. you can try with zero address and random addresses

milapsheth: try different address types that are supported
https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.Address.html#method.from_string_bytes

packages/axelar-std/src/address.rs
```diff
@@ -1,9 +1,11 @@
-use soroban_sdk::{Address, Env, String};
+use soroban_sdk::{Address, Bytes, Env, String};
 
 const ZERO_ADDRESS: &str = "GAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAWHF";
+const STELLAR_ADDRESS_LEN: usize = 56;
 
 pub trait AddressExt {
     fn zero(env: &Env) -> Address;
+    fn to_bytes(&self, env: &Env) -> Bytes;
```
milapsheth, line 8:
```suggestion

    /// add a docstring
    fn to_string_bytes(&self, env: &Env) -> Bytes;
```
rename to be consistent with the from version

contracts/interchain-token-service/src/contract.rs
```diff
@@ -365,7 +365,7 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
 
         let message = Message::InterchainTransfer(InterchainTransfer {
             token_id,
-            source_address: caller.clone().to_xdr(env),
+            source_address: caller.to_bytes(env),
```
milapsheth, line 368: why didn't golden file change due to this? That should be updated as well. Or modify the tests to ensure this is covered

milapsheth (approved):

PR #155, opened 2025-01-18
feat(interchain-token-service): handle stellar native asset metadata

milapsheth:
[AXE-7165]

[AXE-7165]: https://axelarnetwork.atlassian.net/browse/AXE-7165?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/interchain-token-service/tests/testdata/deploy_remote_canonical_token_succeeds_native_token.golden
```diff
@@ -0,0 +1,3 @@
+contract: Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAVAX5)
```
milapsheth, line 1: Please review the new golden files carefully to ensure they match the expectations of the tests

contracts/interchain-token-service/src/error.rs
```diff
@@ -30,6 +30,7 @@ pub enum ContractError {
     NotApproved = 22,
     InvalidDestinationChain = 23,
     InvalidData = 24,
+    InvalidAssetName = 25,
```
ahramy: does this `InvalidAssetName` being used?

contracts/interchain-token-service/src/token_metadata.rs
```diff
@@ -0,0 +1,47 @@
...
+use crate::error::ContractError;
+
+const NATIVE_TOKEN_NAME: &str = "Stellar";
+const NATIVE_TOKEN_SYMBOL: &str = "XLM";
+const MAX_NAME_LENGTH: u32 = 32;
+
+pub fn token_metadata(
+    env: &Env,
+    token_address: &Address,
+    native_token_address: &Address,
```
ahramy, line 15: We don't need to pass `native_token_address` as an argument since it is only one and unique per contract. we could just use `&Self::native_token_address(env)`

contracts/interchain-token-service/src/token_metadata.rs
```diff
@@ -0,0 +1,47 @@
+use soroban_sdk::{token, Address, Env, String};
+use soroban_token_sdk::metadata::TokenMetadata;
+use stellar_axelar_std::ensure;
+use stellar_axelar_std::token::validate_token_metadata;
+
+use crate::error::ContractError;
+
+const NATIVE_TOKEN_NAME: &str = "Stellar";
+const NATIVE_TOKEN_SYMBOL: &str = "XLM";
+const MAX_NAME_LENGTH: u32 = 32;
```
ahramy, line 10: Just wondering, why 32 characters?

contracts/interchain-token-service/tests/deploy_remote_canonical_token.rs
```diff
@@ -127,6 +120,62 @@ fn deploy_remote_canonical_token_succeeds() {
...
+    let token_address = client.native_token_address();
+    let destination_chain = String::from_str(&env, "ethereum");
+
+    client.register_canonical_token(&token_address);
+
+    client
+        .mock_all_auths()
+        .set_trusted_chain(&destination_chain);
+    client
+        .mock_all_auths_allowing_non_root_auth()
```
ahramy, line 137: should we mock auth instead of using `mock_all_auths_allowing_non_root_auth`?

contracts/interchain-token-service/tests/deploy_remote_canonical_token.rs
```diff
@@ -127,6 +120,62 @@ fn deploy_remote_canonical_token_succeeds() {
...
+    let token_address = client.token_address(&token_id);
+    let destination_chain = String::from_str(&env, "ethereum");
+
+    client.register_canonical_token(&token_address);
+
+    client
+        .mock_all_auths()
+        .set_trusted_chain(&destination_chain);
+    client
+        .mock_all_auths_allowing_non_root_auth()
```
ahramy, line 171: same comment as above

contracts/interchain-token-service/tests/deploy_remote_canonical_token.rs
```diff
@@ -127,6 +120,62 @@ fn deploy_remote_canonical_token_succeeds() {
...
+    >(&env, -4));
+}
+
+#[test]
+fn deploy_remote_canonical_token_succeeds_without_name_truncation() {
+    let (env, client, _, _, _) = setup_env();
+    let spender = Address::generate(&env);
+    let gas_token = setup_gas_token(&env, &spender);
+
+    let token_metadata = TokenMetadata::new(&env, "name", "symbol", 255);
```
ahramy, line 151: Nit: Can we use more meaningful strings for the token name and symbol than 'name' and 'symbol'? i.e. "test", "TEST". When I look at the golden file has `String(name), String(symbol)`, it's a bit confusing whether this is the expected result or not.

contracts/interchain-token-service/src/token_metadata.rs
milapsheth: this is not a method on `Self` though

contracts/interchain-token-service/src/error.rs
milapsheth: removed in the follow up PR

contracts/interchain-token-service/src/token_metadata.rs
milapsheth: for now, we'll see if we want to increase it later

contracts/interchain-token-service/tests/deploy_remote_canonical_token.rs
milapsheth: not really checking auth here

contracts/interchain-token-service/src/token_metadata.rs
ahramy: Then we can access the `native_token_address` using `&DataKey::NativeTokenAddress`. Since it is only being used to compare whether the `token_address` is the `native_token_address` or not, does it need to be passed as an argument?

milapsheth: Passing in a data key is clunky and I'd also like to expose this in an entrypoint in ITS itself. We can get this in for now and I'll think about it more

ahramy (approved):

PR #156, opened 2025-01-18
fix(interchain-token-service): validate its hub source address

milapsheth:
[AXE-7190]

[AXE-7190]: https://axelarnetwork.atlassian.net/browse/AXE-7190?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/interchain-token-service/tests/execute.rs
+++ b/contracts/interchain-token-service/tests/execute.rs
@@ -5,8 +5,9 @@ use soroban_sdk::xdr::ToXdr;
 use soroban_sdk::{vec, Address, Bytes, BytesN, String};
 use soroban_token_sdk::metadata::TokenMetadata;
 use stellar_axelar_gateway::types::Message as GatewayMessage;
-use stellar_axelar_std::events;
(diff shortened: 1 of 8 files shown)


contracts/interchain-token-service/tests/execute.rs
```diff
@@ -33,9 +34,9 @@ fn execute_fails_without_gateway_approval() {
 fn execute_fails_with_invalid_message() {
```
AttissNgo, line 34: We should be checking errors on failed calls to `execute` rather than `should_panic`. I'll create a separate ticket for it.

AttissNgo (approved):

ahramy (approved):

contracts/interchain-token-service/tests/execute.rs
milapsheth: Yes, need to cleanup the tests

PR #161, opened 2025-01-20
refactor: make formatted events more readable

milapsheth:
[AXE-7220]

[AXE-7220]: https://axelarnetwork.atlassian.net/browse/AXE-7220?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/axelar-gateway/tests/testdata/call_contract.golden
```diff
@@ -1,3 +1,9 @@
-contract: Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHK3M)
-topics: (Symbol(contract_called), Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAITA4), String(ethereum), String(0x4EFE356BEDeCC817cb89B4E9b796dB8bC188DC59), BytesN<32>(86, 87, 13, 226, 135, 215, 60, 209, 203, 96, 146, 187, 143, 222, 230, 23, 57, 116, 149, 95, 222, 243, 69, 174, 87, 158, 233, 244, 117, 234, 116, 50))
-data: (Bytes(18, 52))
\ No newline at end of file
+Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHK3M)
```
milapsheth, line 1: I'm including the contract id to ensure that the event is emitted from the same contract (it's plausible that two contracts emit the same event, e.g. ownership transfer)

contracts/axelar-gateway/tests/testdata/call_contract.golden
```diff
@@ -1,3 +1,9 @@
...
+
+ContractCalledEvent {
+    caller: Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAITA4),
+    destination_chain: String(ethereum),
+    destination_address: String(0x4EFE356BEDeCC817cb89B4E9b796dB8bC188DC59),
+    payload_hash: BytesN<32>(86, 87, 13, 226, 135, 215, 60, 209, 203, 96, 146, 187, 143, 222, 230, 23, 57, 116, 149, 95, 222, 243, 69, 174, 87, 158, 233, 244, 117, 234, 116, 50),
+    payload: Bytes(18, 52),
```
milapsheth, line 8: Looking into tagging data fields with `#[data]` to catch changes to what goes in data vs topics. Ordering changes should be visible based on the struct fields

contracts/axelar-gateway/tests/testdata/call_contract.golden
```diff
@@ -1,3 +1,9 @@
-contract: Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHK3M)
```
milapsheth, line 1: Since the event topics/data methods are now auto-generated rather than custom, we can format the struct itself

contracts/axelar-gateway/tests/testdata/call_contract.golden
milapsheth: If tagging `#[data]` is trickier, another option is to print the tuple of types that goes into `topics` and `data`

contracts/axelar-gateway/tests/testdata/call_contract.golden
```diff
@@ -1,3 +1,17 @@
...
+    caller: Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAITA4),
+    destination_chain: String(ethereum),
+    destination_address: String(0x4EFE356BEDeCC817cb89B4E9b796dB8bC188DC59),
+    payload_hash: BytesN<32>(86, 87, 13, 226, 135, 215, 60, 209, 203, 96, 146, 187, 143, 222, 230, 23, 57, 116, 149, 95, 222, 243, 69, 174, 87, 158, 233, 244, 117, 234, 116, 50),
+    payload: Bytes(18, 52),
+}
+
+Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAHK3M)
+
+contract_called {
```
milapsheth, line 11: event schema

contracts/axelar-gateway/tests/testdata/call_contract.golden
```diff
@@ -1,3 +1,17 @@
...
+    #[topic] caller: Address,
+    #[topic] destination_chain: String,
+    #[topic] destination_address: String,
+    #[topic] payload_hash: BytesN < 32 >,
```
milapsheth, line 15: due to how tokens work, it splits up brackets, but a bit tricky to handle without using `std` so keeping it for now

packages/axelar-std-derive/src/event.rs
```diff
@@ -6,99 +6,96 @@ use syn::{DeriveInput, LitStr, Type};
...
+        fn from_event(env: &soroban_sdk::Env, topics: soroban_sdk::Vec<soroban_sdk::Val>, data: soroban_sdk::Val) -> Self {
+            use soroban_sdk::TryFromVal;
+
+            // Verify the event name matches
+            let event_name = soroban_sdk::Symbol::try_from_val(env, &topics.get(0)
+                .expect("missing event name in topics"))
+                .expect("invalid event name type");
+            assert_eq!(event_name, soroban_sdk::Symbol::new(env, #event_name), "event name mismatch");
+
+            let mut topic_idx = 1;
```
ahramy, line 43: nit: Could we add some comments since it is quite complicated?

ahramy (approved):

nbayindirli (approved):
lgtm.

packages/axelar-std-derive/src/event.rs
nbayindirli: Would personally push for further helper methods (instead of comments) if the code is not clear enough as is

packages/axelar-std/src/events.rs
```diff
@@ -1,60 +1,47 @@
...
-        let event = env
+        let (contract_id, topics, data) = env
             .events()
             .all()
             .get(idx as u32)
             .expect("no event found at the given index");
-        E::standardized_fmt(env, &event)
+
+        let event = E::from_event(env, topics, data);
+        std::format!("{:#?}\n\n{:?}\n\n{}", event, contract_id, E::schema(env))
```
cgorenflo, line 44: I specifically avoided parsing the emitted event back into an event. The goal of the golden files is to make sure the emitted data doesn't change, so we don't break consumers downstream. If we take that data and print the parsed event again, we only test that `from_event` works as expected, but it's quite feasible that both serialization and deserialization change, so printing the typed event stays the same, but the emitted event data looks different

cgorenflo (commented):
what's the purpose of this change?

PR #173, opened 2025-01-22
refactor: replace event testutils by IntoEvent macro

nbayindirli:
AXE-7227


packages/axelar-std/src/events.rs
```diff
@@ -56,107 +56,64 @@ mod testutils {
...
-            (self.topic1.clone(), self.topic2.clone(), self.topic3)
-        }
+    #[contract]
+    struct Contract;
 
-        fn data(&self, _env: &Env) -> impl IntoVal<Env, Val> + Debug {
-            (self.data1.clone(), self.data2.clone())
+    #[contractimpl]
+    impl Contract {
+        pub fn emit_test_event(env: &Env) {
```
milapsheth, line 88: we already added integration tests for this, so a unit test isn't needed. this should resolve the dependency issue as well i believe

packages/axelar-std/src/interfaces/upgradable.rs
```diff
@@ -1,13 +1,12 @@
 use core::fmt::Debug;
 
 use soroban_sdk::{
-    contractclient, symbol_short, BytesN, Env, FromVal, IntoVal, String, Topics, Val,
+    contractclient, BytesN, Env, FromVal, IntoVal, String, Val,
 };
+use crate::IntoEvent;
```
milapsheth, line 6: does it help if you import from the derive crate directly in this case?

packages/axelar-std/tests/event.rs
```diff
@@ -0,0 +1,199 @@
+use soroban_sdk::testutils::Address as _;
```
milapsheth, line 1: rebase the PR over main correctly. Here's an easier to rebase since this PR was pointing to your other PR:
- Restore the branch from the merged PR
- Merge that branch onto this PR, which is easier since the commit history is preserved
- Merge latest main onto this PR
- Delete old branch and merge this PR

milapsheth: we might want to play around with stacked diffs/graphite

milapsheth (approved):

PR #195, opened 2025-01-23
chore: update reusable build upload workflow

ahramy:


.github/workflows/reusable-build-upload.yaml
```diff
@@ -56,7 +56,7 @@ jobs:
         with:
```
milapsheth, line 56: should we increase number of cpus to 8vcpu above so it builds faster?

.github/workflows/reusable-build-upload.yaml
```diff
@@ -78,7 +78,7 @@ jobs:
           export PACKAGE_VERSION='v${{ inputs.package-version }}'
 
           export BASE_ARTIFACTS_DIR="./target/wasm32-unknown-unknown/release"
-          export ARTIFACT_NAME="axelar-cgp-soroban-wasm-${PACKAGE_NAME}-${PACKAGE_VERSION}"
+          export ARTIFACT_NAME="axelar-cgp-stellar-wasm-${PACKAGE_NAME}-${PACKAGE_VERSION}"
```
milapsheth, line 81: where is there such a prefix? seems unnecessary verbose, talk to Ayoub about removing it

milapsheth (approved):

.github/workflows/reusable-build-upload.yaml
ahramy: yeah, will do after a successful release

PR #198, opened 2025-01-23
chore: update build machine type to 8vcpu

ahramy:


.github/workflows/create-release-pr.yaml
```diff
@@ -10,7 +10,7 @@ concurrency: ${{ github.workflow }}-${{ github.ref }}
 jobs:
   create-release-pr:
     name: Create Release PR
-    runs-on: blacksmith-2vcpu-ubuntu-2204
+    runs-on: blacksmith-8vcpu-ubuntu-2204
```
milapsheth, line 13: Is this a slow action? If not, then 2vcpu should be enough

milapsheth (approved):

.github/workflows/create-release-pr.yaml
ahramy: It was a bit slow for the initial release but afterwards it should not be slow and computationally heavy. Reverted back to 2vcpu.

PR #202, opened 2025-01-23
ci: fail on warnings

nbayindirli:
AXE-7243
Screenshot 2025-01-23 at 3 12 43 PM

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -29,6 +29,6 @@ jobs:
           shared-key: "cache"
 
       - name: Run tests
-        run: cargo test --locked
+        run: RUSTFLAGS="-D warnings" cargo test --locked
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yaml
```diff
@@ -29,6 +29,6 @@ jobs:
           shared-key: "cache"
 
       - name: Run tests
-        run: cargo test --locked
+        run: RUSTFLAGS="-D warnings" cargo test --locked
```
milapsheth, line 32: Also add it for cargo build?

milapsheth (approved):

.github/workflows/test.yaml
nbayindirli: This has the same effect IIUC– we do not explicitly call `cargo build` in CI

PR #203, opened 2025-01-23
test(example): use gateway, ITS, and gas service testutils

nbayindirli:
AXE-7230

### DIFF
--- a/contracts/stellar-example/src/tests/test.rs
+++ b/contracts/stellar-example/src/tests/test.rs
@@ -34,11 +34,7 @@ struct ExampleTestConfig<'a> {
 fn setup_app<'a>(env: &Env) -> ExampleTestConfig<'a> {
     let (signers, gateway_client) = setup_gateway(env, 0, 5);
     let gas_service_client = setup_gas_service(env);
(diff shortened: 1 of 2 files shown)


contracts/stellar-example/src/tests/test.rs
```diff
@@ -1,58 +1,40 @@
...
-    let owner: Address = Address::generate(env);
-    let gas_collector: Address = Address::generate(env);
-    let gas_service = env.register(AxelarGasService, (&owner, &gas_collector));
-    let gas_service_client = AxelarGasServiceClient::new(env, &gas_service);
-
-    gas_service_client
-}
-
-struct TestConfig<'a> {
+struct ExampleTestConfig<'a> {
```
milapsheth, line 26:
```suggestion
struct TestConfig<'a> {
```
Just keep it as test config? Example prefix is implied

milapsheth (approved):

contracts/stellar-example/src/tests/test.rs
nbayindirli: Mm ok yes makes sense

PR #212, opened 2025-01-25
chore: update codecov coverage config

ahramy:


codecov.yml
```diff
@@ -4,7 +4,9 @@ coverage:
       default:
         target: auto
         threshold: 0.25%
+        informational: false
```
nbayindirli, line 7: What's the motivation here?

PR #214, opened 2025-01-25
feat(axelar-std-derive): add macro to execute when contract is not paused

milapsheth:
[AXE-7255]

stellar/rs-soroban-sdk#1426 allows function attributes to be used to check for pause status

[AXE-7255]: https://axelarnetwork.atlassian.net/browse/AXE-7255?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/stellar-axelar-gateway/src/contract.rs
+++ b/contracts/stellar-axelar-gateway/src/contract.rs
@@ -2,7 +2,9 @@ use soroban_sdk::xdr::ToXdr;
 use soroban_sdk::{contract, contractimpl, Address, Bytes, BytesN, Env, String, Vec};
 use stellar_axelar_std::events::Event;
 use stellar_axelar_std::ttl::extend_instance_ttl;
-use stellar_axelar_std::{ensure, interfaces, Operatable, Ownable, Pausable, Upgradable};
+use stellar_axelar_std::{
+    ensure, interfaces, when_not_paused, Operatable, Ownable, Pausable, Upgradable,
+};
 
 use crate::auth;
 use crate::error::ContractError;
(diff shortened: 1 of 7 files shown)


contracts/stellar-axelar-gateway/src/contract.rs
```diff
@@ -146,13 +148,12 @@ impl AxelarGatewayInterface for AxelarGateway {
         auth::previous_signers_retention(env)
     }
 
+    #[when_not_paused]
```
ahramy, line 151: 👍

packages/stellar-axelar-std/src/interfaces/pausable.rs
```diff
@@ -57,11 +57,19 @@ pub struct UnpausedEvent {}
 #[cfg(test)]
 mod test {
     use soroban_sdk::testutils::Address as _;
-    use soroban_sdk::{contract, contractimpl, Address, Env};
+    use soroban_sdk::{contract, contracterror, contractimpl, Address, Env};
 
     use super::{PausedEvent, UnpausedEvent};
```
ahramy, line 62: Optional: add account address to Event Log similar to https://github.com/axelarnetwork/axelar-gmp-sdk-solidity/blob/432449d7b330ec6edf5a8e0746644a253486ca87/contracts/interfaces/IPausable.sol#L11

ahramy (approved):

packages/stellar-axelar-std/src/interfaces/pausable.rs
milapsheth: The caller can be retrieved from the tx itself. It existed for EVM due to restrictions in the monitoring tool/Tenderly I believe, but it's a weird pattern to include the caller in here I feel

PR #216, opened 2025-01-27
feat(axelar-std-derive): add contractstorage attribute macro

nbayindirli:
AXE-7205

Example generation: https://github.com/axelarnetwork/axelar-cgp-stellar/blob/5f9c95b15b6b2791090045986cdc106bb7b32b18/packages/stellar-axelar-std-derive/src/testdata/storage_schema_generation_succeeds.golden#L1-L141


packages/stellar-axelar-std/src/tests/storage.rs
```diff
@@ -0,0 +1,209 @@
...
+        DataKey::set_counter(env, &new_counter);
+        new_counter
+    }
+
+    pub fn set_message(env: &Env, sender: Address, message: String) {
+        DataKey::set_message(env, sender, &message);
+    }
+
+    pub fn message(env: &Env, sender: Address) -> Option<String> {
+        DataKey::get_message(env, sender)
```
milapsheth, line 45: The recommended usage was changed in the task
https://axelarnetwork.atlassian.net/browse/AXE-7205

`storage::message(..)` and `storage::set_message(..)` since it's weird to expose a DataKey type that's not used directly, and potentially allows misuse. Instead, the macro generates pub methods while keeping the enum private to enforce strict usage

packages/stellar-axelar-std-derive/src/contractstorage.rs
```diff
@@ -0,0 +1,177 @@
...
+    };
+
+    let ttl_fn = match storage_attrs.storage_type {
+        StorageType::Persistent => quote! {
+            use stellar_axelar_std::ttl::extend_persistent_ttl;
+            extend_persistent_ttl(env, &key);
+        },
+        StorageType::Instance => quote! {
+            use stellar_axelar_std::ttl::extend_instance_ttl;
+            extend_instance_ttl(env);
```
milapsheth, line 138:
```suggestion
            stellar_axelar_std::ttl::extend_instance_ttl(env);
```
no need to use `use` unless a trait needs to be imported

packages/stellar-axelar-std-derive/src/lib.rs
```diff
@@ -233,3 +234,58 @@ pub fn derive_its_executable(input: TokenStream) -> TokenStream {
...
+///     #[instance]
+///     #[value(Address)]
+///     Owner,
+///
+///     #[persistent]
+///     #[value(String)]
+///     TokenName { token_id: u32 },
+///
+///     #[temporary]
+///     #[value(u64)]
```
milapsheth, line 262: We'll need to add another type of value, `#[status_value]` that sets and delete the status key, instead of storing a value (e.g. Pausable status). You can add in a separate PR, or this one whichever you think will be easier

nbayindirli: I see. Created a task for follow-up PR: https://axelarnetwork.atlassian.net/browse/AXE-7272

packages/stellar-axelar-std/src/tests/storage.rs
nbayindirli: I made all relevant enums private, however access to DataKey in the contract doesn't change (i.e. it doesn't seem like the task code example was updated?)

Let me know if this looks good now

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -134,8 +134,7 @@ fn storage_fns(
             extend_persistent_ttl(env, &key);
```
milapsheth, line 134: this should be updated to reflect the same style as below

packages/stellar-axelar-std/src/tests/storage.rs
```diff
@@ -7,58 +7,62 @@ use crate::contractstorage;
...
+    }
 }
 
 #[contractimpl]
 impl Contract {
     pub const fn __constructor() {}
 
     pub fn increment_counter(env: &Env) -> u32 {
-        let current_counter = DataKey::get_counter(env).unwrap_or(0);
+        let current_counter = storage::DataKey::get_counter(env).unwrap_or(0);
```
milapsheth, line 38:
```suggestion
        let current_counter = storage::counter(env).unwrap_or(0);
```
See my comment / recommended usage in the task again. `DataKey` shouldn't be exposed and `get` prefix isn't needed

packages/stellar-axelar-std/src/tests/storage.rs
```diff
@@ -7,58 +7,62 @@ use crate::contractstorage;
...
-    #[value(Address)]
-    LastCaller { timestamp: u64 },
-
-    #[persistent]
-    #[value(bool)]
-    Flag { key: String, owner: Address },
+mod storage {
+    use super::*;
+
+    #[contractstorage]
```
milapsheth, line 13: To ensure that the storage schema doesn't accidentally break, make the macro auto-create a unit test that does goldie::assert over the below enum definition / token stream?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+
+#[derive(Default)]
+struct StorageAttributes {
+    storage_type: StorageType,
+    value_type: Option<Type>,
+}
+
+#[derive(Default, Debug)]
+enum StorageType {
+    #[default]
```
milapsheth, line 14: A default doesn't really make sense. The type should be mentioned explicitly for clarity

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+    let setter_name = format_ident!("set_{}", variant_ident.to_string().to_snake_case());
+
+    let storage_method = match storage_attrs.storage_type {
+        StorageType::Instance => quote! { instance },
+        StorageType::Persistent => quote! { persistent },
+        StorageType::Temporary => quote! { temporary },
+    };
+
+    let ttl_fn = match storage_attrs.storage_type {
+        StorageType::Persistent => quote! {
```
milapsheth, line 132: Add an
```
impl StorageType {
    pub fn storage_method(&self) -> ...;
}
```
that implements this match. Improves encapsulation

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+            use stellar_axelar_std::ttl::extend_persistent_ttl;
```
milapsheth, line 133: same here, create a method on the type to give the appropriate ttl method

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+
+        impl #name {
+            #(#storage_fns)*
+        }
+    };
+
+    output
+}
+
+fn transform_variant(variant: &Variant) -> TokenStream {
```
milapsheth, line 51: add a docstring on what this is doing, an example would make it readable

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+            quote! {
+                #variant_name(#(#types),*)
+            }
+        }
+        _ => panic!("Only unit variants or named fields are supported in storage enums."),
+    }
+}
+
+fn storage_attributes(attrs: &[Attribute]) -> StorageAttributes {
+    let mut storage_attributes = StorageAttributes::default();
```
milapsheth, line 71: bad pattern to use mutation. separate this into two functions, one to return the storage type, and another for the value. Ensure that exactly one distinct attribute is present for storage type, and one for the value

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+            }
+            "persistent" => {
+                storage_attributes.storage_type = StorageType::Persistent;
+            }
+            "temporary" => {
+                storage_attributes.storage_type = StorageType::Temporary;
+            }
+            "value" => {
+                if let Meta::List(list) = &attr.meta {
+                    if let Ok(ty) = list.parse_args::<Type>() {
```
milapsheth, line 88: Use `list.parse_args::<Type>().ok()` to convert Result to Option type

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let (field_names, field_types) = match &variant.fields {
+        Fields::Unit => (vec![], vec![]),
+        Fields::Named(fields) => {
+            let names = fields.named.iter().map(|f| &f.ident).collect();
+            let types = fields.named.iter().map(|f| &f.ty).collect();
+            (names, types)
+        }
+        _ => panic!("Only unit variants or named fields are supported in storage enums."),
+    };
```
milapsheth, line 115: create a helper for this to reduce function complexity

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+
+    let value_type = storage_attrs
+        .value_type
+        .as_ref()
+        .expect("value type required");
```
milapsheth, line 120: Bad pattern. The StorageAttributes type shouldn't allow an invalid value to begin with. In Rust, types enforce safety / correctness, so they should be defined in a way that prevents invalid values/combinations altogether

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+    };
+
+    quote! {
+        pub fn #getter_name(#param_list) -> Option<#value_type> {
+            let key = #key_construction;
+            let value = env.storage()
+                .#storage_method()
+                .get(&key);
+
+            if value.is_some() { #ttl_fn }
```
milapsheth, line 161:
```suggestion
            if value.is_some() {
                #ttl_fn
            }
```
generated code looks a bit nicer if someone takes a look

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+            use stellar_axelar_std::ttl::extend_persistent_ttl;
+            extend_persistent_ttl(env, &key);
+        },
+        StorageType::Instance => quote! {
+            stellar_axelar_std::ttl::extend_instance_ttl(env);
+        },
+        _ => quote! {},
+    };
+
+    let key_construction = if field_names.is_empty() {
```
milapsheth, line 142:
```suggestion
    let key = if field_names.is_empty() {
```
perhaps

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+            let value = env.storage()
+                .#storage_method()
+                .get(&key);
+
+            if value.is_some() { #ttl_fn }
+
+            value
+        }
+
+        pub fn #setter_name(#param_list, value: &#value_type) {
```
milapsheth, line 166: Add a `delete_*` method as well for convenience. It'll be useful for migrations

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,176 @@
...
+    } else {
+        quote! { env: &soroban_sdk::Env, #(#field_names: #field_types),* }
+    };
+
+    quote! {
+        pub fn #getter_name(#param_list) -> Option<#value_type> {
+            let key = #key_construction;
+            let value = env.storage()
+                .#storage_method()
+                .get(&key);
```
milapsheth, line 159:
```suggestion
                .get::<_, #value_type>(&key);
```
auto-inference might not work for `Option<Option<...>>` case

packages/stellar-axelar-std/src/tests/storage.rs
```diff
@@ -0,0 +1,213 @@
...
+        #[persistent]
+        #[value(String)]
+        Message { sender: Address },
+
+        #[temporary]
+        #[value(Address)]
+        LastCaller { timestamp: u64 },
+
+        #[persistent]
+        #[value(bool)]
```
milapsheth, line 28: Add a test case for `Option<..>` value type

contracts/stellar-example/src/contract.rs
```diff
@@ -64,10 +61,7 @@ impl CustomInterchainTokenExecutable for Example {
     type Error = ExampleError;
 
     fn __interchain_token_service(env: &Env) -> Address {
-        env.storage()
-            .instance()
-            .get(&DataKey::InterchainTokenService)
-            .expect("ITS not found")
+        DataKey::get_interchain_token_service(env).expect("ITS not found")
```
ahramy, line 64: I think we use lower case for most of `.expect` messages, can you double check?
```suggestion
        DataKey::get_interchain_token_service(env).expect("interchain token service not found")
```

packages/stellar-axelar-std/src/tests/storage.rs
```diff
@@ -0,0 +1,213 @@
...
+    client.set_message(&sender2, &message2);
+    client.set_message(&sender3, &message3);
+
+    assert_eq!(client.message(&sender1), Some(message1));
+    assert_eq!(client.message(&sender2), Some(message2));
+    assert_eq!(client.message(&sender3), Some(message3));
+}
+
+#[test]
+fn instance_storage_overwrite_succeeds() {
```
ahramy, line 150: This does same as `instance_storage_succeeds` but only the function name is different

ahramy:
Since we’ve decided to merge this PR, we need to ensure the PR title is appropriate. Using feat will bump up the minor version to 0.x.0 of the package. Do we want to do that?

packages/stellar-axelar-std/src/tests/storage.rs
nbayindirli: Sure I suppose unnecessary since the logic is the same. Could update to increment multiple more times but will remove `instance_storage_overwrite_succeeds` in favor

nbayindirli:
> Since we’ve decided to merge this PR, we need to ensure the PR title is appropriate. Using feat will bump up the minor version to 0.x.0 of the package. Do we want to do that?

This is a new feature, but it does not actually change things functionally, so a major version does not make sense.

IMO this is a good example of why we should not _infer_ semver and that bumps should be manual by default

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,415 @@
...
+            "persistent" => return StorageType::Persistent,
+            "temporary" => return StorageType::Temporary,
+            "value" => continue,
+            unknown => panic!("Unknown storage attribute: {}", unknown),
+        }
+    }
+
+    panic!(
+        "Storage type must be specified exactly once as 'instance', 'persistent', or 'temporary'."
+    )
```
cgorenflo, line 150:
```suggestion
    attrs.iter().flat_map(|attr| match attr {
        _ if attr.path().is_ident("instance") => Some(StorageType::Instance),
        _ if attr.path().is_ident("persistent") => Some(StorageType::Persistent),
        _ if attr.path().is_ident("temporary") => Some(StorageType::Temporary),
        _ => None})
    .exactly_one()
    .expect("Storage type must be specified exactly once as 'instance', 'persistent', or 'temporary'.")
```

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,415 @@
...
+                return list
+                    .parse_args::<Type>()
+                    .expect("Failed to parse value type.");
+            } else {
+                panic!("Value attribute must contain a type parameter: #[value(Type)]");
+            }
+        }
+    }
+
+    panic!("Missing required #[value(Type)] attribute.")
```
cgorenflo, line 168: see the other suggestion about restructuring this

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,415 @@
...
+
+    let param_list = if field_names.is_empty() {
+        quote! { env: &soroban_sdk::Env }
+    } else {
+        quote! { env: &soroban_sdk::Env, #(#field_names: #field_types),* }
+    };
+
+    quote! {
+        pub fn #getter_name(#param_list) -> Option<#value_type> {
+            let key = #key;
```
cgorenflo, line 202: you could inline the key (the compiler will do that anyway, so no big deal)

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,415 @@
...
+
+            if value.is_some() {
+                #ttl_fn
+            }
+
+            value
+        }
+
+        pub fn #setter_name(#param_list, value: &#value_type) {
+            let key = #key;
```
cgorenflo, line 215: you could inline the key (the compiler will do that anyway, so no big deal)

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,415 @@
...
+
+            env.storage()
+                .#storage_method()
+                .set(&key, value);
+
+            #ttl_fn
+        }
+
+        pub fn #deleter_name(#param_list) {
+            let key = #key;
```
cgorenflo, line 225: you could inline the key (the compiler will do that anyway, so no big deal)

packages/stellar-axelar-std-derive/src/lib.rs
```diff
@@ -233,3 +234,58 @@ pub fn derive_its_executable(input: TokenStream) -> TokenStream {
...
+///     }
+///
+///     pub fn token_name(env: &Env, token_id: u32) -> Option<String> {
+///         // Generates: DataKey::get_token_name(env, token_id)
+///         DataKey::get_token_name(env, token_id)
+///     }
+/// }
+/// # }
+/// ```
+#[proc_macro_attribute]
```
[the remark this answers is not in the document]
cgorenflo: I assume it's not so that `contractstorage` can better align with `contractimpl` etc

packages/stellar-axelar-std-derive/src/storage.rs
cgorenflo: `exactly_one` is part of the itertools crate

packages/stellar-axelar-std-derive/Cargo.toml
```diff
@@ -12,10 +12,13 @@ proc-macro = true
 [features]
 
 [dependencies]
+goldie = { workspace = true }
```
milapsheth, line 15: `goldie` should be a dev dependency

packages/stellar-axelar-std-derive/src/lib.rs
```diff
@@ -233,3 +234,58 @@ pub fn derive_its_executable(input: TokenStream) -> TokenStream {
...
+///         token_id: u32,
+///         name: String,
+///     ) {
+///         // Generates: DataKey::set_token_name(env, token_id, &name);
+///         DataKey::set_token_name(env, token_id, &name);
+///     }
+///
+///     pub fn token_name(env: &Env, token_id: u32) -> Option<String> {
+///         // Generates: DataKey::get_token_name(env, token_id)
+///         DataKey::get_token_name(env, token_id)
```
milapsheth, line 282: outdated usage code

packages/stellar-axelar-std-derive/src/testdata/storage_schema_generation_succeeds.golden
```diff
@@ -0,0 +1,141 @@
...
+pub fn set_flag(env: &soroban_sdk::Env, key: String, owner: Address, value: &bool) {
+    DataKey::set_flag(env, key, owner, value)
+}
+pub fn delete_flag(env: &soroban_sdk::Env, key: String, owner: Address) {
+    DataKey::delete_flag(env, key, owner)
+}
+pub fn optional_message(env: &soroban_sdk::Env, id: u32) -> Option<Option<String>> {
+    DataKey::optional_message(env, id)
+}
+pub fn set_optional_message(env: &soroban_sdk::Env, id: u32, value: &Option<String>) {
```
milapsheth, line 136: add newline between functions

packages/stellar-axelar-std-derive/src/testdata/storage_schema_generation_succeeds.golden
```diff
@@ -0,0 +1,141 @@
+#[contracttype]
+enum DataKey {
+    Counter,
+    Message(Address),
+    LastCaller(u64),
+    Flag(String, Address),
+    OptionalMessage(u32),
+}
+impl DataKey {
+    pub fn counter(env: &soroban_sdk::Env) -> Option<u32> {
```
milapsheth, line 10: it's unnecessary to implement these methods on DataKey since they're never used directly on it. We can just generate the root `counter` getter with this implementation

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+        .iter()
+        .map(|variant| {
+            let storage_type = storage_type(&variant.attrs);
+            let value_type = value_type(&variant.attrs);
+            storage_fns(
+                name,
+                variant,
+                &StorageAttributes {
+                    storage_type,
+                    value_type,
```
milapsheth, line 61:
```
storage_type: storage_type(&variant.attrs),
value_type: value_type(&variant.attrs)
```
no need to create local vars

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+                },
+            )
+        })
+        .collect();
+
+    let public_fns: Vec<_> = variants
+        .iter()
+        .map(|variant| {
+            let value_type = value_type(&variant.attrs);
+            public_storage_fns(name, variant, &value_type)
```
milapsheth, line 71: same

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+
+    let output = quote! {
+        #[contracttype]
+        enum #name {
+            #(#transformed_variants,)*
+        }
+
+        impl #name {
+            #(#storage_fns)*
+        }
```
milapsheth, line 83: not needed, just implement the public fns directly since the enum is never exposed

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+/// }
+/// ```
+///
+/// /* Generated */
+/// #[contracttype]
+/// pub enum DataKey {
+///     Gateway, // Unit variant (storage key)
+///     Users(Address), // Tuple variant (storage key map)
+/// }
+/// ```
```
milapsheth, line 117: no opening backticks for this one

contracts/stellar-example/src/storage.rs
```diff
@@ -0,0 +1,18 @@
+use soroban_sdk::{contracttype, Address};
+use stellar_axelar_std::contractstorage;
+
+#[contractstorage]
+#[derive(Clone, Debug)]
+enum DataKey {
+    #[instance]
+    #[value(Address)]
```
milapsheth, line 8: Let's add a `#[status_value]` type in a follow-up PR that sets `()` as the value, and creates 
 `set_*`, `is_*`, `remove_*` methods corresponding to it, e.g. Pausable interface

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+
+                #[persistent]
+                #[value(Option<String>)]
+                OptionalMessage { id: u32 },
+            }
+        };
+
+        let generated = contractstorage(&input);
+        let file: syn::File = syn::parse2(generated).unwrap();
+        let formatted = prettyplease::unparse(&file);
```
milapsheth, line 319: prettyplease should be a dev dependency as well

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+        enum #name {
+            #(#transformed_variants,)*
+        }
+
+        impl #name {
+            #(#storage_fns)*
+        }
+
+        #(#public_fns)*
+    };
```
milapsheth, line 86: I'd like to see an auto-generated test block for every storage type, so that any changes to the actual DataKey enum are flagged, since a migration would likely be needed
```rust
#[cfg(test)]
mod test {
    #[test]
    fn ensure_storage_schema_is_unchanged() {
    }
}
```

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -0,0 +1,400 @@
...
+        })
+        .exactly_one()
+        .expect("Missing required #[value(Type)] attribute.")
+}
+
+/// Generates the storage getter, setter, and deleter functions for a storage enum variant.
+fn storage_fns(
+    enum_name: &Ident,
+    variant: &Variant,
+    storage_attrs: &StorageAttributes,
```
milapsheth, line 174: you can destructure in the arg itself, `StorageAttributes { storage_types, value_types}: StorageAttributes`

packages/stellar-axelar-std-derive/src/testdata/storage_schema_generation_succeeds.golden
nbayindirli: AFAICT a newline between each scope is not possible. They're recognized intra-function, but adding newlines between fns within the quote did not have an effect

contracts/stellar-example/src/storage.rs
nbayindirli: https://github.com/axelarnetwork/axelar-cgp-stellar/pull/234

packages/stellar-axelar-std-derive/src/lib.rs
nbayindirli: https://github.com/axelarnetwork/axelar-cgp-stellar/pull/234

packages/stellar-axelar-std-derive/src/storage.rs
nbayindirli: I'm not example sure why but this fails build if set as a dev-dep

PR #231, opened 2025-01-29
chore: initial v1.0.0 release

ahramy:


contracts/stellar-example/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "stellar-example"
-version = "0.1.0"
+version = "1.0.0"
 edition = { workspace = true }
```
ahramy, line 4: Note that we do not release example contract: https://github.com/axelarnetwork/axelar-cgp-stellar/blob/main/.release-plz.toml#L31

PR #237, opened 2025-02-03
refactor(axelar-gateway): use contractstorage

nbayindirli:
AXE-7283


contracts/stellar-axelar-gateway/src/storage.rs
```diff
@@ -15,17 +16,42 @@ pub enum MessageApprovalValue {
...
-#[contracttype]
+#[contractstorage]
 #[derive(Clone, Debug)]
 pub enum DataKey {
     /// Gateway
-    MessageApproval(MessageApprovalKey),
+
+    #[persistent]
+    #[value(MessageApprovalValue)]
+    MessageApproval { key: MessageApprovalKey },
```
milapsheth, line 26: Create a ticket to remove this wrapper key type altogether, we can use the fields directly. This PR should preserve the existing schema but we can make it cleaner in a follow up with a potential migration

milapsheth, line 26:
```suggestion
    MessageApproval { approval_key: MessageApprovalKey },
```

contracts/stellar-axelar-gateway/src/auth.rs
```diff
@@ -39,24 +29,15 @@ pub fn initialize_auth(
 }
 
 pub fn domain_separator(env: &Env) -> BytesN<32> {
-    env.storage()
-        .instance()
-        .get(&DataKey::DomainSeparator)
-        .expect("domain_separator not found")
+    storage::domain_separator(env).expect("domain_separator not found")
```
milapsheth, line 32: We should consider creating two getter versions, one that requires the key to exist and another that returns an Option. The former can panic if it's not found, and will be the more commonly used. The panic msg is ignored so we can use unwrap instead when we refactor

contracts/stellar-axelar-gateway/src/auth.rs
```diff
@@ -158,11 +125,7 @@ fn message_hash_to_sign(env: &Env, signers_hash: BytesN<32>, data_hash: &BytesN<
 }
 
 fn update_rotation_timestamp(env: &Env, enforce_rotation_delay: bool) -> Result<(), ContractError> {
-    let last_rotation_timestamp: u64 = env
-        .storage()
-        .instance()
-        .get(&DataKey::LastRotationTimestamp)
-        .unwrap_or(0);
+    let last_rotation_timestamp: u64 = storage::last_rotation_timestamp(env).unwrap_or(0);
```
milapsheth, line 128:
```suggestion
    let last_rotation_timestamp = storage::last_rotation_timestamp(env).unwrap_or(0);
```
Type hint shouldnt be needed now

milapsheth, line 128: Actually, the rotation in the constructor should ensure that this is never unset, so we can remove the unwrap_or fallback

contracts/stellar-axelar-gateway/src/contract.rs
```diff
@@ -251,10 +249,7 @@ impl AxelarGateway {
 
     /// Get the message approval value by key, defaulting to `MessageNotApproved`
     fn message_approval_by_key(env: &Env, key: MessageApprovalKey) -> MessageApprovalValue {
```
milapsheth, line 251: Is this method even needed anymore with the simplification?

contracts/stellar-axelar-gateway/src/storage.rs
nbayindirli:
```suggestion
    MessageApproval { message_approval_key: MessageApprovalKey },
```

PR #239, opened 2025-02-03
refactor: move the run_migration function into a clearly defined interface

cgorenflo:
[AXE-7361](https://axelarnetwork.atlassian.net/browse/AXE-7361)

[AXE-7361]: https://axelarnetwork.atlassian.net/browse/AXE-7361?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/stellar-axelar-operators/src/contract.rs
```diff
@@ -87,7 +88,7 @@ impl AxelarOperatorsInterface for AxelarOperators {
     }
 }
 
-impl AxelarOperators {
-    // Modify this function to add migration logic
-    const fn run_migration(_env: &Env, _migration_data: ()) {}
+impl CustomMigratableInterface for AxelarOperators {
+    type MigrationData = ();
+    fn __migrate(_env: &Env, _migration_data: ()) {}
```
ahramy, line 93: it's different than others - should it be the same?
```suggestion
    fn __migrate(_env: &Env, _migration_data: Self::MigrationData) {}
```

ahramy, line 93: Could we add unit test coverage?

packages/stellar-axelar-std/src/interfaces/upgradable.rs
```diff
@@ -16,16 +16,22 @@ pub trait UpgradableInterface: OwnableInterface {
...
     /// Migrates contract state after upgrading the contract code.
     fn migrate(env: &Env, migration_data: Self::MigrationData) -> Result<(), Self::Error>;
 }
 
+pub trait CustomMigratableInterface: UpgradableInterface {
+    /// Data needed during the migration. Each contract can define its own data type.
+    type MigrationData: FromVal<Env, Val>;
+
+    /// Migrates contract state after upgrading the contract code.
+    fn __migrate(env: &Env, migration_data: Self::MigrationData);
```
milapsheth, line 32:
```suggestion
    /// Data needed during the migration. Each contract can define its own data type.
    type MigrationData: FromVal<Env, Val> = ();

    /// Migrates contract state after upgrading the contract code.
    fn __migrate(env: &Env, migration_data: Self::MigrationData) {}
```
Let's have a default implementation that can be overridden when needed?

contracts/stellar-axelar-operators/src/contract.rs
cgorenflo: yes thanks, I knew I did that somewhere but just couldn't find it

packages/stellar-axelar-std/src/interfaces/upgradable.rs
cgorenflo: yeah good idea

ahramy (approved):

PR #241, opened 2025-02-03
fix(axelar-std-derive): incorporate contractstorage feedback

nbayindirli:
AXE-7359

### DIFF
--- a/contracts/stellar-example/Cargo.toml
+++ b/contracts/stellar-example/Cargo.toml
@@ -21,12 +21,15 @@ stellar-interchain-token-service = { workspace = true, features = ["library"] }
 
 [dev-dependencies]
 goldie = { workspace = true }
+prettyplease = { workspace = true }
+quote = { workspace = true }
 soroban-sdk = { workspace = true, features = ["testutils"] }
 soroban-token-sdk = { workspace = true }
 stellar-axelar-gas-service = { workspace = true, features = ["testutils"] }
 stellar-axelar-gateway = { workspace = true, features = ["testutils"] }
 stellar-axelar-std = { workspace = true, features = ["testutils"] }
 stellar-interchain-token-service = { workspace = true, features = ["testutils"] }
+syn = { workspace = true }
 
 [lints]
 workspace = true
--- a/contracts/stellar-example/src/testdata/ensure_data_key_storage_schema_is_unchanged.golden
+++ b/contracts/stellar-example/src/testdata/ensure_data_key_storage_schema_is_unchanged.golden
@@ -1,6 +1,15 @@
-#[derive(Clone, Debug)] enum DataKey
-{
-    #[instance] #[value(Address)] Gateway, #[instance] #[value(Address)]
-    GasService, #[instance] #[value(Address)] InterchainTokenService,
-    #[instance] #[status] Paused,
+#[derive(Clone, Debug)]
+enum DataKey {
+    #[instance]
+    #[value(Address)]
+    Gateway,
+    #[instance]
+    #[value(Address)]
+    GasService,
+    #[instance]
+    #[value(Address)]
+    InterchainTokenService,
+    #[instance]
+    #[status]
+    Paused,
 }
--- a/packages/stellar-axelar-std-derive/src/storage.rs
+++ b/packages/stellar-axelar-std-derive/src/storage.rs
@@ -9,6 +9,24 @@ enum Value {
     Type(Type),
 }
 
+impl Value {
+    fn fn_names(&self, variant_ident: &Ident) -> (Ident, Ident, Ident) {
+        let ident = variant_ident.to_string().to_snake_case();
+        match self {
+            Value::Status => (
+                format_ident!("is_{}", ident),
+                format_ident!("set_{}_status", ident),
+                format_ident!("remove_{}_status", ident),
+            ),
+            Value::Type(_) => (
+                format_ident!("{}", ident),
+                format_ident!("set_{}", ident),
+                format_ident!("remove_{}", ident),
+            ),
+        }
+    }
+}
+
 struct StorageAttributes {
     storage_type: StorageType,
     value: Value,
@@ -68,13 +86,11 @@ pub fn contract_storage(input: &DeriveInput) -> TokenStream {
         .collect();
 
     let contract_storage = quote! {
-        #[doc = "\n* Storage Enum\n"]
         #[contracttype]
         enum #name {
             #(#transformed_variants,)*
         }
 
-        #[doc = "\n* Public Functions\n"]
         #(#public_fns)*
     };
 
@@ -83,7 +99,6 @@ pub fn contract_storage(input: &DeriveInput) -> TokenStream {
     quote! {
         #contract_storage
 
-        #[doc = "\n* Contract Storage Tests\n"]
         #contract_storage_tests
     }
 }
@@ -198,42 +213,50 @@ fn public_storage_fns(
     };
 
     match &value {
-        Value::Status => value_status_fns(variant, &param_list, &storage_method, key),
+        Value::Status => value_status_fns(
+            &param_list,
+            &storage_method,
+            key,
+            &ttl_fn,
+            value.fn_names(variant_ident),
+        ),
         Value::Type(value_type) => value_type_fns(
-            variant,
             &param_list,
             &storage_method,
             key,
             value_type,
             &ttl_fn,
+            value.fn_names(variant_ident),
         ),
     }
 }
 
 fn value_status_fns(
-    variant: &Variant,
     param_list: &TokenStream,
     storage_method: &TokenStream,
     key: TokenStream,
+    ttl_fn: &TokenStream,
+    (getter_name, setter_name, remover_name): (Ident, Ident, Ident),
 ) -> TokenStream {
-    let (getter_name, setter_name, remover_name) = fn_names(variant, true);
-
     quote! {
-        #[doc = " Status Getter"]
         pub fn #getter_name(#param_list) -> bool {
-            env.storage()
+            let value = env.storage()
                 .#storage_method()
-                .has(&#key)
+                .has(&#key);
+
+            if value {
+                #ttl_fn
+            }
+
+            value
         }
 
-        #[doc = " Status Setter"]
         pub fn #setter_name(#param_list) {
             env.storage()
                 .#storage_method()
                 .set(&#key, &());
         }
 
-        #[doc = " Status Remover"]
         pub fn #remover_name(#param_list) {
             env.storage()
                 .#storage_method()
@@ -243,17 +266,14 @@ fn value_status_fns(
 }
 
 fn value_type_fns(
-    variant: &Variant,
     param_list: &TokenStream,
     storage_method: &TokenStream,
     key: TokenStream,
     value_type: &Type,
     ttl_fn: &TokenStream,
+    (getter_name, setter_name, remover_name): (Ident, Ident, Ident),
 ) -> TokenStream {
-    let (getter_name, setter_name, remover_name) = fn_names(variant, false);
-
     quote! {
-        #[doc = " Value Type Getter"]
         pub fn #getter_name(#param_list) -> Option<#value_type> {
             let key = #key;
             let value = env.storage()
@@ -267,7 +287,6 @@ fn value_type_fns(
             value
         }
 
-        #[doc = " Value Type Setter"]
         pub fn #setter_name(#param_list, value: &#value_type) {
             let key = #key;
 
@@ -278,7 +297,6 @@ fn value_type_fns(
             #ttl_fn
         }
 
-        #[doc = " Value Type Remover"]
         pub fn #remover_name(#param_list) {
             env.storage()
                 .#storage_method()
@@ -300,25 +318,6 @@ fn fields_data(fields: &Fields) -> (Vec<&Option<Ident>>, Vec<&Type>) {
     }
 }
 
-fn fn_names(variant: &Variant, status: bool) -> (Ident, Ident, Ident) {
-    if status {
-        (
-            format_ident!("is_{}", variant.ident.to_string().to_snake_case()),
-            format_ident!("set_{}_status", variant.ident.to_string().to_snake_case()),
-            format_ident!(
-                "remove_{}_status",
-                variant.ident.to_string().to_snake_case()
-            ),
-        )
-    } else {
-        (
-            format_ident!("{}", variant.ident.to_string().to_snake_case()),
-            format_ident!("set_{}", variant.ident.to_string().to_snake_case()),
-            format_ident!("remove_{}", variant.ident.to_string().to_snake_case()),
-        )
-    }
-}
-
 fn contract_storage_tests(enum_name: &Ident, input: &DeriveInput) -> TokenStream {
     let test_module_name = format_ident!(
         "{}_storage_layout_tests",
@@ -333,9 +332,21 @@ fn contract_storage_tests(enum_name: &Ident, input: &DeriveInput) -> TokenStream
     quote! {
         #[cfg(test)]
         mod #test_module_name {
+            use super::*;
+            use prettyplease;
+            use quote::quote;
+            use syn::DeriveInput;
+
             #[test]
             fn #test_name() {
-                goldie::assert!(stringify!(#input));
+                let input = quote! {
+                    #input
+                };
+                let file: syn::File = syn::parse2(input).unwrap();
+                let formatted = prettyplease::unparse(&file)
+                    .replace("pub fn ", "\npub fn ")
+                    .replace("#[cfg(test)]", "\n#[cfg(test)]");
+                goldie::assert!(formatted);
             }
         }
     }
@@ -382,7 +393,9 @@ mod tests {
 
         let generated = contract_storage(&input);
         let file: syn::File = syn::parse2(generated).unwrap();
-        let formatted = prettyplease::unparse(&file);
+        let formatted = prettyplease::unparse(&file)
+            .replace("pub fn ", "\npub fn ")
+            .replace("#[cfg(test)]", "\n#[cfg(test)]");
         goldie::assert!(formatted);
     }
 
--- a/packages/stellar-axelar-std-derive/src/testdata/storage_schema_generation_succeeds.golden
+++ b/packages/stellar-axelar-std-derive/src/testdata/storage_schema_generation_succeeds.golden
@@ -1,6 +1,3 @@
-/**
-* Storage Enum
-*/
 #[contracttype]
 enum DataKey {
     Counter,
@@ -11,10 +8,7 @@ enum DataKey {
     Initialized,
     Paused,
 }
-/**
-* Public Functions
-*/
-/// Value Type Getter
+
 pub fn counter(env: &soroban_sdk::Env) -> Option<u32> {
     let key = DataKey::Counter;
     let value = env.storage().instance().get::<_, u32>(&key);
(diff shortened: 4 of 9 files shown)


packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -198,42 +213,50 @@ fn public_storage_fns(
     };
 
     match &value {
-        Value::Status => value_status_fns(variant, &param_list, &storage_method, key),
+        Value::Status => value_status_fns(
```
milapsheth, line 216: move this to an `impl Value` block. `storage_type` can be passed instead of storage_method and ttl separately

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -198,42 +213,50 @@ fn public_storage_fns(
...
+            &param_list,
+            &storage_method,
+            key,
+            &ttl_fn,
```
milapsheth, line 220: Looks like the `&key` is implicitly captured in `ttl_method` implementation which is not good. key should be passed in to `ttl_method` to get the `ttl_fn`

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -198,42 +213,50 @@ fn public_storage_fns(
     };
 
     match &value {
-        Value::Status => value_status_fns(variant, &param_list, &storage_method, key),
+        Value::Status => value_status_fns(
+            &param_list,
+            &storage_method,
+            key,
```
milapsheth, line 219: key and param_list can be constructed within this impl method as well (and potentially in their own helper method)

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -333,9 +332,21 @@ fn contract_storage_tests(enum_name: &Ident, input: &DeriveInput) -> TokenStream
     quote! {
         #[cfg(test)]
         mod #test_module_name {
+            use super::*;
```
milapsheth, line 335: import what you need instead of `super::*`?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -333,9 +332,21 @@ fn contract_storage_tests(enum_name: &Ident, input: &DeriveInput) -> TokenStream
...
             #[test]
             fn #test_name() {
-                goldie::assert!(stringify!(#input));
+                let input = quote! {
+                    #input
+                };
+                let file: syn::File = syn::parse2(input).unwrap();
+                let formatted = prettyplease::unparse(&file)
+                    .replace("pub fn ", "\npub fn ")
+                    .replace("#[cfg(test)]", "\n#[cfg(test)]");
```
milapsheth, line 348: why add these replace calls? The input is just an enum

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -333,9 +332,21 @@ fn contract_storage_tests(enum_name: &Ident, input: &DeriveInput) -> TokenStream
...
+            use syn::DeriveInput;
+
             #[test]
             fn #test_name() {
-                goldie::assert!(stringify!(#input));
+                let input = quote! {
+                    #input
+                };
+                let file: syn::File = syn::parse2(input).unwrap();
+                let formatted = prettyplease::unparse(&file)
```
milapsheth, line 346:
```suggestion
                let formatted_enum = prettyplease::unparse(&file)
```

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -382,7 +393,9 @@ mod tests {
 
         let generated = contract_storage(&input);
         let file: syn::File = syn::parse2(generated).unwrap();
-        let formatted = prettyplease::unparse(&file);
+        let formatted = prettyplease::unparse(&file)
```
milapsheth, line 396:
```suggestion
        let formatted_storage_module = prettyplease::unparse(&file)
```

contracts/stellar-example/src/testdata/ensure_data_key_storage_schema_is_unchanged.golden
```diff
@@ -1,6 +1,15 @@
...
-    GasService, #[instance] #[value(Address)] InterchainTokenService,
-    #[instance] #[status] Paused,
-}
\ No newline at end of file
+#[derive(Clone, Debug)]
+enum DataKey {
+    #[instance]
+    #[value(Address)]
+    Gateway,
+    #[instance]
```
milapsheth, line 6:
```suggestion

    #[instance]
```
add another newline between variants?

packages/stellar-axelar-std-derive/src/storage.rs
nbayindirli: They're unnecessary. This was copied over from below. Replaced these with
```suggestion
                    .replace("#[instance]", "\n\t#[instance]")
                    .replace("#[persistent]", "\n\t#[persistent]")
                    .replace("#[temporary]", "\n\t#[temporary]");
```

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +9,158 @@ enum Value {
...
+                format_ident!("is_{}", ident),
+                format_ident!("set_{}_status", ident),
+                format_ident!("remove_{}_status", ident),
+                None,
+            ),
+            Self::Type(_) => (
+                format_ident!("{}", ident),
+                format_ident!("set_{}", ident),
+                format_ident!("remove_{}", ident),
+                Some(format_ident!("{}_required", ident)),
```
milapsheth, line 134:
```suggestion
                format_ident!("{}", ident),
                format_ident!("set_{}", ident),
                format_ident!("remove_{}", ident),
                Some(format_ident!("try_{}", ident)),
```

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +9,158 @@ enum Value {
...
+
+    /// Returns the getter, setter, and remover functions for a storage enum variant.
+    fn fns(&self, enum_name: &Ident, storage_type: &StorageType, variant: &Variant) -> TokenStream {
+        let (getter_name, setter_name, remover_name, required_getter_name) =
+            self.fn_names(&variant.ident);
+        let param_list = Self::param_list(variant);
+        let key = Self::key(enum_name, variant);
+
+        let storage_method = storage_type.storage_method();
+        let ttl_fn = storage_type.ttl_method(&key);
```
milapsheth, line 51: Having to clone the `key` because of duplicate references isn't great. Let's revert to the old approach instead of defining a local `key` variable. However, instead of implicitly capturing the `key` variable, pass the variable name explicitly to `ttl_method`?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +9,158 @@ enum Value {
...
+            }
+        }
+    }
+
+    /// Returns the getter, setter, and remover functions for a storage enum variant.
+    fn fns(&self, enum_name: &Ident, storage_type: &StorageType, variant: &Variant) -> TokenStream {
+        let (getter_name, setter_name, remover_name, required_getter_name) =
+            self.fn_names(&variant.ident);
+        let param_list = Self::param_list(variant);
+        let key = Self::key(enum_name, variant);
```
milapsheth, line 48:
```suggestion
        let key = Self::storage_key(enum_name, variant);
```
to reduce confusion

contracts/stellar-example/Cargo.toml
```diff
@@ -21,12 +21,15 @@ stellar-interchain-token-service = { workspace = true, features = ["library"] }
 
 [dev-dependencies]
 goldie = { workspace = true }
+prettyplease = { workspace = true }
```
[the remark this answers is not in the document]
nbayindirli: No longer needed with formatting outside of quote, above

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +9,158 @@ enum Value {
...
+    /// Returns the key for a storage enum variant.
+    fn key(enum_name: &Ident, variant: &Variant) -> TokenStream {
+        let (field_names, _) = fields_data(&variant.fields);
+        let variant_ident = &variant.ident;
+
+        if field_names.is_empty() {
+            quote! { #enum_name::#variant_ident }
+        } else {
+            let field_names = field_names.iter().map(|name| {
+                quote! { #name.clone() }
```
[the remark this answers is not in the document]
nbayindirli: Address and String don't implement Copy so not sure we can do this

packages/stellar-axelar-std-derive/src/storage.rs
nbayindirli: Discussed offline– under status or value, each of persistent/instance/temp have their own bundle of pub fns now

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +12,120 @@ enum Value {
...
 
-struct StorageAttributes {
-    storage_type: StorageType,
-    value: Value,
+impl TryFrom<&[Attribute]> for Value {
+    type Error = String;
+
+    fn try_from(attrs: &[Attribute]) -> Result<Self, Self::Error> {
+        let has_status = attrs.iter().any(|attr| attr.path().is_ident("status"));
+        let value_attrs: Vec<_> = attrs
```
cgorenflo, line 20: I still prefer my proposed approach from the previous PR for this https://github.com/axelarnetwork/axelar-cgp-stellar/pull/234#discussion_r1938030463. By splitting the computation into two (has_status and value_attrs), your checks become more complicated and seemingly couple unrelated cases. Compare "if it's a status attribute match to x, if it's a correctly formed value attribute match to y, make sure you got exactly one such match" to the current "if it's a status attribute and there are no other attributes match to x, if it's not a status attribute but a correctly formed value attribute match to y, fail if neither of these cases is true". 
These checks are not logically wrong in their end result, but why can you match a status attribute to the appropriate value only if there are no other matching attributes? To me, the concerns of figuring out what the data type should be and making sure there are no conflicting declarations are two separate concerns.
Additionally, the current implementation only works for additional attributes that also decorate the variant because you filter beforehand.
So you're iterating three times over the attributes to find the value, instead of at most once with a short circuit as soon as you find a second match.
In the end, I won't die on this hill though, it's logically correct and a user won't come into contact with it.

nbayindirli: This was very clear, and leads me to prefer the method you outlined previously using `exactly_one`. I've now updated the implementation to be similar to the following:
```rust
fn try_from(attrs: &[Attribute]) -> Result<Self, Self::Error> {
        let attr = attrs
            .iter()
            .filter(|attr| attr.path().is_ident("status") || attr.path().is_ident("value"))
            .exactly_one()
            .map_err(|_| "exactly one of #[status] and #[value(Type)] must be provided".to_string())?;

        if let Meta::List(list) = &attr.meta {
            if attr.path().is_ident("value") {
                Ok(Self::Type(
                    list.parse_args::<Type>()
                        .map_err(|_| "failed to parse value type".to_string())?,
                ))
            } else {
                Err("status attribute cannot have parameters".to_string())
            }
        } else if attr.path().is_ident("status") {
            Ok(Self::Status)
        } else {
            Err("value attribute must contain a type parameter: #[value(Type)]".to_string())
        }
    }
```

cgorenflo (approved):

packages/stellar-axelar-std-derive/src/testdata/storage_schema_generation_succeeds.golden
```diff
@@ -60,45 +45,66 @@ pub fn message_required(env: &soroban_sdk::Env, sender: Address) -> String {
...
 
+pub fn try_message(env: &soroban_sdk::Env, sender: Address) -> Option<String> {
+    let value = env
+        .storage()
+        .persistent()
+        .get::<_, String>(&DataKey::Message(sender.clone()));
+    if value.is_some() {
+        stellar_axelar_std::ttl::extend_persistent_ttl(
+            env,
+            &DataKey::Message(sender.clone()),
```
milapsheth, line 56: Cloning twice still

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +12,122 @@ enum Value {
...
+            })?;
+
+        if let Meta::List(list) = &attr.meta {
+            if attr.path().is_ident("value") {
+                Ok(Self::Type(
+                    list.parse_args::<Type>()
+                        .map_err(|_| "failed to parse value type".to_string())?,
+                ))
+            } else {
+                Err("status attribute cannot have parameters".to_string())
```
milapsheth, line 34:
```suggestion
                Err("status attribute cannot have parameters".into())
```
auto-convert instead?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +12,122 @@ enum Value {
...
+            }
+        } else if attr.path().is_ident("status") {
```
milapsheth, line 36: if checks are a bit confusing. Check whether path is a value or status first and nest inside them

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +12,122 @@ enum Value {
...
+        let attr = attrs
+            .iter()
+            .filter(|attr| attr.path().is_ident("status") || attr.path().is_ident("value"))
+            .exactly_one()
+            .map_err(|_| {
+                "exactly one of #[status] and #[value(Type)] must be provided".to_string()
+            })?;
+
+        if let Meta::List(list) = &attr.meta {
+            if attr.path().is_ident("value") {
```
milapsheth, line 28: check the ident first and nest the if let check

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +12,122 @@ enum Value {
...
+        if field_names.is_empty() {
+            quote! { env: &soroban_sdk::Env }
+        } else {
+            quote! { env: &soroban_sdk::Env, #(#field_names: #field_types),* }
+        }
+    }
+
+    /// Returns the key for a storage enum variant.
+    fn storage_key(&self, enum_name: &Ident) -> TokenStream {
+        let (field_names, _) = fields_data(&self.fields);
```
milapsheth, line 63:
```suggestion
        let field_names = self.fields_data();
```
should be a trait method as well. And do you need the other return value?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +12,122 @@ enum Value {
...
+        let variant_ident = &self.ident;
+
+        if field_names.is_empty() {
+            quote! { #enum_name::#variant_ident }
+        } else {
+            let field_names = field_names.iter().map(|name| {
+                quote! { #name.clone() }
```
milapsheth, line 70: Avoid the need for a clone

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -9,9 +12,122 @@ enum Value {
...
+            });
+            quote! { #enum_name::#variant_ident(#(#field_names),*) }
+        }
+    }
+}
+
+impl Value {
+    /// Returns the getter, setter, and remover functions for a storage enum variant.
+    fn fns(&self, enum_name: &Ident, storage_type: &StorageType, variant: &Variant) -> TokenStream {
```
milapsheth, line 79:
```suggestion
    fn storage_fns(&self, enum_name: &Ident, storage_type: &StorageType, variant: &Variant) -> TokenStream {
```

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -21,24 +137,223 @@ enum StorageType {
...
+        param_list: &TokenStream,
+        key: &TokenStream,
+    ) -> TokenStream {
         match self {
-            Self::Instance => quote! { instance },
-            Self::Persistent => quote! { persistent },
-            Self::Temporary => quote! { temporary },
+            Self::Persistent => quote! {
+                pub fn #getter_name(#param_list) -> bool {
+                    let value = env.storage()
```
milapsheth, line 172: use `let key = #storage_key;` and then reuse `key` variable below

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -21,24 +137,223 @@ enum StorageType {
...
+                    value
+                }
+
+                pub fn #setter_name(#param_list) {
+                    env.storage()
+                        .persistent()
+                        .set(&#key, &());
+                }
+
+                pub fn #remover_name(#param_list) {
```
milapsheth, line 189: This duplication seems unnecessary? Have a separate match block first that creates the ttl method token stream and the storage type name. Then, all methods can be defined generically?

packages/stellar-axelar-std-derive/src/testdata/storage_schema_generation_succeeds.golden
nbayindirli: Resolved this

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -20,23 +20,36 @@ impl TryFrom<&[Attribute]> for Value {
...
+    }
+}
+
+trait FieldsExt {
+    fn fields_data(&self) -> (Vec<&Option<Ident>>, Vec<&Type>);
+}
+
+impl FieldsExt for Fields {
+    /// Returns the field names and types of a storage enum variant.
+    fn fields_data(&self) -> (Vec<&Option<Ident>>, Vec<&Type>) {
```
milapsheth, line 44: Why is the return element type `Option<Ident>` instead of `Ident`? It should always have a named field?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -20,23 +20,36 @@ impl TryFrom<&[Attribute]> for Value {
...
+            ))
         } else {
-            Err("value attribute must contain a type parameter: #[value(Type)]".to_string())
+            Err("value attribute must contain a type parameter: #[value(Type)]".into())
+        }
+    }
+}
+
+trait FieldsExt {
+    fn fields_data(&self) -> (Vec<&Option<Ident>>, Vec<&Type>);
```
milapsheth, line 39:
```suggestion
    fn names(&self) -> Vec<&Option<Ident>>;

    fn types(&self) -> Vec<&Type>;
```
decouple independent return values. This avoids the `(.., _) = self.fields.fields_data()` oddity

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -110,6 +128,106 @@ impl Value {
         }
     }
 
+    pub fn status_fns(
```
milapsheth, line 131:
```suggestion
    fn status_fns(
```
does it need to be `pub`?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -110,6 +128,106 @@ impl Value {
...
+            pub fn #remover_name(#param_list) {
+                let key = #storage_key;
+                env.storage()
+                    .#storage_method
+                    .remove(&key);
+            }
+        }
+    }
+
+    pub fn value_fns(
```
milapsheth, line 173: same

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -110,6 +128,106 @@ impl Value {
...
+        &self,
+        storage_type: &StorageType,
+        getter_name: &Ident,
+        setter_name: &Ident,
+        remover_name: &Ident,
+        try_getter_name: &Ident,
+        param_list: &TokenStream,
```
milapsheth, line 180: `param_list` -> `params`. Not a fan of suffixing types to variable names

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -158,202 +276,21 @@ impl TryFrom<&[Attribute]> for StorageType {
...
-
-                pub fn #remover_name(#param_list) {
-                    env.storage()
-                        .temporary()
-                        .remove(&#key)
-                }
-            },
+            Self::Persistent => quote! { persistent() },
+            Self::Instance => quote! { instance() },
+            Self::Temporary => quote! { temporary() },
```
milapsheth, line 283:
```suggestion
            Self::Persistent => quote! { env.storage().persistent() },
            Self::Instance => quote! { env.storage().instance() },
            Self::Temporary => quote! { env.storage().temporary() },
```
less weird?

milapsheth (approved):

packages/stellar-axelar-std-derive/src/storage.rs
nbayindirli: Ah you're right, updated this to no longer be optional since it will always contain a named field

PR #244, opened 2025-02-06
refactor(axelar-operators): use contractstorage

nbayindirli:
AXE-7343

### DIFF
--- a/contracts/stellar-axelar-operators/src/contract.rs
+++ b/contracts/stellar-axelar-operators/src/contract.rs
@@ -6,7 +6,7 @@ use stellar_axelar_std::{ensure, interfaces, Ownable, Upgradable};
 use crate::error::ContractError;
 use crate::event::{OperatorAddedEvent, OperatorRemovedEvent};
 use crate::interface::AxelarOperatorsInterface;
-use crate::storage_types::DataKey;
+use crate::storage;
 
 #[contract]
 #[derive(Ownable, Upgradable)]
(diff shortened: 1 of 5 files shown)


contracts/stellar-axelar-operators/src/contract.rs
```diff
@@ -22,22 +22,18 @@ impl AxelarOperators {
...
 
-        let key = DataKey::Operators(account.clone());
-
         ensure!(
-            !env.storage().instance().has(&key),
+            !storage::is_operators(&env, account.clone()),
             ContractError::OperatorAlreadyAdded
         );
 
-        env.storage().instance().set(&key, &true);
```
milapsheth, line 40: note that we were setting `true` before and now we're setting `()` which are different values. however, since we never interpret the value now and only test existence of the key, a storage migration isn't needed. We can just remove and add the same operators back to clean the state. Let's track this as a step in the v1.1 release doc

milapsheth (approved):

contracts/stellar-axelar-operators/src/contract.rs
```diff
@@ -49,14 +45,12 @@ impl AxelarOperatorsInterface for AxelarOperators {
     fn remove_operator(env: Env, account: Address) -> Result<(), ContractError> {
         Self::owner(&env).require_auth();
 
-        let key = DataKey::Operators(account.clone());
-
         ensure!(
-            env.storage().instance().has(&key),
+            storage::is_operators(&env, account.clone()),
```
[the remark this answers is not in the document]
nbayindirli: Yeah.. chuckled at this one. Here's the ticket: https://axelarnetwork.atlassian.net/browse/AXE-7427

contracts/stellar-axelar-operators/src/contract.rs
nbayindirli: https://axelarnetwork.atlassian.net/browse/AXE-7428

PR #245, opened 2025-02-06
refactor(interchain-token): use contractstorage

nbayindirli:
AXE-7344


contracts/stellar-interchain-token/src/contract.rs
```diff
@@ -297,17 +280,23 @@ impl InterchainToken {
...
-                .extend_ttl(&key, live_for, live_for)
+            storage::set_allowance(
+                env,
+                key,
+                Some(TemporaryStorageParamsExt {
+                    threshold: live_for,
+                    extend_to: live_for,
+                }),
+                &allowance,
+            );
```
milapsheth, line 299: this is a pretty ugly workaround. Just create a new storage method `*_extend_ttl` that takes the key fields and extension params. Don't need to make the setter more complicated

nbayindirli: Ended up refactoring `ttn_fn` to `custom_ttl_fn`, where `ttl_fn` is the vanilla .extend_ttl call on the env

There is some room for improvement around `instance` of `ttl_fn` since that does not take a key like persistent/temp

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -96,6 +96,14 @@ impl VariantExt for Variant {
     }
 }
 
+struct FnNames {
```
milapsheth, line 99:
```suggestion
struct StorageFunctions {
```
maybe?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -96,6 +96,14 @@ impl VariantExt for Variant {
...
+    getter_name: Ident,
+    setter_name: Ident,
+    remover_name: Ident,
+    try_getter_name: Ident,
+    ttl_extender_name: Ident,
```
milapsheth, line 104:
```suggestion
    getter: Ident,
    setter: Ident,
    remover: Ident,
    try_getter: Ident,
    ttl_extender: Ident,
```
name is implied?

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -104,35 +112,19 @@ impl Value {
         storage_type: &StorageType,
         variant: &Variant,
     ) -> TokenStream {
-        let (getter_name, setter_name, remover_name, try_getter_name) =
-            self.fn_names(&variant.ident);
+        let fn_names = self.fn_names(&variant.ident);
```
milapsheth, line 115:
```suggestion
        let storage_functions = self.storage_functions(&variant.ident);
```

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -283,6 +296,18 @@ impl StorageType {
...
+                quote! { env.storage().persistent().extend_ttl(&#ttl_fn_key, threshold, extend_to); }
+            }
+            Self::Instance => quote! { env.storage().instance().extend_ttl(threshold, extend_to); },
+            Self::Temporary => {
+                quote! { env.storage().temporary().extend_ttl(&#ttl_fn_key, threshold, extend_to); }
+            }
+        }
+    }
+
+    fn custom_ttl_fn(&self, ttl_fn_key: &TokenStream) -> TokenStream {
```
milapsheth, line 310:
```suggestion
    fn default_ttl_fn(&self, ttl_fn_key: &TokenStream) -> TokenStream {
```

packages/stellar-axelar-std-derive/src/storage.rs
```diff
@@ -142,21 +134,26 @@ impl Value {
...
+            setter_name,
+            remover_name,
+            ttl_extender_name,
+            ..
+        }: &FnNames,
         params: &TokenStream,
         storage_key: &TokenStream,
     ) -> TokenStream {
         let storage_method = storage_type.storage_method();
         let ttl_fn = storage_type.ttl_fn(&quote! { key });
```
milapsheth, line 148:
```suggestion
        let ttl_fn = storage_type.ttl_fn(&quote! { key }, &quote! { threshold }, ...);
```
parameterize the variable names

milapsheth (approved):

packages/stellar-axelar-std-derive/src/storage.rs
nbayindirli: Hm clashes with `storage_fns` below. Updated to `StorageFunctionNames`

Also updated `fns` to `functions` everywhere

packages/stellar-axelar-std-derive/src/storage.rs
nbayindirli: Ah yep, added this already

PR #246, opened 2025-02-06
refactor(interchain-token-service): use contractstorage

nbayindirli:
AXE-7345


contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -53,94 +53,59 @@ impl InterchainTokenService {
...
-        let key = DataKey::TrustedChain(chain.clone());
-
         ensure!(
-            !env.storage().persistent().has(&key),
+            !storage::is_trusted_chain(env, chain.clone()),
             ContractError::TrustedChainAlreadySet
         );
 
-        env.storage().persistent().set(&key, &());
+        storage::set_trusted_chain_status(env, chain.clone());
```
nbayindirli, line 108: https://axelarnetwork.atlassian.net/browse/AXE-7428

contracts/stellar-interchain-token-service/src/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+    #[value(i128)]
+    FlowLimit { token_id: BytesN<32> },
+
+    #[temporary]
+    #[value(i128)]
+    FlowOut { flow_key: FlowKey },
+
+    #[temporary]
+    #[value(i128)]
+    FlowIn { flow_key: FlowKey },
```
milapsheth, line 55: remove the need for a wrapper key similar to message approval key

contracts/stellar-interchain-token-service/src/storage.rs
```diff
@@ -0,0 +1,71 @@
...
+
+    #[instance]
+    #[value(BytesN<32>)]
+    InterchainTokenWasmHash,
+
+    #[instance]
+    #[value(BytesN<32>)]
+    TokenManagerWasmHash,
+
+    #[persistent]
```
milapsheth, line 37: should look into making this instance type. We have to see if it can fit within the instance storage limits (e.g. can we trust 10k chains and still be under the limit). If the gas cost impact isn't significant, it's not needed

contracts/stellar-interchain-token-service/src/contract.rs
milapsheth: why is it needed here? The value is `()` here instead of a bool

milapsheth (approved):

contracts/stellar-interchain-token-service/src/storage.rs
nbayindirli: https://axelarnetwork.atlassian.net/browse/AXE-7485

PR #247, opened 2025-02-07
refactor(upgrader): use contractstorage

nbayindirli:
AXE-7346


contracts/stellar-upgrader/src/tests/utils/dummy_contract.rs
```diff
@@ -38,8 +38,10 @@ impl DummyContract {
     }
 }
 
-#[contracttype]
+#[contractstorage]
 pub enum DataKey {
+    #[instance]
+    #[value(soroban_sdk::String)]
     Data,
 }
```
milapsheth, line 46: wrap this in a `mod storage` to use the same `storage::*` pattern and hide the enum from the scope

milapsheth (approved):

PR #248, opened 2025-02-07
feat(gas-service): allow pay gas with zero amount

ahramy:
[AXE-7480](https://axelarnetwork.atlassian.net/browse/AXE-7480)

[AXE-7480]: https://axelarnetwork.atlassian.net/browse/AXE-7480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/stellar-axelar-gas-service/src/contract.rs
+++ b/contracts/stellar-axelar-gas-service/src/contract.rs
@@ -40,7 +40,7 @@ impl AxelarGasServiceInterface for AxelarGasService {
     ) -> Result<(), ContractError> {
         spender.require_auth();
 
-        ensure!(token.amount > 0, ContractError::InvalidAmount);
+        ensure!(token.amount >= 0, ContractError::InvalidAmount);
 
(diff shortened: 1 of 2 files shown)


contracts/stellar-axelar-gas-service/src/contract.rs
```diff
@@ -40,7 +40,7 @@ impl AxelarGasServiceInterface for AxelarGasService {
     ) -> Result<(), ContractError> {
         spender.require_auth();
 
-        ensure!(token.amount > 0, ContractError::InvalidAmount);
+        ensure!(token.amount >= 0, ContractError::InvalidAmount);
```
milapsheth, line 43: Make `gas_token` an `Option<Token>` in ITS instead. Same for the example contract. If it's not attached, gas service isn't called. This will mimic the EVM ITS reference behaviour

PR #254, opened 2025-02-12
fix(github-workflows): contract-storage-diff hang

nbayindirli:
Addresses issue where the `contract-storage-diff` workflow would hang on `Expected — Waiting for status to be reported` if a `**/*ensure_*_storage_schema_is_unchanged.golden` was not present in the PR's diff.

### DIFF
--- a/.github/workflows/contract-storage-diff.yaml
+++ b/.github/workflows/contract-storage-diff.yaml
@@ -1,9 +1,6 @@
 name: Contract Storage Schema Migration Check
 
-on:
-  pull_request:
-    paths:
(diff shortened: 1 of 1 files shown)


milapsheth (approved):

.github/workflows/contract-storage-diff.yaml
```diff
@@ -1,9 +1,6 @@
 name: Contract Storage Schema Migration Check
```
milapsheth, line 2: PR title should be something like `ci: always run contract-storage-diff`

PR #258, opened 2025-02-16
chore: remove all unused derive macros

nbayindirli:
AXE-7017


contracts/stellar-interchain-token-service/src/deployer.rs
```diff
@@ -84,7 +84,7 @@ pub fn deploy_token_manager(
         token_id,
         token_address,
         token_manager: deployed_address.clone(),
-        token_manager_type,
+        token_manager_type: token_manager_type.clone(),
```
milapsheth: weird change. since you need a copy anyways, require that in the function argument so the behaviour is more clear to the caller. It's also consistent with what we do for other args.

contracts/stellar-axelar-gateway/src/storage.rs
```diff
@@ -2,22 +2,21 @@ use soroban_sdk::{contracttype, BytesN, String};
...
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct MessageApprovalKey {
     pub source_chain: String,
     pub message_id: String,
 }
 
 #[contracttype]
-#[derive(Clone, Debug, Eq, PartialEq)]
+#[derive(PartialEq)]
```
milapsheth, line 12: clippy warns that `Eq` should be derived as if `PartialEq` is provided. Good practice in general, but it's not really needed for storage data key and error types. We could whitelist it in Cargo.toml

contracts/stellar-axelar-gateway/src/storage.rs
```diff
@@ -2,22 +2,21 @@ use soroban_sdk::{contracttype, BytesN, String};
...
 pub enum MessageApprovalValue {
     NotApproved,
     Approved(BytesN<32>),
     Executed,
 }
 
 #[contractstorage]
-#[derive(Clone, Debug)]
 pub enum DataKey {
```
milapsheth:
```suggestion
enum DataKey {
```
DataKey should be private now

contracts/stellar-example/src/storage.rs
```diff
@@ -2,7 +2,6 @@ use soroban_sdk::Address;
 use stellar_axelar_std::contractstorage;
 
 #[contractstorage]
-#[derive(Clone, Debug)]
```
milapsheth, line 5: Fine to remove it for storage enums that are not public

ahramy (approved):

PR #259, opened 2025-02-16
fix: remove redundant ttl extensions

nbayindirli:
AXE-7627

### DIFF
--- a/contracts/stellar-axelar-gateway/src/contract.rs
+++ b/contracts/stellar-axelar-gateway/src/contract.rs
@@ -1,7 +1,6 @@
 use soroban_sdk::xdr::ToXdr;
 use soroban_sdk::{contract, contractimpl, Address, Bytes, BytesN, Env, String, Vec};
 use stellar_axelar_std::events::Event;
-use stellar_axelar_std::ttl::extend_instance_ttl;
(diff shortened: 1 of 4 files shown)


contracts/stellar-axelar-gateway/src/contract.rs
```diff
@@ -183,8 +182,6 @@ impl AxelarGatewayInterface for AxelarGateway {
             MessageApprovedEvent { message }.emit(env);
         }
 
-        extend_instance_ttl(env);
```
milapsheth, line 186: What about `extend_persistent_ttl`? Any references to it?

milapsheth (approved):

contracts/stellar-axelar-gateway/src/contract.rs
```diff
@@ -1,7 +1,6 @@
 use soroban_sdk::xdr::ToXdr;
```
milapsheth, line 1: PR title can be `chore: ..` since it implies multiple contracts are affected

contracts/stellar-axelar-gateway/src/contract.rs
milapsheth: doesn't look like it

PR #260, opened 2025-02-19
fix(interchain-token-service): validate token metadata on registerCanonicalToken

TanvirDeol:
[AXE-7191](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-7191)

[AXE-7191]: https://axelarnetwork.atlassian.net/browse/AXE-7191?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -224,6 +224,8 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
         env: &Env,
```
milapsheth, line 224: This isn't a feature but `fix(...): ...`

contracts/stellar-interchain-token-service/src/tests/register_canonical_token.rs
```diff
@@ -14,18 +14,22 @@ use crate::types::TokenManagerType;
 fn register_canonical_token_succeeds() {
     let (env, client, _, _, _) = setup_env();
     let token_address = Address::generate(&env);
-    let expected_id = client.canonical_interchain_token_id(&token_address);
+    let registered_token = &env.register_stellar_asset_contract_v2(token_address);
```
milapsheth, line 17: Look carefully at what the input to this method should be, and not just whether it passes typecheck

milapsheth, line 17:
```suggestion
    let token = &env.register_stellar_asset_contract_v2(token_address);
```

contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -224,6 +224,8 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
         token_address: Address,
     ) -> Result<BytesN<32>, ContractError> {
+        token_metadata::token_metadata(env, &token_address, &Self::native_token_address(env))?;
```
milapsheth, line 227: add comment to explain that this is used for validating the token address and metadata. Use `let _ = ...` to be explicit that return value is intentionally ignored

contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -224,6 +224,10 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
         env: &Env,
         token_address: Address,
     ) -> Result<BytesN<32>, ContractError> {
+        // Validates the token address and it's associated token metadata
+        let _ =
+            token_metadata::token_metadata(env, &token_address, &Self::native_token_address(env))?;
```
ahramy, line 229: Would we return `InvalidTokenMetaData` or `InvalidTokenAddress` in this case?

ahramy: note: I noticed that `InvalidTokenMetaData` not being used anywhere, we can remove it if it's not being used

contracts/stellar-interchain-token-service/src/tests/register_canonical_token.rs
```diff
@@ -72,3 +74,12 @@ fn canonical_token_id_derivation() {
         hex::encode(token_id.to_array())
     ]);
 }
+
+#[test]
+#[should_panic(expected = "HostError: Error(Storage, MissingValue)")]
```
ahramy, line 79: Instead of checking `should_panic`, checking `InvalidTokenMetaData` or `InvalidTokenAddress` looks better

contracts/stellar-interchain-token-service/src/tests/register_canonical_token.rs
```diff
@@ -14,22 +14,19 @@ use crate::types::TokenManagerType;
 fn register_canonical_token_succeeds() {
     let (env, client, _, _, _) = setup_env();
     let token_address = Address::generate(&env);
-    let registered_token = &env.register_stellar_asset_contract_v2(token_address);
-    let expected_id = client.canonical_interchain_token_id(&registered_token.address());
+    let token = &env.register_stellar_asset_contract_v2(token_address);
```
milapsheth, line 17: The argument to this function still doesn't make sense. See the doc and our existing usage
https://docs.rs/soroban-sdk/latest/soroban_sdk/struct.Env.html#method.register_stellar_asset_contract_v2

contracts/stellar-interchain-token-service/src/tests/deploy_remote_canonical_token.rs
```diff
@@ -162,7 +162,7 @@ fn deploy_remote_canonical_token_succeeds_without_name_truncation() {
 }
 
 #[test]
-#[should_panic(expected = "HostError: Error(Storage, MissingValue)")]
```
TanvirDeol, line 165: Had to make this change since `client.register_canonical_token` now has a `token_metadata()` call which returns `ContractError::InvalidTokenAddress` let me know if that's alright.

contracts/stellar-interchain-token-service/src/tests/deploy_remote_canonical_token.rs
```diff
@@ -162,7 +162,7 @@ fn deploy_remote_canonical_token_succeeds_without_name_truncation() {
...
+#[should_panic(expected = "HostError: Error(Contract, #16)")] // InvalidTokenAddress
 fn deploy_remote_canonical_token_fail_no_actual_token() {
```
milapsheth, line 166: Doesn't `register_canonical_token` fail now and not deploy_remote_canonical_token? Refactor the test accordingly. Also, use `assert_contract_err` instead of testing for panic

contracts/stellar-interchain-token-service/src/tests/register_canonical_token.rs
```diff
@@ -76,10 +76,12 @@ fn canonical_token_id_derivation() {
 }
 
 #[test]
-#[should_panic(expected = "HostError: Error(Storage, MissingValue)")]
 fn register_canonical_token_fails_if_invalid_token_address() {
```
milapsheth, line 79: The `deploy_remote_canonical_*` test is a duplicate now and not needed

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -53,9 +53,18 @@ pub fn token_metadata(
     native_token_address: &Address,
 ) -> Result<TokenMetadata, ContractError> {
     let token = token::Client::new(env, token_address);
-    let decimals = token.decimals();
-    let name = token.name();
-    let symbol = token.symbol();
+    let decimals = token
+        .try_decimals()
+        .map_err(|_| ContractError::InvalidTokenAddress)?
+        .map_err(|_| ContractError::InvalidTokenAddress)?;
```
milapsheth, line 59: this is a bit confusing, take a look at what the two nested results correspond to. they fail due to different errors. We might still want to always return this error in both cases, but would be good to understand the difference

contracts/stellar-interchain-token-service/src/error.rs
```diff
@@ -35,6 +35,7 @@ pub enum ContractError {
     InvalidTokenDecimals = 27,
     ContractPaused = 28,
     InvalidInitialSupply = 29,
+    ConversionError = 30,
```
milapsheth, line 38:
```suggestion
    TokenInvocationError = 30,
```
or something similar, needs to be more readable

milapsheth (approved):

PR #261, opened 2025-02-19
feat(interchain-token-service): make gas token optional for its and example

ahramy:
[AXE-7480](https://axelarnetwork.atlassian.net/browse/AXE-7480)

[AXE-7480]: https://axelarnetwork.atlassian.net/browse/AXE-7480?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/stellar-axelar-gas-service/src/testutils.rs
```diff
@@ -14,7 +14,7 @@ pub fn setup_gas_service<'a>(env: &Env) -> AxelarGasServiceClient<'a> {
     gas_service_client
 }
 
-pub fn setup_gas_token(env: &Env, sender: &Address) -> Token {
+pub fn setup_gas_token(env: &Env, sender: &Address) -> Option<Token> {
```
milapsheth, line 17: Bad API, this function always returns a token, so making the return type into an Option is more confusing for callers as can be seen in setup_tokens later

contracts/stellar-interchain-token-service/src/tests/deploy_remote_canonical_token.rs
```diff
@@ -64,13 +65,15 @@ fn deploy_remote_canonical_token_succeeds() {
         InterchainTokenDeploymentStartedEvent,
     >(&env, -4));
 
+    let _gas_token = gas_token.as_ref().unwrap();
```
milapsheth, line 68: Avoid using underscore prefix when the variable is actually being used

contracts/stellar-interchain-token-service/src/tests/deploy_remote_interchain_token.rs
```diff
@@ -77,7 +80,7 @@ fn deploy_remote_interchain_token_succeeds() {
             its_hub_address,
             payload,
             &sender,
-            gas_token.clone(),
+            _gas_token.clone(),
```
milapsheth, line 83:
```suggestion
            Some(gas_token.clone()),
```
After fixing the confusing setup API

contracts/stellar-interchain-token-service/src/tests/interchain_transfer.rs
```diff
@@ -111,6 +111,86 @@ fn interchain_transfer_canonical_token_send_succeeds() {
...
+        &gas_token,
+    );
+
+    goldie::assert!(events::fmt_emitted_event_at_idx::<
+        InterchainTransferSentEvent,
+    >(&env, -2));
+}
+
+#[test]
+fn interchain_transfer_canonical_token_send_succeeds_without_gas_token() {
```
milapsheth, line 143: Don't need to test each token type since the functionality doesn't depend on it

ahramy: Testing ITS functions and example without a gas token is also for coverage, as we now have a new branch and the event index and authentication check differ since we don't call pay_gas. I think it's better to keep it, do you think we should remove it?

contracts/stellar-axelar-gas-service/src/testutils.rs
ahramy: I see. make sense. updated

contracts/stellar-interchain-token-service/src/tests/deploy_remote_canonical_token.rs
ahramy: make sense. updated

contracts/stellar-interchain-token-service/src/tests/deploy_remote_interchain_token.rs
ahramy: updated

contracts/stellar-interchain-token-service/src/tests/interchain_transfer.rs
ahramy: Had a discussion in a separate channel, it does increase `regions` but it would not increase line coverage of ITS. hence remove those unit tests

contracts/stellar-interchain-token-service/src/tests/interchain_transfer.rs
```diff
@@ -253,7 +253,7 @@ fn interchain_transfer_fails_with_invalid_token() {
             &destination_address,
```
milapsheth, line 253: Why remove all coverage for calling entrypoints without a gas token. My comment was only applicable for interchain transfer of canonical token test, since another interchain transfer test already covered lack of gas token

ahramy: ah, updated PR, removed only interchain transfer of canonical token test

contracts/stellar-interchain-token-service/src/tests/deploy_remote_canonical_token.rs
```diff
@@ -188,6 +228,6 @@ fn deploy_remote_canonical_token_fail_no_actual_token() {
         &token_address,
         &destination_chain,
         &spender,
-        &gas_token,
+        &Some(gas_token),
```
TanvirDeol, line 231: Can we make this call return a more descriptive error than `HostError: Error(Storage, MissingValue)`? Similar to how it is done in `deploy_remote_interchain_token.rs`

contracts/stellar-interchain-token-service/src/tests/deploy_remote_canonical_token.rs
```diff
@@ -92,13 +93,52 @@ fn deploy_remote_canonical_token_succeeds() {
...
+    client.mock_all_auths().deploy_remote_canonical_token(
+        &token_address,
+        &destination_chain,
+        &spender,
+        &gas_token,
+    );
+
+    goldie::assert!(events::fmt_emitted_event_at_idx::<
+        InterchainTokenDeploymentStartedEvent,
+    >(&env, -2));
```
TanvirDeol, line 131: Is it worth adding a small comment here on what the -2 represents or the reason why we assert the 2nd last emitted event rather than the default?

ahramy: actually it's a good call, addressed comments by making it as constant.

contracts/stellar-interchain-token-service/src/tests/utils/mod.rs
```diff
@@ -38,3 +38,9 @@ impl TokenMetadataExt for TokenMetadata {
         }
     }
 }
+
+pub const INTERCHAIN_TOKEN_DEPLOYED_EVENT_IDX: i32 = -4;
+pub const INTERCHAIN_TOKEN_DEPLOYED_WITHOUT_GAS_TOKEN_EVENT_IDX: i32 = -2;
+pub const INTERCHAIN_TOKEN_DEPLOYED_NO_SUPPLY_EVENT_IDX: i32 =
+    INTERCHAIN_TOKEN_DEPLOYED_EVENT_IDX + 1;
+pub const TOKEN_MANAGER_DEPLOYED_EVENT_IDX: i32 = INTERCHAIN_TOKEN_DEPLOYED_EVENT_IDX + 1;
```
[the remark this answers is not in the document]
milapsheth: you'd need to provide the types for each event which is cumbersome. and asserting on raw events is not as readable. and the test doesn't really care about the other event values

contracts/stellar-interchain-token-service/src/tests/deploy_remote_interchain_token.rs
```diff
@@ -40,14 +44,14 @@ fn deploy_remote_interchain_token_succeeds() {
...
-        &gas_token,
+        &Some(gas_token.clone()),
     );
 
     assert_eq!(token_id, deployed_token_id);
 
     goldie::assert!(events::fmt_emitted_event_at_idx::<
         InterchainTokenDeploymentStartedEvent,
-    >(&env, -4));
+    >(&env, INTERCHAIN_TOKEN_DEPLOYED_EVENT_IDX));
```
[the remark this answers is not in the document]
milapsheth: yeah, we've done something similar elsewhere

contracts/stellar-interchain-token-service/src/tests/deploy_remote_canonical_token.rs
```diff
@@ -92,13 +96,55 @@ fn deploy_remote_canonical_token_succeeds() {
...
+        InterchainTokenDeploymentStartedEvent,
+    >(
+        &env,
+        INTERCHAIN_TOKEN_DEPLOYED_WITHOUT_GAS_TOKEN_EVENT_IDX
+    ));
+
+    let deploy_remote_canonical_token_auth = auth_invocation!(
+        &env,
+        spender,
+        client.deploy_remote_canonical_token(token_address, destination_chain, spender, gas_token)
```
milapsheth, line 142:
```suggestion
        spender,
        client.deploy_remote_canonical_token(token_address, destination_chain, spender, gas_token)
```
`env` can be retrieved via `client.env` like for other macros. Create another PR to simplify it

milapsheth (approved):

contracts/stellar-interchain-token-service/src/tests/deploy_remote_canonical_token.rs
ahramy: yeah, make sense. created a ticket: https://axelarnetwork.atlassian.net/browse/AXE-7831

PR #263, opened 2025-02-20
fix(interchain-token-service): ensure token metadata has valid ASCII encoding

TanvirDeol:
[AXE-7195](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-7195)

[AXE-7195]: https://axelarnetwork.atlassian.net/browse/AXE-7195?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -42,11 +46,31 @@ impl TokenMetadataExt for TokenMetadata {
...
+        ensure!(
+            is_ascii_string(&self.symbol),
+            ContractError::InvalidTokenSymbol
+        );
 
         Ok(())
     }
 }
 
+fn is_ascii_string(input: &String) -> bool {
```
milapsheth, line 59: Create a `StringExt` trait that defines an `is_ascii` method in `stellar-axelar-std` to do this in a cleaner way

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -42,11 +46,31 @@ impl TokenMetadataExt for TokenMetadata {
...
+    let mut bytes: Vec<u8> = Vec::new();
```
milapsheth, line 60: Avoid using `alloc` which is expensive

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -42,11 +46,31 @@ impl TokenMetadataExt for TokenMetadata {
...
+    bytes.resize(input.len() as usize, 0);
```
milapsheth, line 61: Since we require the length to be bounded, you can allocate a fixed size slice instead. Call the trait helper `is_bounded_ascii_str(string, bound) -> bool`

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -72,3 +96,44 @@ pub fn token_metadata(
...
+
+    #[test]
+    fn token_metadata_valid_ascii() {
+        let env = Env::default();
+
+        let name = String::from_str(&env, "Test");
+        let symbol = String::from_str(&env, "Test");
+        let decimals = 18;
+
+        let result = TokenMetadata::new(name, symbol, decimals);
```
milapsheth, line 112: use `assert_ok!` instead of `unwrap`, see existing usage

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -72,3 +96,44 @@ pub fn token_metadata(
...
+    #[test]
+    fn token_metadata_invalid_ascii_symbol() {
+        let env = Env::default();
+
+        let name = String::from_str(&env, "Test");
+        let symbol = String::from_str(&env, "Test世界！");
+        let decimals = 18;
+
+        let result = TokenMetadata::new(name, symbol, decimals);
+        assert!(matches!(result, Err(ContractError::InvalidTokenSymbol)));
```
milapsheth, line 137: same

contracts/stellar-interchain-token-service/src/token_metadata.rs
TanvirDeol: I started off using fixed length slices because of the bounding, but for the `copy_into_slice()`method, soroban_sdk requires the String and slice length to be the same size. So that made me switch to dynamically sized slices.

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -72,3 +96,44 @@ pub fn token_metadata(
...
+    #[test]
+    fn token_metadata_invalid_ascii_name() {
+        let env = Env::default();
+
+        let name = String::from_str(&env, "Test世界！");
+        let symbol = String::from_str(&env, "Test");
+        let decimals = 18;
+
+        let result = TokenMetadata::new(name, symbol, decimals);
+        assert!(matches!(result, Err(ContractError::InvalidTokenName)));
```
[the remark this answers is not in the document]
TanvirDeol: The reason I can't use `assert_err` or similar functions here is because `soroban_token_sdk::metadata::TokenMetadata` doesn't implement the `Debug` trait. So when you try to print some `TokenMetadata` result using `{:?}` like in the `assert_err!` implementation it throws an error.

contracts/stellar-interchain-token-service/src/token_metadata.rs
TanvirDeol: See above comment

contracts/stellar-interchain-token-service/src/token_metadata.rs
milapsheth: Hmm, I see. Can you check with stellar team when they can release an improvement for this? Relevant thread:
https://interoplabs.slack.com/archives/C04545F3S59/p1737143412288319?thread_ts=1736889117.914599&cid=C04545F3S59

contracts/stellar-interchain-token-service/src/token_metadata.rs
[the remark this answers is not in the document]
milapsheth: Report an issue on the soroban github so they address it in the future

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -1,6 +1,9 @@
+extern crate alloc;
```
milapsheth, line 1: PR title should be `fix(interchain-token-service): ...`

packages/stellar-axelar-std/src/traits.rs
milapsheth, line 1: Create a `string.rs` instead

packages/stellar-axelar-std/src/traits.rs
```diff
@@ -1,3 +1,28 @@
+extern crate alloc;
```
milapsheth, line 1: Feature gate this under a new `alloc` feature in `Cargo.toml`, so it's not included by default. ITS cargo.toml can explicitly enable `alloc` feature on stellar-axelar-std

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -72,3 +77,46 @@ pub fn token_metadata(
...
+        let name = String::from_str(&env, "Test世界！");
+        let symbol = String::from_str(&env, "Test");
+        let decimals = 18;
+
+        let result = TokenMetadata::new(name, symbol, decimals);
+        assert!(matches!(result, Err(ContractError::InvalidTokenName)));
+    }
+
+    #[test]
+    fn token_metadata_invalid_ascii_symbol() {
```
milapsheth, line 112:
```suggestion
    fn token_metadata_new_fails_with_invalid_ascii_symbol() {
```

milapsheth: similarly update others to be more readable

packages/stellar-axelar-std/src/traits.rs
```diff
@@ -1,3 +1,28 @@
...
+use alloc::vec::Vec;
+
+pub trait StringExt {
+    fn is_ascii(&self) -> bool;
+}
+
+impl StringExt for soroban_sdk::String {
+    fn is_ascii(&self) -> bool {
+        let mut bytes: Vec<u8> = Vec::new();
+        bytes.resize(self.len() as usize, 0);
```
milapsheth, line 12:
```suggestion
        let mut bytes: Vec<u8> = Vec::with_capacity(self.len());
```

packages/stellar-axelar-std/src/traits.rs
```diff
@@ -1,3 +1,28 @@
...
+        self.copy_into_slice(&mut bytes);
+
+        for &byte in bytes.iter() {
+            let character = byte as char;
```
milapsheth, line 16: Casting to char doesn't seem safe/well defined, since it expects a valid unicode value
https://doc.rust-lang.org/std/primitive.char.html#validity-and-layout

Check the byte value directly instead?

packages/stellar-axelar-std/src/string.rs
```diff
@@ -0,0 +1,25 @@
+#[cfg(feature = "alloc")]
+use alloc::vec;
+#[cfg(feature = "alloc")]
+use alloc::vec::Vec;
```
milapsheth, line 4:
```suggestion
#![cfg(any(test, feature = "alloc"))]

extern crate alloc;

use alloc::vec::{self, Vec};
```
apply the cfg for the whole file instead?

packages/stellar-axelar-std/src/lib.rs
```diff
@@ -4,6 +4,9 @@
 #[cfg(any(test, feature = "testutils"))]
 extern crate std;
 
+#[cfg(feature = "alloc")]
+extern crate alloc;
+
```
milapsheth, line 9:
```suggestion
```
just feature gate within the file?

packages/stellar-axelar-std/src/string.rs
```diff
@@ -0,0 +1,25 @@
...
+
+        for &byte in bytes.iter() {
+            // check if byte within ASCII range (0-127)
+            if byte > ASCII_RANGE {
+                return false;
+            }
+        }
+        true
+    }
+}
```
milapsheth, line 25: add unit tests for `is_ascii` in this file

packages/stellar-axelar-std/src/string.rs
```diff
@@ -0,0 +1,25 @@
...
+use alloc::vec::Vec;
+
+pub trait StringExt {
+    fn is_ascii(&self) -> bool;
+}
+
+#[cfg(feature = "alloc")]
+impl StringExt for soroban_sdk::String {
+    fn is_ascii(&self) -> bool {
+        const ASCII_RANGE: u8 = 127;
```
milapsheth, line 13: should be a global constant
```suggestion
        const ASCII_RANGE: u8 = i8::MAX as u8;
```

packages/stellar-axelar-std/src/string.rs
```diff
@@ -0,0 +1,25 @@
...
+        let mut bytes: Vec<u8> = vec![0; self.len() as usize];
+        self.copy_into_slice(&mut bytes);
+
+        for &byte in bytes.iter() {
+            // check if byte within ASCII range (0-127)
```
milapsheth, line 18:
```suggestion
```
implied

packages/stellar-axelar-std/Cargo.toml
```diff
@@ -25,6 +25,7 @@ stellar-axelar-std-derive = { workspace = true }
 [features]
 testutils = ["soroban-sdk/testutils", "dep:hex"]
 derive = ["dep:stellar-axelar-std-derive"]
+alloc = []
```
milapsheth, line 28:
```suggestion
alloc = ["soroban-sdk/alloc"]
```
does soroban-sdk need alloc feature enabled as well for this?

contracts/stellar-interchain-token-service/src/token_metadata.rs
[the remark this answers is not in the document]
milapsheth: Was this reported?

contracts/stellar-interchain-token-service/src/token_metadata.rs
```diff
@@ -72,3 +75,46 @@ pub fn token_metadata(
...
+    #[test]
+    fn token_metadata_new_succeeds() {
+        let env = Env::default();
+
+        let name = String::from_str(&env, "Test");
+        let symbol = String::from_str(&env, "Test");
+        let decimals = 18;
+
+        let result = TokenMetadata::new(name, symbol, decimals);
+        let _ = assert_ok!(result);
```
milapsheth, line 94:
```suggestion
        assert_ok!(TokenMetadata::new(name, symbol, decimals));
```
result is not used

packages/stellar-axelar-std/Cargo.toml
TanvirDeol: Yeah I had manually added the alloc feature for the soroban-sdk in `stellar-interchain-token-service/Cargo.toml` but this is probably a neater way, so I changed it to this now

contracts/stellar-interchain-token-service/src/token_metadata.rs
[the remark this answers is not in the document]
TanvirDeol: Yes they made the changes [here is the PR](https://github.com/stellar/rs-soroban-sdk/pull/1446). I'm assuming I'm supposed to wait till they do another official release before I make the changes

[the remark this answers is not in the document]
milapsheth: add a TODO in the code to fix it once a new release is created, but we don't need to wait for it

packages/stellar-axelar-std/src/string.rs
```diff
@@ -1,25 +1,73 @@
...
+            "Bonjour le monde 🌍",
+            "¡Hola, mundo!",
+            "Γειά σου Κόσμε",
+            "नमस्ते दुनिया",
+            "שלום עולם",
+            "你好，世界",
+        ];
+
+        let env = Env::default();
+        for &s in non_ascii_strings.iter() {
```
milapsheth, line 68: same

PR #265, opened 2025-02-22
feat: add custom executable interface for AxelarExecutableInterface for ease of use

cgorenflo:


packages/stellar-axelar-std/src/interfaces/mod.rs
```diff
@@ -10,11 +10,16 @@ pub use ownable::*;
...
+#[macro_export]
+macro_rules! derive_only {
+    () => {
+        /// Marker trait for interfaces that should not be implemented by using `contractimpl`.
+        ///
+        /// **DO NOT IMPLEMENT THIS MANUALLY!**
+        #[doc(hidden)]
+        pub trait DeriveOnly {}
+    };
+}
```
cgorenflo, line 22: If I use the same DeriveOnly trait for all traits that need it, the auto-implementations inside the macros might collide, e.g. when a contract implements both the AxelarExecutableInterface and the InterchainTokenExecutableInterface, so to work around that restriction, each crate that defines such a trait must have their own local DeriveOnly trait

milapsheth (approved):

contracts/stellar-axelar-example/src/contract.rs
```diff
@@ -18,7 +16,7 @@ use crate::storage;
 
 #[contract]
 #[derive(InterchainTokenExecutable, AxelarExecutable)]
-pub struct Example;
+pub struct AxelarExample;
```
milapsheth, line 19: why is this being changed again? thought the other PR already fixed it
https://github.com/axelarnetwork/axelar-cgp-stellar/pull/270/files#diff-185c8e9bb12fa807cb4f17f3e017fd0ef9bc3f625aa36e06030fdd5a5dc6bfc3R19

milapsheth (approved):

PR #267, opened 2025-03-04
fix(axelar-std-derive): enforce contractstorage enums are private

nbayindirli:
AXE-7777


.github/scripts/check-migration.sh
```diff
@@ -11,12 +11,39 @@ main() {
     echo "Found changed storage schema files:"
     echo "$CHANGED_FILES"
 
-    # Check each changed storage schema file
+    # Check if changes are only visibility modifiers
+    local significant_changes=false
     while IFS= read -r file; do
         [ -z "$file" ] && continue
-        check_migration_file "$file"
+        if ! is_only_visibility_change "$1" "$2" "$file"; then
```
ahramy, line 18: Maybe you can deefine those variables below the `main()`, for better readability
```
local base_sha="$1"
local head_sha="$2"
...
if ! CHANGED_FILES=$(git diff --name-only "$base_sha" "$head_sha" ...
...
```
```suggestion
        if ! is_only_visibility_change "$base_sha" "$head_sha" "$file"; then
```

.github/scripts/check-migration.sh
```diff
@@ -11,12 +11,39 @@ main() {
     echo "Found changed storage schema files:"
     echo "$CHANGED_FILES"
 
-    # Check each changed storage schema file
+    # Check if changes are only visibility modifiers
+    local significant_changes=false
```
ahramy, line 15:
```suggestion
    local has_significant_changes=false
```

ahramy (approved):

.github/scripts/check-migration.sh
```diff
@@ -11,12 +11,42 @@ main() {
...
+        fi
     done <<< "$CHANGED_FILES"
 
+    if [ "$has_significant_changes" = false ]; then
+        echo "All changes are visibility modifiers only, no migration needed"
+        exit 0
+    fi
+}
+
+is_only_visibility_change() {
```
milapsheth, line 35: this shouldn't be changed. I can change gh branch restrictions to skip the check for this PR, but we shouldn't add workarounds like this in the script itself

PR #269, opened 2025-03-05
chore: update auth invocation macros

ahramy:
[AXE-7831](https://axelarnetwork.atlassian.net/browse/AXE-7831)

- removing `$env` if `auth_invocation!` macro includes client.
- There are some case where `&env` is required (i.e. Token)

[AXE-7831]: https://axelarnetwork.atlassian.net/browse/AXE-7831?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/stellar-axelar-std/src/error.rs
```diff
@@ -229,11 +229,37 @@ macro_rules! mock_auth {
             },
```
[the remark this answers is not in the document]
ahramy: resolved

contracts/stellar-axelar-gas-service/src/tests/test.rs
```diff
@@ -225,9 +223,9 @@ fn add_gas_fails_with_insufficient_user_balance() {
     let sender: Address = Address::generate(&env);
     let message_id = message_id(&env);
     let gas_amount: i128 = 2;
-    let Token { address, .. } = setup_token(&env, &spender, gas_amount - 1);
+    let (_token, _) = setup_token(&env, &spender, gas_amount - 1);
     let token = Token {
-        address,
+        address: _token.address,
```
milapsheth, line 228: why is there an underscore if it's actually being used?

contracts/stellar-axelar-gas-service/src/tests/test.rs
```diff
@@ -104,23 +107,21 @@ fn pay_gas_fails_with_insufficient_user_balance() {
     let spender: Address = Address::generate(&env);
     let sender: Address = Address::generate(&env);
     let gas_amount: i128 = 2;
-    let Token { address, .. } = setup_token(&env, &spender, gas_amount - 1);
+    let (_token, token_client) = setup_token(&env, &spender, gas_amount - 1);
```
milapsheth, line 110: as you can see, this is not a clean solution, so try to iterate more. One solution to consider is add a `impl Token` that defines a `client()` method, so you could just do `token.client()` to get the client

milapsheth: the macro might not work, might need to make it only extract the last `.`, or just assign `token_client = token.client()` in the caller where needed, since a lot of places don't need it

contracts/stellar-axelar-gas-service/src/tests/test.rs
ahramy: updated

contracts/stellar-axelar-gas-service/src/tests/test.rs
```diff
@@ -25,17 +25,20 @@ fn setup_env<'a>() -> (Env, Address, Address, AxelarGasServiceClient<'a>) {
...
-        address: asset.address(),
-        amount,
-    }
+    (
+        Token {
+            address: asset.address(),
+            amount,
+        },
+        TokenClient::new(env, &asset.address()),
+    )
```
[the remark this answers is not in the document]
ahramy: updated

contracts/stellar-axelar-gas-service/src/tests/test.rs
ahramy: with `impl`, the code looks much cleaner. thank you!

milapsheth (approved):

PR #274, opened 2025-03-10
refactor(axelar-gateway): destruct storage wrapper keys

nbayindirli:
AXE-7358

### DIFF
--- a/contracts/stellar-axelar-gateway/src/migrate.rs
+++ b/contracts/stellar-axelar-gateway/src/migrate.rs
@@ -45,19 +45,18 @@ impl CustomMigratableInterface for AxelarGateway {
                 message_id: message_id.clone(),
             };
 
-            let legacy_message_approval =
-                legacy_storage::try_message_approval(env, message_approval_key)
-                    .ok_or(ContractError::InvalidMessageApproval)?;
-
-            let message_approval = match legacy_message_approval {
-                legacy_storage::MessageApprovalValue::Approved(hash) => {
-                    Some(storage::MessageApprovalValue::Approved(hash))
-                }
-                legacy_storage::MessageApprovalValue::Executed => {
-                    Some(storage::MessageApprovalValue::Executed)
-                }
-                legacy_storage::MessageApprovalValue::NotApproved => None,
-            };
+            let message_approval =
+                match legacy_storage::try_message_approval(env, message_approval_key)
+                    .ok_or(ContractError::InvalidMessageApproval)?
+                {
+                    legacy_storage::MessageApprovalValue::Approved(hash) => {
+                        Some(storage::MessageApprovalValue::Approved(hash))
+                    }
+                    legacy_storage::MessageApprovalValue::Executed => {
+                        Some(storage::MessageApprovalValue::Executed)
+                    }
+                    legacy_storage::MessageApprovalValue::NotApproved => None,
+                };
 
             storage::set_message_approval(
                 env,


contracts/stellar-axelar-gateway/src/migrate.rs
```diff
@@ -0,0 +1,89 @@
...
+        #[value(BytesN<32>)]
+        DomainSeparator,
+
+        #[instance]
+        #[value(u64)]
+        MinimumRotationDelay,
+
+        #[instance]
+        #[value(u64)]
+        Epoch,
```
milapsheth, line 44: maybe we don't need to add the other legacy variants since they're not being used? Stellar uses the variant name so the others aren't really needed. Also helps improve code coverage

contracts/stellar-axelar-gateway/src/migrate.rs
```diff
@@ -0,0 +1,89 @@
...
+        #[value(u64)]
+        EpochBySignersHash { signers_hash: BytesN<32> },
+    }
+}
+
+impl CustomMigratableInterface for AxelarGateway {
+    type MigrationData = Vec<(String, String)>;
+    type Error = ContractError;
+
+    fn __migrate(_env: &Env, _migration_data: Self::MigrationData) -> Result<(), Self::Error> {
```
milapsheth, line 64:
```suggestion
    fn __migrate(env: &Env, migration_data: Self::MigrationData) -> Result<(), Self::Error> {
```
remove `_` prefix since it's being used

contracts/stellar-axelar-gateway/src/migrate.rs
```diff
@@ -0,0 +1,89 @@
...
+                message_id,
+            };
+
+            ensure!(
+                legacy_storage::try_message_approval(_env, message_approval_key.clone()).is_some(),
+                ContractError::MessageApprovalNotFound
+            );
+
+            let message_approval =
+                legacy_storage::message_approval(_env, message_approval_key.clone());
```
milapsheth, line 77:
```suggestion
            let Some(message_approval) =
                legacy_storage::try_message_approval(_env, message_approval_key) else { return Err(ContractError::MessageApprovalNotFound) };
```
avoids needing to load twice

contracts/stellar-axelar-gateway/src/migrate.rs
```diff
@@ -0,0 +1,89 @@
...
+
+impl CustomMigratableInterface for AxelarGateway {
+    type MigrationData = Vec<(String, String)>;
+    type Error = ContractError;
+
+    fn __migrate(_env: &Env, _migration_data: Self::MigrationData) -> Result<(), Self::Error> {
+        for (source_chain, message_id) in _migration_data {
+            let message_approval_key = legacy_storage::MessageApprovalKey {
+                source_chain,
+                message_id,
```
milapsheth, line 68:
```suggestion
                source_chain: source_chain.clone(),
                message_id: message_id.clone(),
```

contracts/stellar-axelar-gateway/src/migrate.rs
```diff
@@ -0,0 +1,89 @@
...
+                ContractError::MessageApprovalNotFound
+            );
+
+            let message_approval =
+                legacy_storage::message_approval(_env, message_approval_key.clone());
+
+            storage::set_message_approval(
+                _env,
+                message_approval_key.source_chain.clone(),
+                message_approval_key.message_id.clone(),
```
milapsheth, line 82:
```suggestion
                source_chain,
                message_id,
```
avoids the extra clone?

contracts/stellar-axelar-gateway/src/storage.rs
```diff
@@ -21,7 +14,8 @@ enum DataKey {
     #[persistent]
     #[value(MessageApprovalValue)]
     MessageApproval {
-        message_approval_key: MessageApprovalKey,
+        source_chain: String,
+        message_id: String,
```
milapsheth, line 18: one thing that's a disadvantage with this approach is that the getter/setter only perform type checking but if someone passes `source_chain` and `message_id` in flipped order to them, it can lead to accidental errors. Maybe we can improve this in the future somehow, in addition to relying on unit tests

contracts/stellar-axelar-gateway/src/storage.rs
```diff
@@ -1,13 +1,6 @@
...
-#[contracttype]
-#[derive(Clone, Debug)]
-pub struct MessageApprovalKey {
-    pub source_chain: String,
-    pub message_id: String,
-}
-
 #[contracttype]
 #[derive(Clone, Debug, Eq, PartialEq)]
 pub enum MessageApprovalValue {
```
milapsheth, line 6: we should create a new value type that doesn't include `NotApproved` variant. This variant is never stored and just used as a default when an approval doesn't exist. We should just use `None` returned by `try_*` getter to make this cleaner

contracts/stellar-axelar-gateway/src/contract.rs
```diff
@@ -236,13 +237,8 @@ impl AxelarGateway {
...
         message_id: String,
-    ) -> MessageApprovalValue {
-        let key = MessageApprovalKey {
-            source_chain,
-            message_id,
-        };
-
-        storage::try_message_approval(env, key).unwrap_or(MessageApprovalValue::NotApproved)
+    ) -> Option<MessageApprovalValue> {
+        storage::try_message_approval(env, source_chain, message_id)
```
milapsheth, line 241: This helper seems pointless now, use storage directly?

contracts/stellar-axelar-gateway/src/contract.rs
```diff
@@ -161,21 +163,20 @@ impl AxelarGatewayInterface for AxelarGateway {
...
-            let message_approval_key = MessageApprovalKey {
-                source_chain: message.source_chain.clone(),
-                message_id: message.message_id.clone(),
-            };
-
             // Prevent replay if message is already approved/executed
-            let message_approval = storage::try_message_approval(env, message_approval_key.clone())
-                .unwrap_or(MessageApprovalValue::NotApproved);
-            if message_approval != MessageApprovalValue::NotApproved {
+            let message_approval = storage::try_message_approval(
```
milapsheth, line 167: Use `if let` pattern instead to make this more idiomatic

contracts/stellar-axelar-gateway/src/contract.rs
```diff
@@ -161,21 +163,21 @@ impl AxelarGatewayInterface for AxelarGateway {
...
             // Prevent replay if message is already approved/executed
-            let message_approval = storage::try_message_approval(env, message_approval_key.clone())
-                .unwrap_or(MessageApprovalValue::NotApproved);
-            if message_approval != MessageApprovalValue::NotApproved {
+            if storage::try_message_approval(
+                env,
+                message.source_chain.clone(),
+                message.message_id.clone(),
+            )
+            .is_some()
```
milapsheth, line 172:
```suggestion
            if storage::has_message_approval(
                env,
                message.source_chain.clone(),
                message.message_id.clone(),
            )
```
maybe consider adding a `has_*` storage method

contracts/stellar-axelar-gateway/src/migrate.rs
```diff
@@ -0,0 +1,72 @@
...
+            let message_approval_key = legacy_storage::MessageApprovalKey {
+                source_chain: source_chain.clone(),
+                message_id: message_id.clone(),
+            };
+
+            let legacy_message_approval =
+                legacy_storage::try_message_approval(env, message_approval_key)
+                    .ok_or(ContractError::InvalidMessageApproval)?;
+
+            let message_approval = match legacy_message_approval {
```
milapsheth, line 52:
```suggestion
            let message_approval = match legacy_storage::try_message_approval(env, message_approval_key).unwrap_or(legacy_storage::MessageApprovalValue::NotApproved) {
```
avoid the need to handle the `NotApproved` case later

contracts/stellar-axelar-gateway/src/migrate.rs
```diff
@@ -0,0 +1,72 @@
...
+                    Some(storage::MessageApprovalValue::Executed)
+                }
+                legacy_storage::MessageApprovalValue::NotApproved => None,
+            };
+
+            storage::set_message_approval(
+                env,
+                source_chain,
+                message_id,
+                &message_approval.ok_or(ContractError::InvalidMessageApproval)?,
```
milapsheth, line 66:
```suggestion
                message_approval,
```
after the above change, this should never error

contracts/stellar-axelar-gateway/src/tests/migrate.rs
```diff
@@ -0,0 +1,236 @@
...
+    });
+
+    assert_auth!(owner, client.upgrade(&new_wasm_hash));
+
+    let migration_data = vec![&env, (source_chain, message_id)];
+
+    env.as_contract(&client.address, || {
+        let result = AxelarGateway::__migrate(&env, migration_data);
+        assert!(result.is_ok());
+    });
```
milapsheth, line 45: the upgrader client could be used here in the future

milapsheth (approved):

contracts/stellar-axelar-gateway/src/migrate.rs
milapsheth: yeah, the goal was to not handle it again after the match case

PR #277, opened 2025-03-12
feat(multicall): create multicall contract

TanvirDeol:
[AXE-8040](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-8040)

[AXE-8040]: https://axelarnetwork.atlassian.net/browse/AXE-8040?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/stellar-multicall/src/interface.rs
```diff
@@ -0,0 +1,24 @@
...
+    /// # Returns
+    /// - `Ok(Vec<Val>)`: Returns a vector with the return data of each function call
+    ///
+    /// # Errors
+    /// - [`ContractError::MulticallFailed`]: If any of the contract calls fail.
+    fn multicall(
+        env: Env,
+        contract_address: Address,
+        funcs: Vec<Symbol>,
+        args: Vec<Val>,
```
milapsheth, line 22: it doesn't need to multiple functions on the same contract. Allow different contracts to be provided. You can create a type that encapsulates the 3 values and have a vector of it

contracts/stellar-multicall/src/interface.rs
```diff
@@ -0,0 +1,24 @@
...
+    ///
+    /// # Returns
+    /// - `Ok(Vec<Val>)`: Returns a vector with the return data of each function call
+    ///
+    /// # Errors
+    /// - [`ContractError::MulticallFailed`]: If any of the contract calls fail.
+    fn multicall(
+        env: Env,
+        contract_address: Address,
+        funcs: Vec<Symbol>,
```
milapsheth, line 21:
```suggestion
        functions: Vec<Symbol>,
```

contracts/stellar-multicall/src/interface.rs
milapsheth, line 22:
```suggestion
        args: Vec<Vec<Val>>,
```
each function takes a vec of args

contracts/stellar-multicall/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "stellar-multicall"
```
ahramy, line 2: If this package is under `contracts` folder, this will be included to the release. Do we want to have this contract released to crate.io and CF R2? otherwise we need to update `.release-plz.toml` to exclude this contract.

contracts/stellar-multicall/src/contract.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[contract]
+pub struct Multicall;
+
+#[contractimpl]
+impl MulticallInterface for Multicall {
+    fn multicall(env: Env, params: Vec<MulticallData>) -> Result<Vec<Val>, ContractError> {
+        let mut results = Vec::new(&env);
+
+        for i in 0..params.len() {
```
ahramy, line 16: It seems we can optimize this code using `params.iter()` Could you try it?

contracts/stellar-multicall/src/lib.rs
```diff
@@ -0,0 +1,9 @@
+#![no_std]
+
+pub mod error;
+
+pub mod interface;
+
+pub mod contract;
+
+pub mod types;
```
ahramy, line 9: we should add test around authentication to make sure this approach works with different type of functions. i.e. upgrade contract call from multicall will work?

contracts/stellar-multicall/Cargo.toml
milapsheth: yes, it'll be deployed. similar to example, we'll release any contract that gets deployed

contracts/stellar-multicall/src/contract.rs
```diff
@@ -0,0 +1,33 @@
+use soroban_sdk::{contract, contractimpl, Env, Val, Vec};
+use stellar_axelar_std::assert_some;
+
+use crate::error::ContractError;
+use crate::interface::MulticallInterface;
+use crate::types::MulticallData;
+
+#[contract]
+pub struct Multicall;
```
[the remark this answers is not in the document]
milapsheth: it's just a util contract, we can just redeploy when needed to avoid having to think about permissioning for it. similar to the upgrader

contracts/stellar-multicall/src/contract.rs
milapsheth, line 16:
```suggestion
        for MulticallData { contract_address, function, args } in params.into_iter()  {
```
use idiomatic Rust. take a look at other contracts/rust book/ask claude to think about making it cleaner

contracts/stellar-multicall/src/contract.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#[contractimpl]
+impl MulticallInterface for Multicall {
+    fn multicall(env: Env, params: Vec<MulticallData>) -> Result<Vec<Val>, ContractError> {
+        let mut results = Vec::new(&env);
+
+        for i in 0..params.len() {
+            let param = assert_some!(params.get(i));
```
milapsheth, line 17: if it looks ugly, then there's always a better way to do it in Rust

contracts/stellar-multicall/src/contract.rs
```diff
@@ -0,0 +1,33 @@
...
+            let contract_address = &param.contract_address;
+            let function = &param.function;
+            let args = &param.args;
+
+            let result: Result<Val, ContractError> =
+                env.invoke_contract(contract_address, function, args.clone());
```
milapsheth, line 23:
```suggestion
            let result: Val =
                env.invoke_contract(contract_address, function, args.clone());
```
`invoke_contract` doesn't return a result, but fails if an error is returned. see the upgrader contract, and other usage

contracts/stellar-multicall/src/types.rs
```diff
@@ -0,0 +1,9 @@
+use soroban_sdk::{contracttype, Address, Symbol, Val, Vec};
+
+#[contracttype]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct MulticallData {
+    pub contract_address: Address,
+    pub function: Symbol,
+    pub args: Vec<Val>,
+}
```
milapsheth, line 9:
```suggestion
pub struct FunctionCall {
    pub contract: Address,
    pub function: Symbol,
    pub args: Vec<Val>,
}
```
this is a single function call, not a multicall. And avoid using generic terms like data, params for names

contracts/stellar-multicall/Cargo.toml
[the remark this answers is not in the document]
TanvirDeol: Since Milap mentioned this is just a util contract like upgrader, it'd be fine to leave it out right?

contracts/stellar-multicall/src/contract.rs
```diff
@@ -0,0 +1,33 @@
...
+
+            match result {
+                Ok(val) => results.push_back(val),
+                Err(_) => return Err(ContractError::MulticallFailed),
```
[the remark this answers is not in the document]
TanvirDeol: I changed the code to propagate errors from the contract function we want to call. It is demonstrated in the tests now (`multicall_fails_when_target_returns_error`)

contracts/stellar-multicall/src/types.rs
```diff
@@ -0,0 +1,9 @@
+use soroban_sdk::{contracttype, Address, Symbol, Val, Vec};
+
+#[contracttype]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct FunctionCall {
+    pub contract: Address,
```
milapsheth, line 6:
```suggestion
    pub approver: Address,
    pub contract: Address,
```
for each function call, `approver.require_auth()` should be called so that an attacker can't frontrun the individual function calls. `approver` can be set to the caller, even if the function doesn't need an approval

contracts/stellar-multicall/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
```
milapsheth, line 1: The root `Cargo.toml` needs to be updated to mention this crate

contracts/stellar-multicall/src/interface.rs
```diff
@@ -0,0 +1,16 @@
...
+#[contractclient(name = "MulticallClient")]
+pub trait MulticallInterface {
+    /// Executes an arbitrary list of contract calls and returns the results of all the calls.
+    ///
+    /// # Arguments
+    /// * `params` - A list of params containing the contract address, function name and arguments for each contract call.
+    ///
+    /// # Returns
+    /// - `Ok(Vec<Val>)`: Returns a vector with the return data of each function call
+    fn multicall(env: &Env, function_calls: Vec<FunctionCall>) -> Result<Vec<Val>, ContractError>;
```
milapsheth, line 15: There is no error being returned from the contract
```suggestion
    fn multicall(env: &Env, function_calls: Vec<FunctionCall>) -> Vec<Val>;
```

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -0,0 +1,248 @@
...
+pub struct TestTarget;
+
+#[derive(Debug, PartialEq, Eq, IntoEvent)]
+pub struct ExecutedEvent {
+    pub value: u32,
+}
+
+#[contractimpl]
+impl OwnableInterface for TestTarget {
+    fn owner(env: &Env) -> Address {
```
milapsheth, line 26: why is this implemented manually? Use our derive macros. See how we do it for other contracts

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -0,0 +1,248 @@
...
+
+    let function_calls = vec![
+        &env,
+        FunctionCall {
+            contract: target.clone(),
+            function: symbol_short!("method"),
+            args: vec![&env, IntoVal::<_, Val>::into_val(&42u32, &env)],
+        },
+        FunctionCall {
+            contract: target,
```
milapsheth, line 77: add more coverage. test multicalls with different targets

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -0,0 +1,248 @@
...
+            function: symbol_short!("owner"),
+            args: vec![&env],
+        },
+    ];
+
+    client.multicall(&function_calls);
+    goldie::assert!(fmt_last_emitted_event::<ExecutedEvent>(&env));
+}
+
```
milapsheth, line 86: test zero function call case

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -0,0 +1,248 @@
...
+fn multicall_auth_succeeds() {
+    let (env, client, target, current_owner) = setup();
+    let new_owner = Address::generate(&env);
+
+    let function_calls = vec![
+        &env,
+        FunctionCall {
+            contract: target.clone(),
+            function: symbol_short!("method"),
+            args: vec![&env, IntoVal::<_, Val>::into_val(&42u32, &env)],
```
milapsheth, line 110: create a simple macro in this test file similar to `assert_auth!` where you give it `contract.function(args)` and it creates a `FunctionCall` struct from it. Makes the test more readable

contracts/stellar-multicall/src/contract.rs
```diff
@@ -0,0 +1,33 @@
...
+    fn multicall(env: &Env, function_calls: Vec<FunctionCall>) -> Result<Vec<Val>, ContractError> {
+        let mut results = Vec::new(env);
+
+        for FunctionCall {
+            contract,
+            function,
+            args,
+        } in function_calls.into_iter()
+        {
+            let result: Val = env.invoke_contract(&contract, &function, args.clone());
```
milapsheth, line 26: multicall should support invoking methods that call `require_auth` on some `approver`. Explicitly add `approver.require_auth()` here so sub-invocation authorization is nested

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -0,0 +1,248 @@
...
+            args: vec![&env],
+        },
+    ];
+
+    client.multicall(&function_calls);
+    goldie::assert!(fmt_last_emitted_event::<ExecutedEvent>(&env));
+}
+
+#[test]
+fn transfer_ownership_auth() {
```
milapsheth, line 88: this test is irrelevant, it's calling the test contract directly without a multicall

contracts/stellar-multicall/Cargo.toml
```diff
@@ -15,6 +15,10 @@ stellar-axelar-std = { workspace = true }
 goldie = { workspace = true }
 soroban-sdk = { workspace = true, features = ["testutils"] }
 stellar-axelar-std = { workspace = true, features = ["testutils"] }
+stellar-axelar-operators = { workspace = true }
+stellar-interchain-token = { workspace = true, features = ["testutils"] }
+soroban-token-sdk = { workspace = true }
+stellar-axelar-gas-service = { workspace = true, features = ["testutils"]}
```
milapsheth: we should use test contracts for unit testing. what's the reason for using gas service etc.? The unit test should do the minimal necessary to make the test work. Also, if we make changes to the gas service, suddenly the multicall tests can break which is unexpected.

contracts/stellar-multicall/src/contract.rs
```diff
@@ -19,10 +19,12 @@ impl MulticallInterface for Multicall {
 
         for FunctionCall {
             contract,
+            approver,
             function,
             args,
         } in function_calls.into_iter()
         {
+            approver.require_auth();
             let result: Val = env.invoke_contract(&contract, &function, args.clone());
```
milapsheth:
```suggestion
            approver.require_auth();

            let result: Val = env.invoke_contract(&contract, &function, args.clone());
```

contracts/stellar-multicall/src/types.rs
```diff
@@ -4,6 +4,7 @@ use soroban_sdk::{contracttype, Address, Symbol, Val, Vec};
 #[derive(Clone, Debug, PartialEq, Eq)]
 pub struct FunctionCall {
     pub contract: Address,
+    pub approver: Address,
```
milapsheth, line 7:
```suggestion
    /// There must be an address authorizing each function call, even if the function call itself doesn't need authorization, to prevent frontrunning of individual calls within a multicall
    pub approver: Address,
```

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -51,198 +59,413 @@ impl TestTarget {
...
+            target_id,
+            owner,
+            method(IntoVal::<_, Val>::into_val(&42u32, &env))
+        ),
+        construct_function_call!(
+            env,
+            target_id,
+            owner,
+            transfer_ownership(new_owner.to_val())
+        ),
```
milapsheth, line 279:
```suggestion
        function_call!(owner, client.transfer_ownership(new_owner)),
```
why can't this be made simpler just like `assert_auth!`?

https://github.com/axelarnetwork/axelar-cgp-stellar/blob/main/packages/stellar-axelar-std/src/error.rs#L171

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -1,37 +1,45 @@
...
+use soroban_sdk::testutils::{Address as _, BytesN as _};
 use soroban_sdk::{
-    contract, contractimpl, symbol_short, vec, Address, Env, IntoVal, Symbol, Val, Vec,
+    bytes, contract, contractimpl, vec, Address, Bytes, BytesN, Env, IntoVal, String, Val, Vec,
 };
+use soroban_token_sdk::metadata::TokenMetadata;
+use stellar_axelar_gas_service::event::GasPaidEvent;
+use stellar_axelar_gas_service::testutils::setup_gas_token;
+use stellar_axelar_gas_service::{AxelarGasService, AxelarGasServiceClient};
+use stellar_axelar_operators::{AxelarOperators, AxelarOperatorsClient};
```
milapsheth, line 12: why are these being used? just created test contracts with minimal code needed to test the scenario to keep it simple

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -51,198 +59,413 @@ impl TestTarget {
...
+
     assert_contract_err!(
-        client.try_multicall(&function_calls),
+        client
+            .mock_auths(&[
+                multicall_auth.clone(),
+                multicall_auth.clone(),
+                multicall_auth
+            ])
+            .try_multicall(&function_calls),
```
milapsheth, line 468:
```suggestion
            .mock_all_auths()
            .try_multicall(&function_calls),
```
When testing for a non-auth error, you can mock all auths for simplicity

contracts/stellar-multicall/src/tests/test.rs
TanvirDeol: I applied the fixed as you described, however I had to leave in the `target_id` (contract ID) as a param because the `FunctionCall` needs it, and I don't think the contract ID is accessible from client

contracts/stellar-multicall/src/tests/test.rs
TanvirDeol: `mock_all_auths()` works for one function call at a time, so in this case with 3 function calls the authorization is incomplete. However, for all tests w/ only 1 function call this works so I applied it there

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -0,0 +1,302 @@
...
+
+    let subscriber_id = env.register(TestSubscriptionContract, (owner.clone(),));
+    let subscriber_client = TestSubscriptionContractClient::new(&env, &subscriber_id);
+
+    let function_calls = vec![
+        &env,
+        construct_function_call!(bank_id, owner, bank_client.deposit(42u32)),
+        construct_function_call!(bank_id, owner, bank_client.withdraw(10u32)),
+        construct_function_call!(target_id, owner, client.method(0u32)),
+        construct_function_call!(target_id, owner, client.owner()),
```
milapsheth, line 99: a test with different approvers is needed

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -0,0 +1,302 @@
...
+#[test]
+fn multicall_succeeds() {
+    let TestConfig {
+        env,
+        client,
+        target_id,
+        owner,
+    } = setup();
+
+    let bank_id = env.register(TestBankContract, (owner.clone(),));
```
milapsheth, line 88: if it's small enough, this contract could be defined in this file since it's not used anywhere else

contracts/stellar-multicall/src/tests/test.rs
milapsheth: That doesn't quite make sense. What's the reason for it? Take a deeper look, you can talk to others in the team, or ask the Stellar team

contracts/stellar-multicall/src/tests/test.rs
milapsheth: `client.address`. look at the implementation of `assert_auth!` again, it doesn't have this problem

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -25,34 +22,107 @@ macro_rules! construct_function_call {
...
+            let owner = Self::owner(env);
+            owner.require_auth();
+
+            let current_balance: u32 = env
+                .storage()
+                .instance()
+                .get(&Self::BALANCE_KEY)
+                .unwrap_or(0u32);
+            if current_balance < amount {
+                return Err(ContractError::FunctionCallFailed);
```
milapsheth, line 76: define an error specific to this contract, instead of reusing the crate's error type

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -25,34 +22,107 @@ macro_rules! construct_function_call {
...
+        pub fn method(env: &Env, value: u32) {
+            ExecutedEvent { value }.emit(env);
+        }
+
+        pub fn failing(_env: &Env) {
+            panic!("This method should fail");
+        }
+
+        pub const fn failing_with_error(_env: &Env) -> Result<Val, ContractError> {
+            Err(ContractError::FunctionCallFailed)
```
milapsheth, line 118: same

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -2,19 +2,16 @@
...
-use crate::tests::subscriber_contract::{TestSubscriptionContract, TestSubscriptionContractClient};
 use crate::types::FunctionCall;
 use crate::{Multicall, MulticallClient};
 
 #[macro_export]
-macro_rules! construct_function_call {
+macro_rules! function_call {
     ($contract:expr, $approver:expr, $client:ident . $function:ident ( $($arg:expr),* $(,)? )) => {{
         FunctionCall {
             contract: $contract.clone(),
```
milapsheth, line 17:
```suggestion
            contract: $client.address.clone(),
```

contracts/stellar-multicall/src/tests/test.rs
```diff
@@ -2,19 +2,16 @@
...
 use crate::error::ContractError;
-use crate::tests::bank_contract::{TestBankContract, TestBankContractClient};
-use crate::tests::subscriber_contract::{TestSubscriptionContract, TestSubscriptionContractClient};
 use crate::types::FunctionCall;
 use crate::{Multicall, MulticallClient};
 
 #[macro_export]
-macro_rules! construct_function_call {
+macro_rules! function_call {
     ($contract:expr, $approver:expr, $client:ident . $function:ident ( $($arg:expr),* $(,)? )) => {{
```
milapsheth, line 15:
```suggestion
    ($approver:expr, $client:ident . $function:ident ( $($arg:expr),* $(,)? )) => {{
```
this is not needed

contracts/stellar-multicall/src/tests/test.rs
TanvirDeol: I addressed this, however I wasn't able to use `assert_err` or `assert_contract_err` macros since those wanted a `ContractErr`, so instead I used `should_panic`

contracts/stellar-multicall/src/tests/test.rs
TanvirDeol: Talked to the stellar team, and I think the best option is just to leave it as is https://interoplabs.slack.com/archives/C04545F3S59/p1743451741779649?thread_ts=1743450804.674979&cid=C04545F3S59

milapsheth (approved):

PR #279, opened 2025-03-13
feat: block regular contract endpoints during migration

cgorenflo:


packages/stellar-axelar-std/src/tests/mod.rs
```diff
@@ -1,4 +1,7 @@
 mod contractstorage;
 mod event;
+mod operatable;
+mod ownable;
+mod pausable;
```
cgorenflo, line 5: these were missing from test runs

packages/stellar-axelar-std-derive/src/modifier.rs
```diff
@@ -2,7 +2,7 @@ use proc_macro2::TokenStream as TokenStream2;
 use quote::quote;
 use syn::ItemFn;
 
-pub fn modifier_impl(input_fn: ItemFn, auth_statement: TokenStream2) -> TokenStream2 {
+pub fn modifier_impl(input_fn: &ItemFn, auth_statement: TokenStream2) -> TokenStream2 {
```
cgorenflo, line 5: I will be able to merge this helper with how I add the migration check

contracts/stellar-axelar-example/src/contract.rs
```diff
@@ -27,7 +25,6 @@ pub enum AxelarExampleError {
     InvalidAmount = 3,
 }
 
-#[contractimpl]
```
cgorenflo, line 30: This should not have had the annotation, because the `CustomAxelarExecutable` shouldn't be part of the contract API

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -0,0 +1,132 @@
...
+            let env_ident = parse_env_identifier(&method.sig.inputs)?;
+            let error_handling = if can_return_contract_error(&method.sig.output) {
+                return_migration_in_progress()
+            } else {
+                panic_on_failure()
+            };
+
+            method.prepend_statement(migration_in_progess_check(env_ident, error_handling));
+            Ok(())
+        })?;
```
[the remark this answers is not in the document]
cgorenflo: the issue with inlining the trait check is that it's going to be repeated for each function of the trait, instead of just once for the whole block. But I'll try to find a more straightforward solution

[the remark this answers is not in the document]
cgorenflo: I think I found a good solution for improvement. I added comments in the code where inlining is impossible

PR #280, opened 2025-03-18
refactor(interchain-token-service): migrate token-manager, interchain-token & destruct flow keys

nbayindirli:
AXE-8123

### DIFF
--- a/contracts/stellar-interchain-token-service/src/tests/utils/mod.rs
+++ b/contracts/stellar-interchain-token-service/src/tests/utils/mod.rs
@@ -53,28 +53,28 @@ pub fn format_auths(
     auths: std::vec::Vec<(Address, AuthorizedInvocation)>,
     description: &str,
 ) -> String {
-    let mut formatted_auths = format!("{:?}:\n", description.to_string());
+    let mut result = format!("{description:?}:\n");
 
     for (caller, invocation) in auths {
         let (client, method, args) = match invocation.function {
-            AuthorizedFunction::Contract((client, method, args)) => (client, method, args),
+            AuthorizedFunction::Contract(data) => data,
             _ => panic!("Expected a contract function"),
         };
-        formatted_auths.push_str(&format!("    caller: {:?}\n", caller));
-        formatted_auths.push_str(&format!(
-            "        invocation: {:?}.{:?}({:?})\n",
+
+        result.push_str(&format!("    caller: {caller:?}\n"));
+        result.push_str(&format!(
+            "        invocation: {:?}.{:?}({args:?})\n",
             client.to_string(),
-            method.to_string(),
-            args
+            method.to_string()
         ));
 
         for sub_invocation in invocation.sub_invocations {
-            formatted_auths.push_str(&format_auths(
+            result.push_str(&format_auths(
                 vec![(caller.clone(), sub_invocation)],
                 "sub_invocation",
             ));
         }
     }
 
-    formatted_auths
+    result
 }


contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+        pub token_id: BytesN<32>,
+        pub epoch: u64,
+    }
+}
+
+impl CustomMigratableInterface for InterchainTokenService {
+    type MigrationData = (
+        BytesN<32>, /* new_wasm_hash */
+        Vec<(BytesN<32> /* token_id */, u64 /* epoch */)>,
+    );
```
milapsheth, line 76: define a custom contract type instead?

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+        #[value(Address)]
+        Gateway,
+
+        #[instance]
+        #[value(Address)]
+        GasService,
+
+        #[instance]
+        #[value(String)]
+        ChainName,
```
milapsheth, line 25: shall we keep only the variants that are used around? pro: more readable, better coverage. cons: not just a copy paste of the data key, accidental edits

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    fn __migrate(env: &Env, migration_data: Self::MigrationData) -> Result<(), Self::Error> {
+        let (new_wasm_hash, token_ids_and_epochs) = migration_data;
+
+        for (token_id, epoch) in token_ids_and_epochs {
+            let token_id_config = storage::try_token_id_config(env, token_id.clone())
+                .ok_or(ContractError::InvalidTokenId)?;
+
+            Self::migrate_token_manager(env, token_id_config.token_manager, &new_wasm_hash)?;
+            Self::migrate_interchain_token(env, &token_id, &new_wasm_hash)?;
```
milapsheth, line 87: same contract wasm hash for both token manager and interchain token?

contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -35,6 +35,7 @@ const EXECUTE_WITH_INTERCHAIN_TOKEN: &str = "execute_with_interchain_token";
```
milapsheth, line 35: gh actions are failing

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+        pub epoch: u64,
+    }
+}
+
+impl CustomMigratableInterface for InterchainTokenService {
+    type MigrationData = (
+        BytesN<32>, /* new_wasm_hash */
+        Vec<(BytesN<32> /* token_id */, u64 /* epoch */)>,
+    );
+    type Error = ContractError;
```
milapsheth, line 77: Shall we define a separate `MigrateError` type? Migration error variants can only be raised during migration, and will never be used in normal contract calls, so adding them to `ContractError` doesn't seem quite right. Those variants will also become unused once the migration handler is removed in the next version

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+            Self::migrate_token_manager(env, token_id_config.token_manager, &new_wasm_hash)?;
+            Self::migrate_interchain_token(env, &token_id, &new_wasm_hash)?;
+            Self::migrate_flow_key(env, &token_id, epoch)?;
+        }
+
+        Ok(())
+    }
+}
+
+impl InterchainTokenService {
```
milapsheth, line 95: Don't think this should be implemented on the contract type, since they'll get exposed to other functions that can call them?

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+
+        Ok(())
+    }
+
+    fn migrate_interchain_token(
+        env: &Env,
+        token_id: &BytesN<32>,
+        new_wasm_hash: &BytesN<32>,
+    ) -> Result<(), ContractError> {
+        let interchain_token_address = deployer::interchain_token_address(env, token_id.clone());
```
milapsheth, line 113: incorrect, use the token address from the token id config. This entrypoint is only for predicting a token address deployed via ITS

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+
+impl InterchainTokenService {
+    fn migrate_token_manager(
+        env: &Env,
+        token_manager_address: Address,
+        new_wasm_hash: &BytesN<32>,
+    ) -> Result<(), ContractError> {
+        UpgradableClient::new(&env, &token_manager_address).upgrade(new_wasm_hash);
+
+        storage::set_token_manager_wasm_hash(env, new_wasm_hash);
```
milapsheth, line 103: why is this being set multiple times? It's not token_id specific

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+    type MigrationData = (
+        BytesN<32>, /* new_wasm_hash */
+        Vec<(BytesN<32> /* token_id */, u64 /* epoch */)>,
+    );
+    type Error = ContractError;
+
+    fn __migrate(env: &Env, migration_data: Self::MigrationData) -> Result<(), Self::Error> {
+        let (new_wasm_hash, token_ids_and_epochs) = migration_data;
+
+        for (token_id, epoch) in token_ids_and_epochs {
```
milapsheth, line 82:
```suggestion
        for (token_id, epoch) in token_ids_and_epochs.into_iter() {
```
to take by value

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+    pub struct FlowKey {
+        pub token_id: BytesN<32>,
+        pub epoch: u64,
+    }
+}
+
+impl CustomMigratableInterface for InterchainTokenService {
+    type MigrationData = (
+        BytesN<32>, /* new_wasm_hash */
+        Vec<(BytesN<32> /* token_id */, u64 /* epoch */)>,
```
milapsheth, line 75: only the current epoch needs to be migrated. The old epochs get automatically deleted due to the use of temporary storage, and are never accessed once the epoch changes

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+    fn __migrate(env: &Env, migration_data: Self::MigrationData) -> Result<(), Self::Error> {
+        let (new_wasm_hash, token_ids_and_epochs) = migration_data;
+
+        for (token_id, epoch) in token_ids_and_epochs {
+            let token_id_config = storage::try_token_id_config(env, token_id.clone())
+                .ok_or(ContractError::InvalidTokenId)?;
+
+            Self::migrate_token_manager(env, token_id_config.token_manager, &new_wasm_hash)?;
+            Self::migrate_interchain_token(env, &token_id, &new_wasm_hash)?;
+            Self::migrate_flow_key(env, &token_id, epoch)?;
```
milapsheth, line 88:
```suggestion
            Self::migrate_flow(env, &token_id, epoch)?;
```
since the enum variant is called `Flow`

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+impl InterchainTokenService {
+    fn migrate_token_manager(
+        env: &Env,
+        token_manager_address: Address,
+        new_wasm_hash: &BytesN<32>,
+    ) -> Result<(), ContractError> {
+        UpgradableClient::new(&env, &token_manager_address).upgrade(new_wasm_hash);
```
milapsheth, line 101: Calling `upgrade` isn't enough, `migrate` needs to be called as well to complete the upgrade. You can also use the atomic upgrader client instead so it does both of it for you

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+    type Error = ContractError;
+
+    fn __migrate(env: &Env, migration_data: Self::MigrationData) -> Result<(), Self::Error> {
+        let (new_wasm_hash, token_ids_and_epochs) = migration_data;
+
+        for (token_id, epoch) in token_ids_and_epochs {
+            let token_id_config = storage::try_token_id_config(env, token_id.clone())
+                .ok_or(ContractError::InvalidTokenId)?;
+
+            Self::migrate_token_manager(env, token_id_config.token_manager, &new_wasm_hash)?;
```
ahramy, line 86: should take different wasm hash for token manager and interchain token separately

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,138 @@
...
+
+    #[contracttype]
+    #[derive(Clone, Debug)]
+    pub struct FlowKey {
+        pub token_id: BytesN<32>,
+        pub epoch: u64,
+    }
+}
+
+impl CustomMigratableInterface for InterchainTokenService {
```
ahramy, line 72: It would be great to have a tested example in the README.md of the contract deployment script repository, prepared for this migration.

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -0,0 +1,276 @@
...
+        legacy_storage::set_flow_out(&env, flow_key, &flow_out_amount);
+
+        storage::set_token_id_config(&env, token_id.clone(), &token_config);
+        storage::set_token_manager_wasm_hash(&env, &new_token_manager_wasm_hash);
+        storage::set_interchain_token_wasm_hash(&env, &new_interchain_token_wasm_hash);
+    });
+
+    assert_auth!(owner, client.upgrade(&new_its_wasm_hash));
+
+    let migration_data = (
```
ahramy, line 63: `new_token_manager_wasm_hash` and `new_interchain_token_wasm_hash` should be passed as migration data to the migrate function

contracts/stellar-interchain-token-service/src/types.rs
```diff
@@ -50,3 +50,12 @@ pub enum TokenManagerType {
     // LockUnlockFee = 3,
     // MintBurn = 4,
 }
+
+#[contracttype]
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct CustomMigrationData {
```
milapsheth, line 56: this should live in `migrate.rs` since it's only used during migration

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -70,68 +30,44 @@ pub mod legacy_storage {
...
-    ) -> Result<(), ContractError> {
-        UpgradableClient::new(&env, &token_manager_address).upgrade(new_wasm_hash);
-
-        storage::set_token_manager_wasm_hash(env, new_wasm_hash);
-
-        Ok(())
-    }
+            let flow_key = legacy_storage::FlowKey {
+                token_id: token_id.clone(),
+                epoch: current_epoch,
```
milapsheth, line 57: compute the `current_epoch` via the helper instead of passing it as an arg. It's more robust this way.

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -70,68 +30,44 @@ pub mod legacy_storage {
...
 
-    fn migrate_interchain_token(
-        env: &Env,
-        token_id: &BytesN<32>,
-        new_wasm_hash: &BytesN<32>,
-    ) -> Result<(), ContractError> {
-        let interchain_token_address = deployer::interchain_token_address(env, token_id.clone());
+            let flow_out = legacy_storage::try_flow_out(env, flow_key.clone())
+                .ok_or(ContractError::InvalidFlowKey)?;
+            let flow_in = legacy_storage::try_flow_in(env, flow_key.clone())
```
milapsheth, line 62: move flow_in before flow_out for consistency?

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,74 @@
...
+            UpgradableClient::new(env, &token_manager).upgrade(&new_token_manager_wasm_hash);
+            UpgradableClient::new(env, &interchain_token).upgrade(&new_interchain_token_wasm_hash);
+
+            let flow_key = legacy_storage::FlowKey {
+                token_id: token_id.clone(),
+                epoch: current_epoch,
+            };
+
+            let flow_out = legacy_storage::try_flow_out(env, flow_key.clone())
+                .ok_or(ContractError::InvalidFlowKey)?;
```
milapsheth, line 61: A flow limit might not be set for the token_id and thus no flow is tracked, so an error shouldn't be thrown

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,74 @@
...
+        let CustomMigrationData {
+            new_token_manager_wasm_hash,
+            new_interchain_token_wasm_hash,
+            token_ids,
+            current_epoch,
+        } = migration_data;
+
+        for token_id in token_ids.into_iter() {
+            let TokenIdConfigValue {
+                token_address: interchain_token,
```
milapsheth, line 46: This might be an existing token/asset, such as XLM. In this case, the interchain token contract can't/shouldn't be upgraded, only the token manager. Check the token manager type to determine if it should be upgraded.

milapsheth (dismissed):

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,101 @@
...
+}
+
+pub mod legacy_storage {
+    use stellar_axelar_std::{contractstorage, contracttype, soroban_sdk, BytesN};
+
+    #[contractstorage]
+    enum LegacyDataKey {
+        #[temporary]
+        #[value(i128)]
+        FlowOut { flow_key: FlowKey },
```
milapsheth, line 28: In hindsight, it would have been better to migrate flow limit in a separate PR since it can be merged independently (and likely quicker) and make this one less complex

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,101 @@
...
+                epoch: current_epoch,
+            };
+
+            if let Some(flow_out) = legacy_storage::try_flow_out(env, flow_key.clone()) {
+                storage::set_flow_out(env, token_id.clone(), current_epoch, &flow_out);
+            }
+            if let Some(flow_in) = legacy_storage::try_flow_in(env, flow_key.clone()) {
+                storage::set_flow_in(env, token_id.clone(), current_epoch, &flow_in);
+            }
+        }
```
milapsheth, line 97: Stellar has pretty stringent read/write limit per txn
https://developers.stellar.org/docs/networks/resource-limits-fees

I'm concerned that we'll hit it with just a few token ids. A better approach is to simplify the atomic migration to just update the wasm hash so new tokens use the correct version, and defining a separate entrypoint `migrate_token_id` that migrates an existing token_id. There is no breaking interface between v1.0 and v1.1 I believe. This way, `migrate_token_id` can be called for each `token_id` without hitting any resource limits (gas, read/write, txn size etc.)

contracts/stellar-interchain-token-service/src/tests/interchain_transfer.rs
```diff
@@ -158,7 +158,7 @@ fn interchain_transfer_send_fails_when_paused() {
 }
 
 #[test]
-#[should_panic(expected = "burn, Error(Contract, #9)")]
+#[should_panic(expected = "burn, Error(Contract, #6)")] // InvalidPayload
```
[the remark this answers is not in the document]
milapsheth: Not sure I understand the plan. Since `required_auths` isn't available, the upgrader address provided during migration will need to be the `v1.0` version that doesn't depend on `required_auths`, right?

contracts/stellar-interchain-token/src/contract.rs
```diff
@@ -41,21 +41,6 @@ impl InterchainToken {
...
-    fn owner(env: &Env) -> Address {
-        interfaces::owner(env)
-    }
-
-    fn transfer_ownership(env: &Env, new_owner: Address) {
-        let old_owner = Self::owner(env);
-
-        interfaces::transfer_ownership::<Self>(env, new_owner.clone());
-
-        TokenEvents::new(env).set_admin(old_owner, new_owner);
```
milapsheth, line 55: Actually, this was being implemented manually since it emits an extra event that the default implementation doesn't do (to stay compatible with the token standard)

Add a notice explaining this so we don't remove it accidentally, and the event unit test should test this event to catch changes

contracts/stellar-interchain-token/src/tests/test.rs
```diff
@@ -91,9 +91,7 @@ fn transfer_ownership_succeeds() {
     let (token, _) = setup_token(&env);
 
     assert_auth!(token.owner(), token.transfer_ownership(&new_owner));
-    goldie::assert!(fmt_emitted_event_at_idx::<OwnershipTransferredEvent>(
-        &env, -2
-    ));
+    goldie::assert!(fmt_last_emitted_event::<OwnershipTransferredEvent>(&env));
```
milapsheth, line 94: breaking event change, be careful when you need to "fix" golden files once a contract is stabilized

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,101 @@
...
+                .ok_or(ContractError::InvalidTokenId)?;
+
+            upgrader_client.upgrade(
+                &token_manager,
+                &new_version,
+                &new_token_manager_wasm_hash,
+                &stellar_axelar_std::vec![env, ().into()],
+            );
+
+            if token_manager_type != TokenManagerType::LockUnlock {
```
milapsheth, line 77:
```suggestion
            if token_manager_type == TokenManagerType::NativeInterchainToken {
```
only this token manager type corresponds to a token deployed by ITS. add a comment to explain this behaviour

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -0,0 +1,101 @@
...
+                token_manager,
+                token_manager_type,
+            } = storage::try_token_id_config(env, token_id.clone())
+                .ok_or(ContractError::InvalidTokenId)?;
+
+            upgrader_client.upgrade(
+                &token_manager,
+                &new_version,
+                &new_token_manager_wasm_hash,
+                &stellar_axelar_std::vec![env, ().into()],
```
milapsheth, line 74:
```suggestion
                &vec![env, ().into()],
```
simplify?

contracts/stellar-interchain-token-service/src/testutils.rs
```diff
@@ -9,9 +9,9 @@ use crate::{InterchainTokenService, InterchainTokenServiceClient};
 // Note: On changes to `interchain-token` and `token-manager` crates, recompile via `stellar contract build && ./optimize.sh`
 // and copy the built `target/wasm32-unknown-unknown/release/stellar_*.optimized.wasm` to ../testdata.
 pub const INTERCHAIN_TOKEN_WASM: &[u8] =
-    include_bytes!("./testdata/stellar_interchain_token.optimized.wasm");
+    include_bytes!("./testdata/stellar_interchain_token_v100.optimized.wasm");
```
milapsheth, line 12:
```suggestion
    include_bytes!("./testdata/stellar_interchain_token_v1_0_0.optimized.wasm");
```

contracts/stellar-interchain-token-service/src/testutils.rs
```diff
@@ -9,9 +9,9 @@ use crate::{InterchainTokenService, InterchainTokenServiceClient};
...
 pub const TOKEN_MANAGER_WASM: &[u8] =
-    include_bytes!("./testdata/stellar_token_manager.optimized.wasm");
+    include_bytes!("./testdata/stellar_token_manager_v100.optimized.wasm");
```
milapsheth, line 14:
```suggestion
    include_bytes!("./testdata/stellar_token_manager_v1_0_0.optimized.wasm");
```

contracts/stellar-interchain-token-service/src/migrate.rs
milapsheth, line 74: This interface doesn't actually make sense. `Upgradable` contracts only take one migration data arg. So the upgrader shouldn't be requiring a list of args I think, and the usage should be more like the following
```suggestion
                &().into(),
```

contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -311,6 +311,17 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
 
         Ok(())
     }
+
+    #[only_owner]
+    #[allow_during_migration]
```
milapsheth, line 316:
```suggestion
```
this is a normal entrypoint, it's not being called during upgrade but after it for each token

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -46,56 +43,60 @@ impl CustomMigratableInterface for InterchainTokenService {
...
-        }
-
         Ok(())
     }
 }
+
+pub fn migrate_token(
+    env: &Env,
+    token_id: BytesN<32>,
+    upgrader_client: Address,
```
milapsheth, line 60:
```suggestion
    upgrader: Address,
```
this is not a client

contracts/stellar-interchain-token-service/src/migrate.rs
```diff
@@ -46,56 +43,60 @@ impl CustomMigratableInterface for InterchainTokenService {
...
+    if token_manager_type == TokenManagerType::NativeInterchainToken {
+        upgrader_client.upgrade(
+            &interchain_token,
+            &new_version,
+            &storage::interchain_token_wasm_hash(env),
+            &vec![env, ().into()],
+        );
+    }
+
+    let flow_key = legacy_storage::FlowKey {
```
milapsheth, line 89:
```suggestion
    let current_epoch = current_epoch(env);
    let flow_key = legacy_storage::FlowKey {
```
move the definition to where it's needed

contracts/stellar-interchain-token-service/src/interface.rs
```diff
@@ -266,4 +266,19 @@ pub trait InterchainTokenServiceInterface:
         metadata: Option<Bytes>,
         gas_token: Option<Token>,
     ) -> Result<(), ContractError>;
+
+    /// Migrates a token to a new version.
```
milapsheth, line 270: Explain why this isn't done during upgrade

contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -311,6 +311,17 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
 
         Ok(())
     }
+
+    #[only_owner]
```
milapsheth, line 315: If we saved the upgrader in ITS, and used the stored version, technically we could even make this permissionless. There's not much benefit in this situation, so it's fine as-is, but worth considering in the future

contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -311,6 +311,17 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
...
+    #[allow_during_migration]
+    fn migrate_token(
```
milapsheth, line 317: Another interesting pattern to keep in mind, is that we could dynamically call this method during interchain_transfer to auto-upgrade any token_id that's not yet upgraded. We used this approach for EVM ITS upgrade where there's a lot of tokens already deployed, and the upgrade itself was breaking

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -17,9 +16,9 @@ use crate::InterchainTokenService;
 
 const ITS_WASM: &[u8] = include_bytes!("testdata/stellar_interchain_token_service.optimized.wasm");
 const TOKEN_MANAGER_WASM_V110: &[u8] =
-    include_bytes!("testdata/stellar_token_manager_v110.optimized.wasm");
+    include_bytes!("testdata/stellar_token_manager_v1_1_0.optimized.wasm");
 const INTERCHAIN_TOKEN_WASM_V110: &[u8] =
-    include_bytes!("testdata/stellar_interchain_token_v110.optimized.wasm");
+    include_bytes!("testdata/stellar_interchain_token_v1_1_0.optimized.wasm");
 
 const VERSION_V110: &str = "1.1.0";
```
milapsheth, line 23:
```suggestion
const NEW_VERSION: &str = "1.1.0";
```

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -78,6 +74,15 @@ fn migrate_native_interchain_token_succeeds() {
         assert_ok!(InterchainTokenService::__migrate(&env, migration_data));
     });
 
+    env.as_contract(&its_client.address, || {
```
milapsheth, line 77: avoid calling entrypoints using `env.as_contract` since it allows doing anything

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -0,0 +1,589 @@
...
+    let owner = its_client.owner();
+    let non_existent_token_id = BytesN::random(&env);
+
+    let its_wasm_hash = env.deployer().upload_contract_wasm(ITS_WASM);
+    let token_manager_wasm_hash_v110 = env.deployer().upload_contract_wasm(TOKEN_MANAGER_WASM_V110);
+    let interchain_token_wasm_hash_v110 = env
+        .deployer()
+        .upload_contract_wasm(INTERCHAIN_TOKEN_WASM_V110);
+
+    assert_auth!(owner, its_client.upgrade(&its_wasm_hash));
```
milapsheth, line 216: add a helper and remove duplication across tests?

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -0,0 +1,589 @@
...
+        new_interchain_token_wasm_hash: interchain_token_wasm_hash_v110,
+    };
+
+    env.as_contract(&its_client.address, || {
+        assert_ok!(InterchainTokenService::__migrate(&env, migration_data));
+    });
+
+    assert_contract_err!(
+        its_client
+            .mock_all_auths_allowing_non_root_auth()
```
milapsheth, line 229: if we see this used anywhere, we're likely introducing frontrunning vulnerability suggesting authorization isn't being handled properly

milapsheth: `migrate_token` only allows the `owner` to call it, so why is non root auth needed?

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -0,0 +1,589 @@
...
+    let token_manager_v100_2 = its_client.deployed_token_manager(&token_id_2);
+
+    let interchain_token_v100_1 = its_client.interchain_token_address(&token_id_1);
+    let interchain_token_v100_2 = its_client.interchain_token_address(&token_id_2);
+
+    let token_config_1 = TokenIdConfigValue {
+        token_address: interchain_token_v100_1,
+        token_manager: token_manager_v100_1,
+        token_manager_type: TokenManagerType::LockUnlock,
+    };
```
milapsheth, line 275: create a helper to set legacy token params and flow limit to reduce duplication in tests

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -0,0 +1,589 @@
...
+        assert_ok!(InterchainTokenService::migrate_token(
+            &env,
+            token_id.clone(),
+            upgrader_client.address,
+            String::from_str(&env, VERSION_V110),
+        ));
+    });
+
+    assert_eq!(
+        env.as_contract(&its_client.address, || {
```
milapsheth, line 465: same, entrypoint should be called. avoid use of `env.as_contract`

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -0,0 +1,589 @@
...
+    });
+
+    assert_auth!(owner, its_client.upgrade(&its_wasm_hash));
+
+    let migration_data = CustomMigrationData {
+        new_token_manager_wasm_hash: token_manager_wasm_hash_v110.clone(),
+        new_interchain_token_wasm_hash: interchain_token_wasm_hash_v110.clone(),
+    };
+
+    env.mock_all_auths_allowing_non_root_auth();
```
milapsheth, line 543: don't globally mock auths, very easy to overlook it in future PRs

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -285,305 +363,283 @@ fn migrate_succeeds_with_multiple_token_ids() {
...
+        );
+
+    its_client
+        // .mock_auths(&[its_migrate_token_auth]) // FIXME: Use
+        .mock_all_auths_allowing_non_root_auth()
+        .migrate_token(
+            &token_id_2,
+            &upgrader_client.address,
+            &String::from_str(&env, NEW_VERSION),
+        );
```
milapsheth, line 427: now that tokens aren't migrated atomically, this test for migrating multiple tokens doesn't really make sense. There should be separate tests for the two token manager types, but otherwise they're independent

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -285,305 +363,283 @@ fn migrate_succeeds_with_multiple_token_ids() {
...
+        .upgrade(
+            &its_client.address,
+            &String::from_str(&env, NEW_VERSION),
+            &its_wasm_hash,
+            &vec![&env, migration_data.into_val(&env)],
+        );
 }
 
 #[test]
 fn migrate_with_native_interchain_token_legacy_flow_data() {
```
milapsheth, line 488:
```suggestion
fn migrate_native_interchain_token_with_flow_amount_succeeds() {
```

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -285,305 +363,283 @@ fn migrate_succeeds_with_multiple_token_ids() {
...
+        migration_data,
+        token_id,
+        current_epoch,
+        flow_in_amount,
+        flow_out_amount,
     );
 }
 
 #[test]
 fn migrate_with_lock_unlock_legacy_flow_data() {
```
milapsheth, line 568: update name as per above

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -1,19 +1,199 @@
...
+    use crate::testutils::setup_its_token;
+    use crate::types::TokenManagerType;
+    use crate::InterchainTokenServiceClient;
+
+    const NEW_INTERCHAIN_TOKEN_SERVICE_WASM: &[u8] =
+        include_bytes!("testdata/stellar_interchain_token_service.optimized.wasm");
+    const TOKEN_MANAGER_WASM_V110: &[u8] =
+        include_bytes!("testdata/stellar_token_manager_v1_1_0.optimized.wasm");
+    const INTERCHAIN_TOKEN_WASM_V110: &[u8] =
+        include_bytes!("testdata/stellar_interchain_token_v1_1_0.optimized.wasm");
```
milapsheth, line 33:
```suggestion
    const TOKEN_MANAGER_WASM_V1_1_0: &[u8] =
        include_bytes!("testdata/stellar_token_manager_v1_1_0.optimized.wasm");
    const INTERCHAIN_TOKEN_WASM_V1_1_0: &[u8] =
        include_bytes!("testdata/stellar_interchain_token_v1_1_0.optimized.wasm");
```

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -1,19 +1,199 @@
...
+        upgrader_client
+            .mock_auths(&[upgrader_upgrade_auth])
+            .upgrade(
+                &its_client.address,
+                &String::from_str(env, NEW_VERSION),
+                &its_wasm_hash,
+                &vec![&env, migration_data.into_val(env)],
+            );
+
+        format_auths(env.auths(), "upgrader.upgrade(...)")
```
milapsheth, line 148: this is not needed since we're using `mock_auths` already. Also, we should only save auth for the method that's being tested in the test (rather than save auths for helper calls)

milapsheth: the caller can call this where needed

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -24,165 +204,44 @@ fn migrate_native_interchain_token_succeeds() {
...
-        .migrate_token(
-            &token_id,
-            &upgrader_client.address,
-            &String::from_str(&env, NEW_VERSION),
-        );
-    log!(&env, "----------CALLED: ITS.MIGRATE_TOKEN----------"); // TODO: Remove
+    let its_migrate_token_auths =
+        migrate_token(&env, &its_client, &upgrader_client, token_id.clone());
+
+    goldie::assert!([upgrader_upgrade_auths, its_migrate_token_auths].join("\n\n"));
```
milapsheth, line 235:
```suggestion
    migrate_token(&env, &its_client, &upgrader_client, token_id.clone());

    goldie::assert!(format_auth(env, "its.migrate_token"));
```

contracts/stellar-interchain-token-service/src/tests/utils/mod.rs
```diff
@@ -70,97 +49,32 @@ pub fn setup_env<'a>() -> (
...
-
-    let its_wasm_hash = env
-        .deployer()
-        .upload_contract_wasm(NEW_INTERCHAIN_TOKEN_SERVICE_WASM);
-    let new_token_manager_wasm_hash = env.deployer().upload_contract_wasm(TOKEN_MANAGER_WASM_V110);
-    let new_interchain_token_wasm_hash = env
-        .deployer()
-        .upload_contract_wasm(INTERCHAIN_TOKEN_WASM_V110);
+pub fn format_auths(
+    auths: std::vec::Vec<(Address, AuthorizedInvocation)>,
```
milapsheth, line 53:
```suggestion
    env: &Env,
```
call `env.auths()` internally perhaps

milapsheth (approved):

contracts/stellar-interchain-token-service/src/tests/migrate.rs
```diff
@@ -0,0 +1,455 @@
...
+        Some(FlowData {
+            token_id,
+            flow_in_amount,
+            flow_out_amount,
+        }),
+    );
+}
+
+#[test]
+fn coverage_migrate_native_interchain_token_succeeds() {
```
milapsheth, line 280:
```suggestion
fn migrate_native_interchain_token_succeeds() {
```

PR #286, opened 2025-03-21
test(interchain-token): add tests for token events

TanvirDeol:
[AXE-8186](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-8186)

[AXE-8186]: https://axelarnetwork.atlassian.net/browse/AXE-8186?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/stellar-interchain-token-service/src/contract.rs
+++ b/contracts/stellar-interchain-token-service/src/contract.rs
@@ -96,7 +96,7 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
         storage::is_trusted_chain(env, chain)
     }
 
-    #[only_owner]
+    #[only_operator]
     fn set_trusted_chain(env: &Env, chain: String) -> Result<(), ContractError> {
         ensure!(
             !storage::is_trusted_chain(env, chain.clone()),
@@ -110,7 +110,7 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
         Ok(())
     }
 
-    #[only_owner]
+    #[only_operator]
     fn remove_trusted_chain(env: &Env, chain: String) -> Result<(), ContractError> {
         ensure!(
             storage::is_trusted_chain(env, chain.clone()),
--- a/contracts/stellar-interchain-token-service/src/tests/trusted_chain.rs
+++ b/contracts/stellar-interchain-token-service/src/tests/trusted_chain.rs
@@ -13,7 +13,7 @@ fn set_trusted_address() {
 
     let chain = String::from_str(&env, "chain");
 
-    assert_auth!(client.owner(), client.set_trusted_chain(&chain));
+    assert_auth!(client.operator(), client.set_trusted_chain(&chain));
 
     goldie::assert!(events::fmt_last_emitted_event::<TrustedChainSetEvent>(&env));
 
@@ -21,13 +21,14 @@ fn set_trusted_address() {
 }
 
 #[test]
-fn set_trusted_chain_fails_if_not_owner() {
+fn set_trusted_chain_fails_if_not_operator() {
     let (env, client, _, _, _) = setup_env();
 
-    let not_owner = Address::generate(&env);
+    let not_operator = Address::generate(&env);
     let chain = String::from_str(&env, "chain");
 
-    assert_auth_err!(not_owner, client.set_trusted_chain(&chain));
+    assert_auth_err!(not_operator, client.set_trusted_chain(&chain));
+    assert_auth_err!(client.owner(), client.set_trusted_chain(&chain));
 }
 
 #[test]
@@ -49,15 +50,26 @@ fn remove_trusted_chain() {
 
     let chain = String::from_str(&env, "chain");
 
-    assert_auth!(client.owner(), client.set_trusted_chain(&chain));
+    assert_auth!(client.operator(), client.set_trusted_chain(&chain));
 
-    assert_auth!(client.owner(), client.remove_trusted_chain(&chain));
+    assert_auth!(client.operator(), client.remove_trusted_chain(&chain));
 
     goldie::assert!(events::fmt_last_emitted_event::<TrustedChainRemovedEvent>(
         &env
     ));
 
     assert!(!client.is_trusted_chain(&chain));
+}
+
+#[test]
+fn remove_trusted_chain_fails_if_not_operator() {
+    let (env, client, _, _, _) = setup_env();
+
+    let not_operator = Address::generate(&env);
+    let chain = String::from_str(&env, "chain");
+
+    assert_auth_err!(not_operator, client.remove_trusted_chain(&chain));
+    assert_auth_err!(client.owner(), client.remove_trusted_chain(&chain));
 }
 
 #[test]


contracts/stellar-interchain-token/src/contract.rs
```diff
@@ -81,7 +83,7 @@ impl StellarAssetInterface for InterchainToken {
 
         Self::receive_balance(&env, to.clone(), amount);
 
-        TokenUtils::new(&env).events().mint(owner, to, amount);
+        MintedEvent { to, amount }.emit(&env);
```
milapsheth, line 86: the new events don't match the old ones so this is a breaking change. Consider why it was being done this way first however. We're using the library provided events for tokens to adhere to the standard token implementation, similar to what ERC20 tokens do. So we shouldn't deviate here. The event definition should match instead with the same fields and topics
https://docs.rs/soroban-token-sdk/latest/src/soroban_token_sdk/event.rs.html#25-28

contracts/stellar-interchain-token/src/tests/test.rs
```diff
@@ -53,6 +55,8 @@ fn register_interchain_token() {
         ),
     );
 
+    goldie::assert!(fmt_last_emitted_event::<MinterAddedEvent>(&env));
```
milapsheth, line 58: another option instead of redefining the events as structs is to retrieve the raw events and goldie assert on that so we still keep using their SDK

milapsheth (changes requested):

contracts/stellar-interchain-token/src/tests/test.rs
milapsheth: We don't need `assert_last_emitted_event`. We can just `goldie::assert` on the full event vector. It won't be pretty printed like our typed events, but it's good enough in this special case. I'd prefer using the sdk util instead of redefining the event ourselves if possible

contracts/stellar-interchain-token/src/contract.rs
```diff
@@ -107,16 +109,13 @@ impl InterchainTokenInterface for InterchainToken {
...
-            ContractError::NotMinter
-        );
+        ensure!(Self::is_minter(env, minter), ContractError::NotMinter);
 
         Self::validate_amount(env, amount);
 
         Self::receive_balance(env, to.clone(), amount);
 
-        TokenUtils::new(env).events().mint(minter, to, amount);
+        MintedEvent { to, amount }.emit(env);
```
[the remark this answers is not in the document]
TanvirDeol: removed event, reverted to default soroban-sdk events

contracts/stellar-interchain-token/src/tests/testdata/transfer_from_succeeds.golden
```diff
@@ -0,0 +1 @@
+Vec(Ok((Contract(CAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAMDR4), Vec(Ok(Symbol(transfer)), Ok(Address(obj#451)), Ok(Address(obj#453))), I128(400))))
```
milapsheth, line 1: this is printing `obj` ids instead of the values. Format it differently, might need to convert to/from val

milapsheth: you could also define the token events using `IntoEvent` but only within the tests for testing purposes. So the contract continues using the TokenUtils but the tests benefit from a well-typed event struct

contracts/stellar-interchain-token/src/tests/test.rs
```diff
@@ -179,6 +183,10 @@ fn transfer() {
     assert_eq!(token.balance(&user1), amount);
 
     assert_auth!(user1, token.transfer(&user1, &user2, &600_i128));
+
+    let events = env.events().all();
+    goldie::assert!(format!("{:#?}", events));
```
milapsheth, line 188:
```suggestion
    goldie::assert!(format!("{:#?}", env.events().all()));
```
intermediate var doesn't seem necessary

contracts/stellar-interchain-token/src/event.rs
```diff
@@ -9,3 +9,12 @@ pub struct MinterAddedEvent {
 pub struct MinterRemovedEvent {
     pub minter: Address,
 }
+
+#[derive(Debug, PartialEq, Eq, IntoEvent)]
+#[event_name("transfer")]
```
milapsheth, line 14:
```suggestion
```
don't think this is needed since the type name matches this

packages/stellar-axelar-std-derive/src/into_event.rs
```diff
@@ -55,9 +55,12 @@ pub fn into_event(input: &DeriveInput) -> proc_macro2::TokenStream {
...
             // and assign them to a variable with the same name as the struct field
             // E.g. let message = Message::try_from_val(env, &data.get(0));
-            // `data` is required to be a `Vec<Val>`
             let data = Vec::<Val>::try_from_val(env, &data)
-                .expect("invalid data format");
+                .unwrap_or_else(|_| {
+                    let mut vec = Vec::<Val>::new(env);
+                    vec.push_back(data);
+                    vec
+                });
```
milapsheth, line 63: This implicit fallback is bad practice. Consider how it can cause unintentional issues. E.g. if another event accidentally used the value directly instead of a vector, then this fallback could cause it to be decoded correctly, even though the data field wasn't formatted as expected.

packages/stellar-axelar-std-derive/src/into_event.rs
```diff
@@ -55,9 +55,12 @@ pub fn into_event(input: &DeriveInput) -> proc_macro2::TokenStream {
             // Parse data from Val to the corresponding types,
             // and assign them to a variable with the same name as the struct field
             // E.g. let message = Message::try_from_val(env, &data.get(0));
-            // `data` is required to be a `Vec<Val>`
             let data = Vec::<Val>::try_from_val(env, &data)
```
milapsheth, line 58:
```suggestion
            let data = if singleton_data {
                Vec::<Val>::try_from_val(env, &data)
            } else {
                let mut vec = Vec::<Val>::new(env);
                vec.push_back(data);
                vec
            };
```
make it explicit instead. Something like
```rust
pub struct TransferEvent {
    pub from: Address,
    pub to: Address,
    #[singleton_data]
    pub amount: i128,
}
```
or
```rust
    #[singleton_data]
    #[data]
    pub amount: i128,
```

@nbayindirli

packages/stellar-axelar-std-derive/src/lib.rs
```diff
@@ -280,7 +280,7 @@ fn ensure_no_args(attr: &Attribute) -> syn::Result<&Path> {
 /// }
 /// }
 /// ```
-#[proc_macro_derive(IntoEvent, attributes(data, event_name))]
+#[proc_macro_derive(IntoEvent, attributes(data, event_name, singleton_data))]
```
milapsheth, line 283:
```suggestion
#[proc_macro_derive(IntoEvent, attributes(event_name, datum, data))]
```
A bit cryptic if someone doesn't know but I like @nbayindirli's `datum` proposal

packages/stellar-axelar-std-derive/src/into_event.rs
```diff
@@ -55,12 +55,17 @@ pub fn into_event(input: &DeriveInput) -> proc_macro2::TokenStream {
...
             // and assign them to a variable with the same name as the struct field
             // E.g. let message = Message::try_from_val(env, &data.get(0));
-            let data = Vec::<Val>::try_from_val(env, &data)
-                .unwrap_or_else(|_| {
-                    let mut vec = Vec::<Val>::new(env);
-                    vec.push_back(data);
-                    vec
-                });
+            let data = if #singleton_data {
+                Vec::<Val>::try_from_val(env, &data)
```
milapsheth, line 59: why are we still trying to convert to Vec<Val>. We know it's supposed to be a singleton, so the unwrap_or_else body is what should always be done

packages/stellar-axelar-std-derive/src/into_event.rs
```diff
@@ -143,18 +149,35 @@ fn event_struct_fields(input: &DeriveInput) -> (EventStructFields, EventStructFi
...
+                    .any(|attr| attr.path().is_ident("singleton_data"))
+            {
                 data_idents.push(ident);
                 data_types.push(&field.ty);
+                if field
+                    .attrs
+                    .iter()
+                    .any(|attr| attr.path().is_ident("singleton_data"))
+                {
+                    singleton_data = true;
```
milapsheth, line 169: Not very robust. What if the user set more than one `singleton_data` attributes?

packages/stellar-axelar-std-derive/src/into_event.rs
```diff
@@ -143,18 +149,35 @@ fn event_struct_fields(input: &DeriveInput) -> (EventStructFields, EventStructFi
...
-            if field.attrs.iter().any(|attr| attr.path().is_ident("data")) {
+            if field.attrs.iter().any(|attr| attr.path().is_ident("data"))
+                || field
+                    .attrs
+                    .iter()
+                    .any(|attr| attr.path().is_ident("singleton_data"))
+            {
                 data_idents.push(ident);
                 data_types.push(&field.ty);
+                if field
```
milapsheth, line 164: nested ifs are ugly. use if / else if / else to separate these cases

packages/stellar-axelar-std-derive/src/into_event.rs
```diff
@@ -133,8 +138,9 @@ fn event_name_snake_case(input: &DeriveInput) -> String {
 type EventIdent<'a> = Vec<&'a Ident>;
 type EventType<'a> = Vec<&'a Type>;
 type EventStructFields<'a> = (EventIdent<'a>, EventType<'a>);
+type EventFieldsInfo<'a> = (EventStructFields<'a>, EventStructFields<'a>, bool);
```
milapsheth, line 141: too complicated of an alias. create a struct

milapsheth, line 141:
```suggestion
type EventFields<'a> = (EventStructFields<'a>, EventStructFields<'a>, bool);
```

packages/stellar-axelar-std-derive/src/into_event.rs
```diff
@@ -55,12 +55,17 @@ pub fn into_event(input: &DeriveInput) -> proc_macro2::TokenStream {
             // Parse data from Val to the corresponding types,
             // and assign them to a variable with the same name as the struct field
             // E.g. let message = Message::try_from_val(env, &data.get(0));
-            let data = Vec::<Val>::try_from_val(env, &data)
-                .unwrap_or_else(|_| {
-                    let mut vec = Vec::<Val>::new(env);
-                    vec.push_back(data);
-                    vec
-                });
+            let data = if #singleton_data {
```
milapsheth, line 58: what about the implementation of `emit`? That needs to change to account for `singleton` as well, otherwise it's not consistent

TanvirDeol: I'm not sure what exactly you're looking for. Do you want the `emit` impl to change so that if there is a `datum`, the actual event struct says `datum`? Or do you want separate if cases to handle `datum` and `data` differently?

packages/stellar-axelar-std-derive/src/into_event.rs
```diff
@@ -25,7 +27,11 @@ pub fn into_event(input: &DeriveInput) -> proc_macro2::TokenStream {
                 #(, IntoVal::<_, Val>::into_val(&self.#data_field_idents, env))*
             ];
 
-            env.events().publish(topics, data);
+            if #has_datum {
+                env.events().publish(topics, data.get(0));
+            } else {
+                env.events().publish(topics, data);
+            }
```
milapsheth, line 34:
```suggestion
            let data = match fields.datum {
                Some((datum_field_ident, _)) => self.#datum_field_ident,
                None => {
                    let (data_field_idents, data_types) = fields.data;
                    ...
                    data
                }
            }

            env.events().publish(topics, data);
```
Add `fields.datum` as a `Option<Val>` so it can be accessed more cleanly?

milapsheth (approved):

PR #296, opened 2025-03-25
fix(interchain-token-service): emit data_hash instead of data in InterchainTransferReceivedEvent

nbayindirli:
AXE-8261

### DIFF
--- a/contracts/stellar-interchain-token-service/src/contract.rs
+++ b/contracts/stellar-interchain-token-service/src/contract.rs
@@ -282,7 +282,7 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
         token_handler::take_token(
             env,
             &caller,
-            Self::token_id_config_with_extended_ttl(env, token_id.clone())?,
+            Self::token_id_config(env, token_id.clone())?,
             amount,
         )?;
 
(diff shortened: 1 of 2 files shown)


contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -527,7 +527,7 @@ impl InterchainTokenService {
             source_address: source_address.clone(),
```
milapsheth, line 527: make PR title scope more accurate and readable, e.g. `fix(interchain-token-service): emit data_hash instead of data in InterchainTransferReceivedEvent`

packages/stellar-axelar-std/src/testdata/panics_when_golden_file_does_not_exist.golden
```diff
@@ -1 +0,0 @@
-something
```
milapsheth, line 1: why wasn't this caught in the original PR? stale golden files should cause test to fail
https://github.com/axelarnetwork/axelar-cgp-stellar/blob/root-auth-upgrader/.github/workflows/test.yaml#L42

milapsheth (approved):

PR #298, opened 2025-03-25
fix(interchain-token): validate minter existence in add_minter and remove_minter

TanvirDeol:
[AXE-8256](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-8256)

### DIFF
--- a/contracts/stellar-interchain-token/src/contract.rs
+++ b/contracts/stellar-interchain-token/src/contract.rs
@@ -125,6 +125,12 @@ impl InterchainTokenInterface for InterchainToken {
 
     #[only_owner]
     fn add_minter(env: &Env, minter: Address) {
+        assert_with_error!(
+            env,
+            !Self::is_minter(env, minter.clone()),
+            ContractError::MinterAlreadyExists
+        );
+
         storage::set_minter_status(env, minter.clone());
 
         MinterAddedEvent { minter }.emit(env);
@@ -132,6 +138,12 @@ impl InterchainTokenInterface for InterchainToken {
 
     #[only_owner]
     fn remove_minter(env: &Env, minter: Address) {
+        assert_with_error!(
+            env,
+            Self::is_minter(env, minter.clone()),
+            ContractError::NotMinter
+        );
+
         storage::remove_minter_status(env, minter.clone());
 
         MinterRemovedEvent { minter }.emit(env);
--- a/contracts/stellar-interchain-token/src/error.rs
+++ b/contracts/stellar-interchain-token/src/error.rs
@@ -11,4 +11,5 @@ pub enum ContractError {
     InsufficientAllowance = 5,
     InsufficientBalance = 6,
     MigrationInProgress = 7,
+    MinterAlreadyExists = 8,
 }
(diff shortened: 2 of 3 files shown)


TanvirDeol:
Isn't the test `remove_minter_succeeds` [defined here](https://github.com/axelarnetwork/axelar-amplifier-stellar/blob/037dd8a60bd0e5ee4157647e50a403227c943956/contracts/stellar-interchain-token/src/tests/test.rs#L401) supposed to fail since it removes an address that was never added? Similar case with `mint_succeeds`

contracts/stellar-interchain-token/src/contract.rs
```diff
@@ -125,13 +125,25 @@ impl InterchainTokenInterface for InterchainToken {
 
     #[only_owner]
     fn add_minter(env: &Env, minter: Address) {
+        assert_with_error!(
+            env,
+            !Self::is_minter(env, minter.clone()),
+            ContractError::MinterAlreadyExists
+        );
+
```
milapsheth, line 133: Look at where `add_minter` is being used. The ITS logic relies adding the same minter being a no-op. If we restrict it here, then that logic should be updated to handle not cause a failure there

milapsheth (changes requested):

contracts/stellar-interchain-token/src/tests/test.rs
```diff
@@ -379,6 +381,22 @@ fn add_minter_fails_without_owner_auth() {
...
+
+    let minter = Address::generate(&env);
+
+    let (token, _) = setup_token(&env);
+
+    assert_auth!(token.owner(), token.add_minter(&minter));
+
+    goldie::assert!(fmt_last_emitted_event::<MinterAddedEvent>(&env));
+
+    token.mock_all_auths().add_minter(&minter);
```
milapsheth, line 397:
```suggestion
    token.mock_all_auths().add_minter(&minter);
```
remove goldie assert since the method being tested here is the failure when re-addition, which doesn't emit an event

contracts/stellar-interchain-token/src/tests/test.rs
```diff
@@ -407,6 +425,7 @@ fn remove_minter_succeeds() {
 
     let (token, _) = setup_token(&env);
 
+    assert_auth!(token.owner(), token.add_minter(&minter1));
```
milapsheth, line 428: rename to `minter`
```suggestion
    assert_auth!(token.owner(), token.add_minter(&minter));
```

contracts/stellar-interchain-token/src/tests/test.rs
```diff
@@ -426,6 +445,23 @@ fn remove_minter_fails_without_minter_auth() {
...
+    let minter = Address::generate(&env);
+
+    let (token, _) = setup_token(&env);
+
+    assert_auth!(token.owner(), token.add_minter(&minter));
+    assert_auth!(token.owner(), token.remove_minter(&minter));
+
+    goldie::assert!(fmt_last_emitted_event::<MinterRemovedEvent>(&env));
+
+    token.mock_all_auths().remove_minter(&minter);
```
milapsheth, line 462:
```suggestion
    token.mock_all_auths().remove_minter(&minter);
```
same

milapsheth (approved):

PR #299, opened 2025-03-26
fix(interchain-token-service): ensure deploy interchain token requires non zero supply or a minter

ahramy:
[AXE-8252](https://axelarnetwork.atlassian.net/browse/AXE-8252)

[AXE-8252]: https://axelarnetwork.atlassian.net/browse/AXE-8252?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -185,6 +185,10 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
 
         ensure!(initial_supply >= 0, ContractError::InvalidInitialSupply);
 
+        if initial_supply == 0 && minter.is_none() {
+            return Err(ContractError::InvalidMinter);
+        }
```
ahramy, line 190: Note: This logic is functionally the same as the examples below. I think using an if statement is more readable, but let me know if you think A or B is better - I'm open to either.

A. 
``` 
  ensure!(
      !(initial_supply == 0 && minter.is_none()),
      ContractError::InvalidMinter
  );
```
B. 
```
  ensure!(
      (initial_supply > 0 || minter.is_some()),
      ContractError::InvalidMinter
  );
```

milapsheth: We use `ensure` everywhere, should be consistent. B looks fine to me, parentheses probably aren't needed

contracts/stellar-interchain-token-service/src/tests/deploy_interchain_token.rs
```diff
@@ -225,26 +225,16 @@ fn deploy_interchain_token_fails_with_zero_initial_supply_and_no_minter() {
     let minter: Option<Address> = None;
     let initial_supply = 0;
 
-    let token_id = client.mock_all_auths().deploy_interchain_token(
```
milapsheth, line 228: the test is called `fails` but it succeeds, we need to review more carefully lol

contracts/stellar-interchain-token-service/src/contract.rs
milapsheth: error type doesn't quite make sense. Create a new error type to better explain this scenario

contracts/stellar-interchain-token-service/src/tests/execute.rs
```diff
@@ -455,7 +455,13 @@ fn execute_fails_with_invalid_amount() {
 
     let invalid_amount = 0;
     let deployer = Address::generate(&env);
-    let (token_id, _) = setup_its_token(&env, &client, &deployer, invalid_amount);
+    let (token_id, _) = setup_its_token(
+        &env,
+        &client,
+        &deployer,
+        invalid_amount,
+        &Some(Address::generate(&env)),
```
milapsheth, line 463: think more about what this test is trying to do. it's failing on receiving a transfer msg with zero amount. It's not trying to fail on setting up token. So a minter doesn't need to be passed here. Just use a different initial_supply instead of reusing `invalid_amount` in the wrong place

contracts/stellar-interchain-token-service/src/tests/executable.rs
```diff
@@ -116,7 +116,7 @@ fn interchain_transfer_execute_succeeds() {
 
     let amount = 1000;
     let deployer = Address::generate(&env);
-    let (token_id, _) = setup_its_token(&env, &client, &deployer, amount);
+    let (token_id, _) = setup_its_token(&env, &client, &deployer, amount, &None);
```
milapsheth, line 119: Passing `&None` is ugly. Avoid using references to `Option` and pass by value instead

contracts/stellar-interchain-token-service/src/testutils.rs
```diff
@@ -53,6 +53,7 @@ pub fn setup_its_token(
     client: &InterchainTokenServiceClient,
     sender: &Address,
     supply: i128,
+    minter: &Option<Address>,
```
milapsheth, line 56: This isn't really needed for any tests, it shouldn't be added

contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -185,6 +185,10 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
```
milapsheth, line 185: Why is the PR title `feat` instead of `fix`?

milapsheth (approved):

PR #300, opened 2025-03-26
feat(example): remove unused pause state variable

ahramy:
[AXE-8257](https://axelarnetwork.atlassian.net/browse/AXE-8257)

### DIFF
--- a/contracts/stellar-axelar-example/src/storage.rs
+++ b/contracts/stellar-axelar-example/src/storage.rs
@@ -14,8 +14,4 @@ enum DataKey {
     #[instance]
     #[value(Address)]
     InterchainTokenService,
-
-    #[instance]
-    #[status]
-    Paused,
 }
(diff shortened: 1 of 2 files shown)


contracts/stellar-axelar-example/src/storage.rs
```diff
@@ -14,8 +14,4 @@ enum DataKey {
     #[instance]
```
milapsheth, line 14: Scope type and value don't make sense. PR title should be `fix(axelar-example): ...`

contracts/stellar-axelar-example/src/storage.rs
```diff
@@ -14,8 +14,4 @@ enum DataKey {
...
     #[value(Address)]
     InterchainTokenService,
-
-    #[instance]
-    #[status]
-    Paused,
```
milapsheth, line 20: Improve the migration CI action to not fail for non-upgradable contracts

contracts/stellar-axelar-example/src/storage.rs
ahramy: got it, updated.

PR #310, opened 2025-03-31
feat(axelar-std-derive): globally extend instance TTL on all contract endpoints

nbayindirli:
AXE-8300

### DIFF
--- a/packages/stellar-axelar-std-derive/src/contractimpl.rs
+++ b/packages/stellar-axelar-std-derive/src/contractimpl.rs
@@ -1,3 +1,4 @@
+use itertools::Itertools;
 use proc_macro2::Ident;
 use quote::quote;
 use syn::{
@@ -8,30 +9,56 @@ use syn::{
 use crate::utils::{parse_env_identifier, PrependStatement};
 
 pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStream, syn::Error> {
-    for method in impl_block.items.iter_mut() {
-        if let ImplItem::Fn(method_fn) = method {
-            let is_contract_endpoint =
-                impl_block.trait_.is_some() || matches!(method_fn.vis, Visibility::Public(_));
-            let is_stateful = has_args(&method_fn);
+    // this needs to be defined before the iteration, because during it, we can't get a reference to the impl block
+    let any_stateful_endpoints = any_stateful_endpoints(impl_block);
+    let all_contract_endpoints = all_contract_endpoints(impl_block);
 
-            if is_stateful && is_contract_endpoint {
-                if is_allowed_during_migration(&method_fn) {
-                    remove_allow_during_migration_attribute(method_fn);
-                } else {
-                    block_during_migration(method_fn)?;
-                }
+    impl_block
+        .items
+        .iter_mut()
+        .filter_map(any_stateful_endpoints)
+        .chunk_by(is_allowed_during_migration)
+        .into_iter()
+        .try_for_each(|(is_allowed, mut method)| {
+            if is_allowed {
+                // if this the attribute is not removed, the compiler will try to resolve it,
+                // and it will need to be defined as a standalone attribute macro
+                method.for_each(remove_allow_during_migration_attribute);
+            } else {
+                method.try_for_each(block_during_migration)?
             }
+            Ok::<_, syn::Error>(())
+        })?;
 
-            if is_contract_endpoint && has_env_param(&method_fn) {
-                instance_ttl_extension(method_fn)?;
-            }
-        }
-    }
+    impl_block
+        .items
+        .iter_mut()
+        .filter_map(all_contract_endpoints)
+        .try_for_each(|method| {
+            instance_ttl_extension(method)?;
+            Ok::<_, syn::Error>(())
+        })?;
 
     Ok(quote! {
         #[soroban_sdk::contractimpl]
         #impl_block
     })
+}
+
+fn all_contract_endpoints(
+    impl_block: &ItemImpl,
+) -> impl Fn(&mut ImplItem) -> Option<&mut ImplItemFn> {
+    let any_contract_endpoint = if impl_block.trait_.is_some() {
+        any
+    } else {
+        any_pub_fn
+    };
+
+    move |item| {
+        any_fn(item)
+            .and_then(any_contract_endpoint)
+            .filter(has_env_param)
+    }
 }
 
 fn instance_ttl_extension(method: &mut ImplItemFn) -> Result<(), syn::Error> {
@@ -47,6 +74,25 @@ fn instance_ttl_extension(method: &mut ImplItemFn) -> Result<(), syn::Error> {
 
 fn has_env_param(fn_: &&mut ImplItemFn) -> bool {
     parse_env_identifier(&fn_.sig.inputs).is_ok()
+}
+
+/// If a function doesn't have any arguments it cannot modify the environment, so it's safe to be called during migration
+fn any_stateful_endpoints(
+    impl_block: &ItemImpl,
+) -> impl Fn(&mut ImplItem) -> Option<&mut ImplItemFn> {
+    // if the block implements a trait, all of its functions are implicitly public, otherwise we only need to match public functions
+    // to help the compiler resolve lifetimes, this is defined first and moved into the closure
+    let any_contract_endpoint = if impl_block.trait_.is_some() {
+        any
+    } else {
+        any_pub_fn
+    };
+
+    move |item| {
+        any_fn(item)
+            .and_then(any_contract_endpoint)
+            .filter(has_args)
+    }
 }
 
 fn block_during_migration(method: &mut ImplItemFn) -> Result<(), syn::Error> {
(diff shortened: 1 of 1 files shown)


packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -35,6 +44,58 @@ pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStrea
     })
 }
 
+/// Returns fn that finds all contract endpoints with an Env parameter
+fn all_contract_endpoints(
```
ahramy, line 48: 👍

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -2,16 +2,25 @@ use itertools::Itertools;
...
-    PathArguments, PathSegment, ReturnType, Stmt, Type, TypePath, Visibility,
+    parse_quote, AngleBracketedGenericArguments, FnArg, GenericArgument, ImplItem, ImplItemFn,
+    ItemImpl, Pat, PathArguments, PathSegment, ReturnType, Stmt, Type, TypePath, Visibility,
 };
 
 use crate::utils::{parse_env_identifier, PrependStatement};
 
 pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStream, syn::Error> {
-    // this needs to be defined before the iteration, because during it, we can't get a reference to the impl block
+    let all_contract_endpoints = all_contract_endpoints(impl_block);
```
ahramy, line 12: It would be cleaner if we can do something like this for better readability:
```
pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStream, syn::Error> {
    process_contract_endpoints(impl_block)?;
    
    process_stateful_endpoints(impl_block)?;

    Ok(quote! {
        #[soroban_sdk::contractimpl]
        #impl_block
    })
}
```

packages/stellar-axelar-std-derive/src/contractstorage.rs
```diff
@@ -330,10 +374,10 @@ impl StorageType {
 
     fn default_ttl_function(&self, key: &TokenStream) -> TokenStream {
         match self {
-            Self::Persistent => {
-                quote! { stellar_axelar_std::ttl::extend_persistent_ttl(env, &#key); }
-            }
-            Self::Instance => quote! { stellar_axelar_std::ttl::extend_instance_ttl(env); },
+            Self::Persistent => quote! {
```
ahramy, line 377: It would be great to add some comments here: for example, extending the TTL isn’t necessary for `Instance` and `Temporary` in this case since we globally extend the instance TTL on all contract endpoints. However, `Persistent` does need an extended TTL function call here.

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -35,6 +44,58 @@ pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStrea
...
+
+    method.prepend_statement(extend_ttl_stmt);
+    Ok(())
+}
+
+fn has_env_param(fn_: &&mut ImplItemFn) -> bool {
+    parse_env_identifier(&fn_.sig.inputs).is_ok()
+}
+
+fn is_env_ref(method: &ImplItemFn, env_ident: &Ident) -> bool {
```
milapsheth, line 85: create a separate ticket to make all env usage use references?

packages/stellar-axelar-std-derive/src/contractstorage.rs
milapsheth: you can rename it, but it'll still be good to clarify why other types don't need extending it for anyone not familiar. The comment could live in the macro's docstring since it's a user facing behaviour

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -35,6 +44,58 @@ pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStrea
...
+    move |item| {
+        any_fn(item)
+            .and_then(any_contract_endpoint)
+            .filter(has_env_param)
+    }
+}
+
+fn instance_ttl_extension(method: &mut ImplItemFn) -> Result<(), syn::Error> {
+    let env_ident = parse_env_identifier(&method.sig.inputs)?;
+
```
milapsheth, line 66: maybe add something like a `normalize_env_ref` helper, which normalizes to `&env`, and then it can be used anywhere without additional checks

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -2,16 +2,25 @@ use itertools::Itertools;
...
-    parse_quote, AngleBracketedGenericArguments, GenericArgument, ImplItem, ImplItemFn, ItemImpl,
-    PathArguments, PathSegment, ReturnType, Stmt, Type, TypePath, Visibility,
+    parse_quote, AngleBracketedGenericArguments, FnArg, GenericArgument, ImplItem, ImplItemFn,
+    ItemImpl, Pat, PathArguments, PathSegment, ReturnType, Stmt, Type, TypePath, Visibility,
 };
 
 use crate::utils::{parse_env_identifier, PrependStatement};
 
 pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStream, syn::Error> {
-    // this needs to be defined before the iteration, because during it, we can't get a reference to the impl block
```
cgorenflo, line 12: don't remove this comment

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -35,6 +44,57 @@ pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStrea
     })
 }
 
+fn all_contract_endpoints(
```
cgorenflo, line 47: I don't think we need to make a distinction between all_contract_endpoints and any_stateful_endpoints. This function also doesn't return all, just the ones with env in the parameter list

cgorenflo: I suggest going with the existing `has_args` check, because you're parsing for `env` later anyway

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -35,6 +44,57 @@ pub fn contractimpl(impl_block: &mut ItemImpl) -> Result<proc_macro2::TokenStrea
...
+}
+
+fn instance_ttl_extension(method: &mut ImplItemFn) -> Result<(), syn::Error> {
+    let env_ident = parse_env_identifier(&method.sig.inputs)?;
+
+    let extend_ttl_stmt: Stmt = if is_env_ref(method, env_ident) {
+        parse_quote! {
+            stellar_axelar_std::ttl::extend_instance_ttl(#env_ident);
+        }
+    } else {
```
cgorenflo, line 70: no need for this, the compiler automatically derefs multiple references, i.e. `&&env` will get resolved to `&env` anyway

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -2,16 +2,25 @@ use itertools::Itertools;
...
     let any_stateful_endpoints = any_stateful_endpoints(impl_block);
 
+    impl_block
+        .items
+        .iter_mut()
+        .filter_map(all_contract_endpoints)
+        .try_for_each(|method| {
+            instance_ttl_extension(method)?;
+            Ok::<_, syn::Error>(())
+        })?;
```
cgorenflo, line 22: with the rest of my suggestions it's possible to add this to the other iteration, but you can also leave it separate if you prefer that

packages/stellar-axelar-std-derive/src/contractstorage.rs
cgorenflo: I suggest instead of renaming the parent function (I like the name `default_ttl_function`) to extract the quotes into `default_{persistent, instance, temporary}_storage_strategy` or something similar, to make it clear that the empty blocks are intentional. I do agree with @ahramy though that commenting the "why" is still a good idea. To me, comments are only a code smell if you need to explain the "how" or the "what"

packages/stellar-axelar-std-derive/src/contractimpl.rs
cgorenflo: If you want to keep the two iterations separate I suggest you swap them so the storage top up will be the first line in each function

cgorenflo (changes requested):

packages/stellar-axelar-std-derive/src/contractimpl.rs
```diff
@@ -9,49 +8,45 @@ use syn::{
...
-                // if this the attribute is not removed, the compiler will try to resolve it,
-                // and it will need to be defined as a standalone attribute macro
-                method.for_each(remove_allow_during_migration_attribute);
-            } else {
-                method.try_for_each(block_during_migration)?
+    for method in impl_block.items.iter_mut() {
+        if let ImplItem::Fn(method_fn) = method {
+            let is_contract_endpoint =
+                impl_block.trait_.is_some() || matches!(method_fn.vis, Visibility::Public(_));
+            let is_stateful = has_args(&method_fn);
```
cgorenflo, line 15: the issue isn't really that the fn has arguments, but that we can restrict the code gen to fns that have env as a parameter. I split that previously as an optimization to filter more before doing the more expensive parsing, but now, the env parsing is actually done multiple times instead, so there is no benefit to that check.
I suggest that you try to parse the env_param here (either with `if let Ok(env) = parse_env_identifier(...)` or with `parse_env_identifier(...).and_then(...)`) and then pass it into `block_during_migration` and `instance_ttl_extension` instead of parsing it three times

cgorenflo (approved):

PR #307, opened 2025-03-28
fix(interchain-token-service): prevent large flow_limit from causing failures

TanvirDeol:
[AXE-8254](https://axelarnetwork.atlassian.net/jira/software/c/projects/AXE/boards/21?selectedIssue=AXE-8254)

[AXE-8254]: https://axelarnetwork.atlassian.net/browse/AXE-8254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #313, opened 2025-04-03
fix(interchain-token-service): handle large flow limits

milapsheth:
[AXE-8254]
- [x] Update tests

[AXE-8254]: https://axelarnetwork.atlassian.net/browse/AXE-8254?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

closes #307


contracts/stellar-interchain-token-service/src/flow_limit.rs
```diff
@@ -54,19 +54,28 @@ impl FlowDirection {
...
+            ContractError::FlowAmountExceededLimit
+        );
 
-        let new_flow = self
-            .flow(env, token_id.clone())
+        let flow = self.flow(env, token_id.clone());
+        let reverse_flow = self.reverse_flow(env, token_id.clone());
+
+        let new_flow = flow
             .checked_add(flow_amount)
```
ahramy, line 66: Should we ensure that `flow_amount` is always positive here? I see that the `add_flow` function is only called in `interchain_transfer` or `execute_transfer_message`, both of which ensure that the amount is positive. However, since this is a public function, it might be better to explicitly enforce that `flow_amount` is positive here.

contracts/stellar-interchain-token-service/src/tests/flow_limit.rs
```diff
@@ -146,6 +149,120 @@ fn execute_its_transfer(
...
+    flow_limit: i128,
+    flows: std::vec::Vec<Flow>,
+}
+
+impl TestCase {
+    /// Flips the direction of the flows in the test case. The test case should still logically behave in the same way.
+    fn flip(self) -> Self {
+        Self {
+            flow_limit: self.flow_limit,
+            flows: self.flows.into_iter().map(|flow| flow.flip()).collect(),
```
ahramy, line 178:
```suggestion
            flows: self.flows.into_iter().map(Flow::flip).collect(),
```

contracts/stellar-interchain-token-service/src/tests/flow_limit.rs
```diff
@@ -146,6 +149,120 @@ fn execute_its_transfer(
...
+                    )
+                }
+            }
+        });
+
+    if let Some(expected_error) = expected_error {
+        assert_contract_err!(result, expected_error)
+    } else {
+        assert_ok!(assert_ok!(result));
+    }
```
ahramy, line 249: perhaps using `match` would be simpler?
```suggestion
    match expected_error {
        Some(error) => assert_contract_err!(result, error),
        None => assert_ok!(assert_ok!(result)),
    }
```

contracts/stellar-interchain-token-service/src/flow_limit.rs
milapsheth: For any amount value, it should be validated in the contract entrypoint that first receives it, so all internal methods don't need to start worrying about validation. I'd consider solving it with a restrictive type in the future. Even if I add a check, there's no way to trigger the error case for code coverage, since entrypoints already verify it

contracts/stellar-interchain-token-service/src/tests/flow_limit.rs
milapsheth: sure, sometimes clippy suggests collapsing a simple match into an `if let else`, but seems fine in this case

ahramy (approved):

PR #322, opened 2025-04-04
fix: update compile gh action to build and test all contracts separately

ahramy:


.github/workflows/compile.yaml
```diff
@@ -28,11 +28,25 @@ jobs:
...
           RUST_BACKTRACE: 1
 
+      - name: Build and test all packages and contracts separately
+        run: |
+          set -e
+          for group in contracts packages; do
+            for dir in $group/*; do
+              if [ -d "$dir" ] && [ -f "$dir/Cargo.toml" ]; then
+                echo "==> Building and testing $dir"
+                (cd "$dir" && cargo build && cargo test)
```
milapsheth, line 44: there is a separate test action, so why test again? The build cmd doesn't match the above one

milapsheth: refer to how we do it in the test action for consistency
https://github.com/axelarnetwork/axelar-cgp-stellar/blob/main/.github/workflows/test.yaml#L37

ahramy: make sense, updated.

contracts/stellar-axelar-gas-service/src/interface.rs
```diff
@@ -3,7 +3,7 @@ use stellar_axelar_std::types::Token;
 use stellar_axelar_std::{contractclient, soroban_sdk, Address, Bytes, Env, String};
 
 use crate::error::ContractError;
-
+#[allow(dead_code)]
```
ahramy, line 6: due to: https://github.com/axelarnetwork/axelar-amplifier-stellar/actions/runs/14273939831/job/40013151568?pr=322

contracts/stellar-axelar-gas-service/src/interface.rs
```diff
@@ -4,6 +4,7 @@ use stellar_axelar_std::{contractclient, soroban_sdk, Address, Bytes, Env, Strin
 
 use crate::error::ContractError;
 
+#[allow(dead_code)]
```
milapsheth, line 7: This wasn't needed before, why is it needed now?

.github/workflows/compile_and_test.yaml
```diff
@@ -12,9 +12,43 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  compile:
+    name: Compile
```
milapsheth, line 16: Should be ran parallelly otherwise we have to wait longer on PRs

contracts/stellar-axelar-gas-service/src/interface.rs
ahramy: Yeah, with a separate build `for C in contracts/*; do cd $C; RUSTFLAGS="-D warnings" cargo build --locked --release; cd ../../; done`, it's now throwing dead_code errors, so I had to add it

.github/workflows/compile_and_test.yaml
ahramy: I found that there's case that the build fails but the test passes. I believe if the build fails, test shouldn't run - so I combined them into one

.github/workflows/compile.yaml
```diff
@@ -15,6 +15,10 @@ jobs:
   soroban-compilation:
     name: Soroban Compilation
     runs-on: blacksmith-4vcpu-ubuntu-2204
```
milapsheth, line 17: Consider bumping to 8vcpu if it makes a significant difference

milapsheth (approved):

.github/workflows/compile.yaml
ahramy: There's no difference for this gh action, I will just use 4vcpu.

- 4vcpu: [Compile 1m 40s](https://github.com/axelarnetwork/axelar-amplifier-stellar/actions/runs/14275709995/job/40018222559) 
- 8vcpu: [Compile 1m 41s](https://github.com/axelarnetwork/axelar-amplifier-stellar/actions/runs/14275680408/job/40018145595)

PR #326, opened 2025-04-06
fix: release github action to have dry run and skip publish release

ahramy:
[AXE-8492](https://axelarnetwork.atlassian.net/browse/AXE-8492)

- Dry Run Test: https://github.com/axelarnetwork/axelar-cgp-soroban-testing/actions/runs/14319039058
- Upload to R2 Test: https://github.com/axelarnetwork/axelar-cgp-soroban-testing/actions/runs/14319837771
- Release Test: https://github.com/axelarnetwork/axelar-cgp-soroban-testing/actions/runs/14319180125

[AXE-8492]: https://axelarnetwork.atlassian.net/browse/AXE-8492?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/.github/workflows/release-dry-run.yaml
+++ b/.github/workflows/release-dry-run.yaml
@@ -1,0 +1,50 @@
+# This workflow performs a dry-run release using release-plz
+# when a release PR is opened (not merged). It helps verify the release without publishing.
+
+name: Dry-Run Release
+
+on:
+  pull_request:
+    branches:
+      - main
+      - "releases/**"
+    types: [opened] # Triggers when PR is opened
+
+  workflow_dispatch: # Optional manual trigger
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  dry-run-release:
+    name: Dry-Run Release
+
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.pull_request.merged == false && contains(github.event.pull_request.labels.*.name, 'release'))
+
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+
+    outputs:
+      releases: ${{ steps.prepare-matrix.outputs.releases }}
+      commit_hash: ${{ steps.get-commit-hash.outputs.hash }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.PAT_TOKEN }}
+
+      - name: Get commit hash
+        id: get-commit-hash
+        run: echo "hash=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
+      - name: Dry-run
+        id: dry-run
+        uses: MarcoIeni/release-plz-action@fff938ea2923fdfa35797fff63537b00f30fe6f9 # v0.5.82
+        with:
+          command: release --dry-run
+          version: "0.3.123"
+        env:
+          CARGO_NET_GIT_FETCH_WITH_CLI: true
+          GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
+          CARGO_REGISTRY_TOKEN: ${{ secrets.CARGO_REGISTRY_TOKEN }}
--- a/.github/workflows/release-dry-run.yaml
+++ b/.github/workflows/release-dry-run.yaml
@@ -3,7 +3,7 @@
 name: Dry-Run Release
 
 on:
-  workflow_dispatch: # Manual trigger only
+  workflow_dispatch:
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
--- a/.github/workflows/release-dry-run.yaml
+++ b/.github/workflows/release-dry-run.yaml
@@ -1,16 +1,9 @@
 # This workflow performs a dry-run release using release-plz
-# when a release PR is opened (not merged). It helps verify the release without publishing.
 
 name: Dry-Run Release
 
 on:
-  pull_request:
-    branches:
-      - main
-      - "releases/**"
-    types: [opened] # Triggers when PR is opened
-
-  workflow_dispatch: # Optional manual trigger
+  workflow_dispatch: # Manual trigger only
 
 concurrency: ${{ github.workflow }}-${{ github.ref }}
 
@@ -18,11 +11,7 @@ jobs:
   dry-run-release:
     name: Dry-Run Release
 
-    if: |
-      github.event_name == 'workflow_dispatch' ||
-      (github.event.pull_request.merged == false && contains(github.event.pull_request.labels.*.name, 'release'))
-
-    runs-on: blacksmith-4vcpu-ubuntu-2204
+    runs-on: blacksmith-2vcpu-ubuntu-2204
 
     outputs:
       releases: ${{ steps.prepare-matrix.outputs.releases }}
--- a/.github/workflows/release-manual.yaml
+++ b/.github/workflows/release-manual.yaml
@@ -1,0 +1,88 @@
+# This workflow only build and upload to CF R2 for a specific contract name passed as input.
+name: Manual Release (Only build and upload to CF R2)
+
+on:
+  workflow_dispatch:
+    inputs:
+      contract-path:
+        description: "Path of the contract to release (e.g. contracts)"
+        required: true
+        type: string
+      contract-name:
+        description: "Name of the contract to release (e.g. stellar-axelar-gateway)"
+        required: true
+        type: string
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  prepare-release:
+    name: Prepare release metadata
+
+    runs-on: blacksmith-4vcpu-ubuntu-2204
+
+    outputs:
+      releases: ${{ steps.prepare-matrix.outputs.releases }}
+      commit_hash: ${{ steps.get-commit-hash.outputs.hash }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.PAT_TOKEN }}
+
+      - name: Get commit hash
+        id: get-commit-hash
+        run: echo "hash=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
+      - name: Prepare release matrix from input
+        id: prepare-matrix
+        run: |
+          path="${{ github.event.inputs.contract-path }}"
+          name="${{ github.event.inputs.contract-name }}"
+          cargo_toml="$path/$name/Cargo.toml"
+
+          if [ ! -f "$cargo_toml" ]; then
+            echo "Contract not found at $path/$name"
+            exit 1
+          fi
+
+          version=$(grep -E '^version\s*=' "$cargo_toml" | sed -E 's/.*=\s*"(.*)"/\1/' | head -1)
+          tag="${name}-v${version}"
+
+          release="{\"package_name\":\"$name\",\"prs\":[],\"tag\":\"$tag\",\"version\":\"$version\"}"
+          releases_json=$(echo "[$release]" | jq -c '.')
+
+          echo "Parsed release from input: $releases_json"
+          echo "releases=$releases_json" >> "$GITHUB_OUTPUT"
+
+  build:
+    needs: prepare-release
+    uses: ./.github/workflows/reusable-build.yaml
+    with:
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -57,7 +72,38 @@ jobs:
       - name: Prepare JSON output to be a matrix GHA format
         id: prepare-matrix
         run: |
```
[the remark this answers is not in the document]
milapsheth: :+1: Create a script that can be independently run and tested

.github/workflows/release.yaml
```diff
@@ -35,20 +41,29 @@ jobs:
...
 
-      - name: Install Rust toolchain
-        uses: dtolnay/rust-toolchain@stable
-
       - name: Get commit hash
         id: get-commit-hash
         run: echo "hash=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
 
-      # Creates git tags and publishes the crates of the new releases
+      - name: Dry-run
```
milapsheth, line 48: Why run the dry run and the release cmd at the same time? The release action will still fail on the dry run making us manually deal with it. Let's have a dry run flag so we can manually do a dry run from the release PR to verify before merging

.github/workflows/release.yaml
```diff
@@ -57,7 +72,38 @@ jobs:
       - name: Prepare JSON output to be a matrix GHA format
         id: prepare-matrix
         run: |
-          echo "releases=$(echo '${{ steps.publish-release.outputs.releases }}' | jq -c '.')" >> $GITHUB_OUTPUT
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ] && [ "${{ github.event.inputs.skip-publish-release }}" == "true" ]; then
```
milapsheth, line 75: The skip publish release flag should not be needed. You said that `release-plz` will only output the contracts that actually got released. Those are the only ones that should be published then. Here's how this could introduce another bug:
- Contract A was released at v1.0.0
- Contract B was released at v1.1.0 later, but the R2 upload didn't happen
- We re-ran the release action but skipped the publish release. Based on this code, it's going to build Contract A v1.0.0 again but on the current commit and overwrite the R2 bytecode for it, or it'll fail since R2 doesn't allow overwrite.

So this approach isn't sound. Let's create a separate workflow that given a contract name, it'll build and publish that contract to R2. Ideally, we don't even need this and can remove the workflow later. We have already pushed to crates.io so there's no reason to overwrite that. Instead, we can use this other action to build and push to R2

milapsheth (changes requested):

.github/workflows/release.yaml
ahramy: Yeah, the reasoning behind having a dry run here is that if the dry run fails, it won't publish the packages to crates.io. This helps prevent accidental publishing when there's an issue, the issue was that dummy-contract was being published, but it didn’t actually exist. 
And Makes sense to have a dry run flag, updated it.

.github/workflows/release.yaml
ahramy: Make sense, updated.

.github/workflows/release-dry-run.yaml
```diff
@@ -0,0 +1,50 @@
...
+name: Dry-Run Release
+
+on:
+  pull_request:
+    branches:
+      - main
+      - "releases/**"
+    types: [opened] # Triggers when PR is opened
+
+  workflow_dispatch: # Optional manual trigger
```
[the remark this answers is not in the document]
ahramy: Yes, workflow_dispatch is defined intentionally without inputs to allow manual triggering without requiring any user input

ahramy:
> Will approve once `release-dry-run` is actually running on this PR's CI

It won’t run for this PR since it’s not tagged with the `release` label. I’ve configured the dry-run to only run for PRs that aren’t merged and have the release label as you can see in the code - Do we want the dry-run to run on every PR instead?

`
      (github.event.pull_request.merged == false && contains(github.event.pull_request.labels.*.name, 'release'))
`

.github/workflows/release.yaml
milapsheth: the problem is that it's overwriting with a new build when there's no release for it, i.e. Contract A v1.0.0 release gets overwritten with potentially new bytecode which is breaking

.github/workflows/release-dry-run.yaml
```diff
@@ -0,0 +1,50 @@
...
+# when a release PR is opened (not merged). It helps verify the release without publishing.
+
+name: Dry-Run Release
+
+on:
+  pull_request:
+    branches:
+      - main
+      - "releases/**"
+    types: [opened] # Triggers when PR is opened
```
milapsheth, line 11: why is it being run on every PR? This only needs to be run manually

.github/workflows/release-dry-run.yaml
milapsheth, line 13: workflow dispatch is the way to trigger this action

.github/workflows/release-dry-run.yaml
```diff
@@ -0,0 +1,50 @@
...
+
+concurrency: ${{ github.workflow }}-${{ github.ref }}
+
+jobs:
+  dry-run-release:
+    name: Dry-Run Release
+
+    if: |
+      github.event_name == 'workflow_dispatch' ||
+      (github.event.pull_request.merged == false && contains(github.event.pull_request.labels.*.name, 'release'))
```
milapsheth, line 23: checks aren't needed

.github/workflows/release.yaml
```diff
@@ -41,29 +35,20 @@ jobs:
         with:
           token: ${{ secrets.PAT_TOKEN }}
 
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
```
milapsheth, line 40: why was this removed before? Is this not needed in the dry run action?

.github/workflows/release-dry-run.yaml
```diff
@@ -0,0 +1,50 @@
...
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.PAT_TOKEN }}
+
+      - name: Get commit hash
+        id: get-commit-hash
+        run: echo "hash=$(git rev-parse HEAD)" >> "$GITHUB_OUTPUT"
+
+      - name: Dry-run
+        id: dry-run
```
milapsheth, line 42: don't need to setup rust toolchain here?

.github/workflows/release-manual.yaml
```diff
@@ -0,0 +1,88 @@
+# This workflow only build and upload to CF R2 for a specific contract name passed as input.
+name: Manual Release (Only build and upload to CF R2)
```
milapsheth, line 2:
```suggestion
name: Upload to R2
```
provide a more concise name

.github/workflows/release-manual.yaml
```diff
@@ -0,0 +1,88 @@
+# This workflow only build and upload to CF R2 for a specific contract name passed as input.
```
milapsheth, line 1:
```suggestion
# This workflow builds the contract provided as input and uploads it to CF R2.
```
rephrase for clarity

.github/workflows/release-manual.yaml
```diff
@@ -0,0 +1,88 @@
...
+name: Manual Release (Only build and upload to CF R2)
+
+on:
+  workflow_dispatch:
+    inputs:
+      contract-path:
+        description: "Path of the contract to release (e.g. contracts)"
```
milapsheth, line 8: isn't this always `contracts`? `packages` don't have wasm code compiled

.github/workflows/release.yaml
ahramy: Yeah, the release failed after some crates had already been published - that’s what happened with `dummy-contract`. All contracts were published to crates.io except for dummy-contract, because it wasn’t compilable and caused the workflow to fail. The dry-run doesn’t upload to crates.io, but it performs all the checks as if it would. I’ll add the dry-run back as a safety check.

.github/workflows/release-manual.yaml
```diff
@@ -0,0 +1,88 @@
...
+
+    outputs:
+      releases: ${{ steps.prepare-matrix.outputs.releases }}
+      commit_hash: ${{ steps.get-commit-hash.outputs.hash }}
+
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+        with:
+          token: ${{ secrets.PAT_TOKEN }}
```
milapsheth, line 32: to allow arbitrary overwrite, I'd prefer if we take the `version` as input and checkout the contract at that tag. This ensures that we're building from the right tag, rather than from main which is a different version. If the contract needs a patch, then a new release should be created anyways.

.github/workflows/release.yaml
ahramy: it's not required in this step, removed.

.github/workflows/release-upload-to-r2.yaml
```diff
@@ -26,35 +27,45 @@ jobs:
...
-          version=$(grep -E '^version\s*=' "$cargo_toml" | sed -E 's/.*=\s*"(.*)"/\1/' | head -1)
           tag="${name}-v${version}"
 
+          # If version is provided, checkout the corresponding git tag.
+          # Otherwise, extract the version from Cargo.toml and update the tag.
+          if [ -n "$version" ]; then
+            git fetch --tags
+            git checkout "$tag"
+          else
+            echo "No version provided. Detecting version from Cargo.toml..."
```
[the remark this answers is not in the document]
ahramy: I agree with that opinion. then We should manually delete and re-create the tag and release page. This is because we merged a PR that triggered a release but failed, then reverted the release PR, which created the release page but the associated commit is no longer available. - Let me know if it sounds ok.

For example:
- a release page: https://github.com/axelarnetwork/axelar-amplifier-stellar/releases/tag/stellar-upgrader-v1.1.0
- associated commit: https://github.com/axelarnetwork/axelar-amplifier-stellar/commit/0346a09f5fef998ff2e7ee2e31cae8806f0256f5

.github/workflows/release-upload-to-r2.yaml
```diff
@@ -26,35 +27,45 @@ jobs:
...
+            echo "Contract not found at contracts/$name"
             exit 1
           fi
 
-          version=$(grep -E '^version\s*=' "$cargo_toml" | sed -E 's/.*=\s*"(.*)"/\1/' | head -1)
           tag="${name}-v${version}"
 
+          # If version is provided, checkout the corresponding git tag.
+          # Otherwise, extract the version from Cargo.toml and update the tag.
+          if [ -n "$version" ]; then
```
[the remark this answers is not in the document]
ahramy: got it, updated

.github/workflows/release.yaml
```diff
@@ -57,6 +52,10 @@ jobs:
       - name: Prepare JSON output to be a matrix GHA format
         id: prepare-matrix
         run: |
+          if [ "${{ steps.publish-release.outputs.releases }}" == "[]" ]; then
+            echo "release-plz output contains an empty release list — exiting early."
+            exit 1
+          fi
           echo "releases=$(echo '${{ steps.publish-release.outputs.releases }}' | jq -c '.')" >> $GITHUB_OUTPUT
```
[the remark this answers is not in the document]
ahramy: I applied the commit, but my editor removed the newline because there was already one at line 60.

.github/workflows/release.yaml
```diff
@@ -1,15 +1,8 @@
...
-# It's triggered once a release PR (created by the create-release-pr workflow) gets merged to other release branches.
+
 name: Release
 on:
-  # Once the PR gets merged to `main` or other `releases` branches
-  pull_request:
-    branches:
-      - main
-      - "releases/**"
-    types: [closed]
```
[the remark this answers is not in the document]
ahramy: yeah, currently it will fail right away - so I'd like to keep it manual for now, we can update it later

milapsheth (approved):

.github/workflows/release-upload-to-r2.yaml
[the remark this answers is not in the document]
ahramy: Had discussion and decide not to delete and re-create the release page, decided to bump up the version.

PR #331, opened 2025-04-08
chore: bump up minor versions

ahramy:
bump up minor versions

### DIFF
--- a/packages/stellar-axelar-std-derive/src/lib.rs
+++ b/packages/stellar-axelar-std-derive/src/lib.rs
@@ -442,3 +442,4 @@ impl<T> MapTranspose<T> for Option<T> {
         self.map(f).transpose()
     }
 }
+ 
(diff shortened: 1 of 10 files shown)


packages/stellar-axelar-std-derive/src/lib.rs
```diff
@@ -1,7 +1,6 @@
 //! Note: The tests are located in the `stellar-axelar-std` package instead of `stellar-axelar-std-derive`
 //!
 //! This ensures compatibility and prevents cyclic dependency issues during testing and release.
-
```
milapsheth, line 4: Update harmless comments instead of removing newlines maybe in Cargo.toml or something

milapsheth (approved):

packages/stellar-axelar-std-derive/src/lib.rs
ahramy: updated

PR #336, opened 2025-04-10
fix(interchain-token-service): optimize code for creating upgrade auth entires

ahramy:
This is a follow-up PR: https://github.com/axelarnetwork/axelar-amplifier-stellar/pull/334

### DIFF
--- a/packages/stellar-axelar-std/src/error.rs
+++ b/packages/stellar-axelar-std/src/error.rs
@@ -234,3 +234,24 @@ macro_rules! mock_auth {
         $crate::mock_auth!($caller, $client.$method($($arg),*), &[])
     }};
 }
+
+/// Creates a SubContractInvocation for contract upgrades and migrations
+///
+/// # Arguments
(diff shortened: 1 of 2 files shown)


packages/stellar-axelar-std/src/error.rs
```diff
@@ -234,3 +234,24 @@ macro_rules! mock_auth {
         $crate::mock_auth!($caller, $client.$method($($arg),*), &[])
     }};
 }
+
+/// Creates a SubContractInvocation for contract upgrades and migrations
```
milapsheth, line 238: Having this in error.rs doesn't make sense. Also, will modifying the std crate update all other crates on release? If so, keep the helper in the ITS file for now

ahramy: yeah, it will update all other crates on release. I see. updated to have it in the ITS file for now

milapsheth (approved):

PR #358, opened 2025-07-17
feat(interchain-token-service): add minter and test execute

ahramy:


contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -439,6 +439,29 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
 
         Ok(())
     }
+
+    fn test_execute(
```
ahramy, line 443: `fn test_execute` will be removed

contracts/stellar-interchain-token-service/src/token_handler.rs
```diff
@@ -38,16 +39,46 @@ pub fn give_token(
...
         TokenManagerType::NativeInterchainToken => {
-            token_manager.mint(env, &token_address, recipient, amount)
+            token_manager_client.mint_from(env, &token_address, recipient, amount)
         }
         TokenManagerType::LockUnlock => {
-            token_manager.transfer(env, &token_address, recipient, amount)
+            token_manager_client.transfer(env, &token_address, recipient, amount)
         }
-        TokenManagerType::MintBurn => token_manager.mint(env, &token_address, recipient, amount),
+        TokenManagerType::MintBurn => {
```
isi8787, line 51: can you explain what the authorization mechanism is to prevent anyone from minting. Why cant a random account access the `else` clause to still mint tokens

PR #359, opened 2025-07-21
feat(interchain-token-service): add minter for interchain token and mint from for stellar classic asset

ahramy:
[AXE-9815](https://axelarnetwork.atlassian.net/browse/AXE-9815)

[AXE-9815]: https://axelarnetwork.atlassian.net/browse/AXE-9815?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/contracts/stellar-axelar-example/src/tests/link_token.rs
+++ b/contracts/stellar-axelar-example/src/tests/link_token.rs
@@ -185,6 +185,7 @@ fn setup_interchain_tokens(
             &0,
             &Some(test_env.deployer.clone()),
         );
+
     let source_token_address = test_env
         .source_config
         .its
@@ -220,6 +221,7 @@ fn setup_interchain_tokens(
             &test_env.initial_supply,
             &Some(test_env.deployer.clone()),
         );
+
     let destination_token_address = test_env
         .destination_config
         .its
@@ -291,6 +293,7 @@ fn execute_link_token(
         .destination_config
         .gateway
         .approve_messages(&link_messages, &proof);
+
     test_env.destination_config.its.execute(
         &test_env.hub_chain,
         &message_id,
@@ -384,20 +387,13 @@ fn verify_balances(
     source_manager_type: TokenManagerType,
     destination_manager_type: TokenManagerType,
 ) {
-    // Destination: tokens should be transferred to recipient
     let destination_token_client =
         token::TokenClient::new(&test_env.env, destination_token_address);
-    assert_eq!(
-        destination_token_client.balance(&test_env.destination_config.app.address),
-        0
-    );
-
     assert_eq!(
         destination_token_client.balance(&test_env.recipient),
         test_env.transfer_amount
     );
 
-    // Check destination token manager balance based on manager type
     let destination_token_manager = test_env
         .destination_config
         .its
@@ -417,7 +413,6 @@ fn verify_balances(
         );
     }
 
-    // Check source token balances
     let source_token_client = token::TokenClient::new(&test_env.env, source_token_address);
     let source_token_manager = test_env
         .source_config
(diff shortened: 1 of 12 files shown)


contracts/stellar-interchain-token-service/src/interface.rs
```diff
@@ -365,4 +365,24 @@ pub trait InterchainTokenServiceInterface:
...
+    /// # Returns
+    /// - `Ok(())`: If the minter was successfully added.
+    ///
+    /// # Errors
+    /// - [`ContractError::InvalidMinter`]: If the minter address is invalid.
+    /// - [`ContractError::InvalidTokenAddress`]: If the token address is invalid or the token contract doesn't exist.
+    ///
+    /// # Authorization
+    /// - [`OperatableInterface::operator`] must authorize.
+    fn add_minter(env: &Env, token_address: Address, minter: Address) -> Result<(), ContractError>;
```
milapsheth, line 387: there's no such ability in EVM ITS

contracts/stellar-interchain-token-service/src/contract.rs
```diff
@@ -439,6 +439,22 @@ impl InterchainTokenServiceInterface for InterchainTokenService {
 
         Ok(())
     }
+
+    #[only_operator]
```
milapsheth, line 443: dangerous to allow adding a minter for any existing token, that too from the operator and not the owner

milapsheth: we shouldn't add functionality in production contracts just to facilitate testing. should handle this in a different manner

contracts/stellar-interchain-token-service/src/token_handler.rs
```diff
@@ -38,16 +39,25 @@ pub fn give_token(
...
+        }
+        TokenManagerType::MintBurn => {
+            if InterchainTokenClient::new(env, &token_address)
+                .try_token_id()
+                .is_ok()
+            {
+                token_manager_client.mint_from(env, &token_address, recipient, amount);
+            } else {
+                token_manager_client.mint(env, &token_address, recipient, amount);
+            }
```
milapsheth, line 59: interchain tokens aren't meant to use mint/burn type. We should always call `mint` here. Also, this is not reliable. A token may implement `token_id` method but might not actually be an interchain token.

To facilitate testing, create a test/example custom token that implements this interface instead

contracts/stellar-interchain-token-service/src/token_manager.rs
```diff
@@ -24,6 +27,14 @@ impl TokenManagerClientExt for TokenManagerClient<'_> {
...
 
     fn mint(&self, env: &Env, token_address: &Address, recipient: &Address, amount: i128) {
+        let _: Val = self.execute(
+            token_address,
+            &Symbol::new(env, "mint"),
+            &vec![env, recipient.to_val(), amount.into_val(env)],
+        );
+    }
+
+    fn mint_from(&self, env: &Env, token_address: &Address, recipient: &Address, amount: i128) {
```
milapsheth, line 37: actually, a proper fix would be to transfer the admin for interchain tokens to their token manager from ITS, so that it can call `mint` directly. This way the same interface can be used for both types. This requires a migration since admin transfer needs to be done for all deployed interchain tokens. Could consider it in a follow up PR perhaps

contracts/stellar-axelar-example/src/tests/link_token.rs
milapsheth, line 1: how is this relevant to the example contract? The example contract isn't being used to facilitate link token at all. The unit tests in ITS should cover this. If a scenario is missing, add it there. Otherwise, the actual example of how to link will be in the deployment scripts

ahramy:
Closing this one.

PR #362, opened 2025-08-05
chore(interchain-token-service): update comments for token handlers

ahramy:


contracts/stellar-interchain-token-service/src/token_handler.rs
```diff
@@ -80,9 +80,9 @@ pub fn post_token_manager_deploy(
         // account abstraction, which eliminates the need for ERC20-like approvals used on EVM chains.
         // The token manager can directly transfer tokens as needed.
         TokenManagerType::LockUnlock => {}
-        // For mint/burn token managers, the user needs to grant mint permission to the token manager
```
milapsheth, line 83: This is applicable for both types of tokens

contracts/stellar-interchain-token-service/src/token_handler.rs
```diff
@@ -80,9 +80,9 @@ pub fn post_token_manager_deploy(
...
-        // Stellar Classic Assets require setting the token manager as the admin to allow minting the token,
```
milapsheth, line 84: Only classic tokens require setting it as the admin

contracts/stellar-interchain-token-service/src/token_handler.rs
ahramy: make sense, reverted

contracts/stellar-interchain-token-service/src/token_handler.rs
ahramy: updated.

PR #363, opened 2025-08-05
feat(interchain-token): add token manager to interchain token

ahramy:
- add token manager to interchain token
- `mint` is being called from token manager instead of owner (initially the owner and token manager are the same)

### DIFF
--- a/contracts/stellar-interchain-token/src/contract.rs
+++ b/contracts/stellar-interchain-token/src/contract.rs
@@ -157,14 +157,6 @@ impl InterchainTokenInterface for InterchainToken {
     #[only_owner]
     fn set_token_manager(env: &Env, token_manager: Address) {
         storage::set_token_manager(env, &token_manager);
-
-        if !Self::is_minter(env, token_manager.clone()) {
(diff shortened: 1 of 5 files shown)


contracts/stellar-interchain-token/src/contract.rs
```diff
@@ -29,6 +29,8 @@ impl InterchainToken {
     ) {
         interfaces::set_owner(&env, &owner);
 
+        storage::set_token_manager(&env, &owner);
```
milapsheth, line 32: Why are we changing the interchain token contract again? That requires migrating every instance again. The owner can be kept as ITS as we discussed in a previous PR. The token manager is already an additional minter

milapsheth: The point is that the interchain token isn't meant to be used for custom token linking. That's meant for other custom logic, they just need to implement the mint interface and allow the token manager permission to call it

milapsheth (changes requested):

ahramy:
Thanks for comments and those make sense. I will leave comments open and close this PR.

